PR #141, opened 2013-12-05
Allow channels to be usefully imported by other nix expressions.

shlevy:
Is backwards-compatible, but requires NixOS/nix#180 for new feature

### DIFF
--- a/src/lib/Hydra/View/NixExprs.pm
+++ b/src/lib/Hydra/View/NixExprs.pm
@@ -31,6 +31,10 @@ sub process {
 
 let
 
+  maybeStorePath = if builtins ? langVersion and builtins.lessThan 1 builtins.langVersion
+    then builtins.storePath
+    else x: x;
+
   mkFakeDerivation = attrs: outputs:
     let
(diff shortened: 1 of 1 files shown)


domenkozar:
Is this still relevant?

shlevy:
I think so

shlevy:
Doesn't seem wanted.

domenkozar:
@shlevy I'd happily test this, but I fail to understand what it improves exactly.

shlevy:
@domenkozar With this you can import a hydra-built channel in nix and use the packages there as build inputs etc., and they will be fetched from the cache as-needed. Before the channels were only usable with nix-env.

domenkozar:
So this PR answers https://github.com/NixOS/hydra/issues/374

shlevy:
Yep

domenkozar:
@aszlig is there any reason to keep these old style channels alive? Besides that it's easier to set them up initially?

aszlig:
@domenkozar: Well, I guess that's the only reason I can think of...

PR #154, opened 2014-04-13
Ensure Hydra user has a UID

thoughtpolice:
This allows the Hydra module to be used when `mutableUsers = false`;

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -159,6 +159,7 @@ in
         home = baseDir;
         createHome = true;
         useDefaultShell = true;
+        uid = config.ids.uids.hydra;
       };
 
     nix.extraOptions = ''


domenkozar:
Also see https://github.com/peti/hydra-tutorial/pull/4

edolstra:
I don't think this is needed anymore, since `mutableUsers = false` no longer requires specifying a UID.

domenkozar:
It's needed because of https://github.com/NixOS/nixpkgs/issues/14148

edolstra:
Then we should add a check to `update-user-groups.pl`. It should not be necessary in general to assign fixed uids to every service, in fact due to the scarcity of system uids that's not desirable.

domenkozar:
Agreed :)

domenkozar:
Closing based on that consensus.

PR #160, opened 2014-04-30
hydra-module: add config.extraEnv

thoughtpolice:
This makes it easy to set environment variables for the Hydra server
(for example, your configuration.nix can use readFile to read an API
token to upload build results somewhere).

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -19,7 +19,7 @@ let
     { NIX_REMOTE = "daemon";
       OPENSSL_X509_CERT_FILE = "/etc/ssl/certs/ca-bundle.crt";
       GIT_SSL_CAINFO = "/etc/ssl/certs/ca-bundle.crt";
-    } // hydraEnv;
+    } // hydraEnv // cfg.extraEnv;
 
   serverEnv = env //
     { HYDRA_TRACKER = cfg.tracker;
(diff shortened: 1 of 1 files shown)


MaxGabriel:
Was this meant to work for all the environment variables hydra uses? I tried using it like so:

```nix
extraEnv = { NIX_PREFETCH_GIT_LEAVE_DOT_GIT = "1"; };
```

But didn't see any changes. Whereas copying postInstall and adding it the environment variable there worked:

```
            postInstall = ''
              mkdir -p $out/nix-support
              for i in $out/bin/*; do
                  read -n 4 chars < $i
                  if [[ $chars =~ ELF ]]; then continue; fi
                  wrapProgram $i \
                      --prefix PERL5LIB ':' $out/libexec/hydra/lib:$PERL5LIB \
                      --prefix PATH ':' $out/bin:$hydraPath \
                      --set HYDRA_RELEASE ${_oldAttrs.version} \
                      --set HYDRA_HOME $out/libexec/hydra \
                      --set NIX_RELEASE ${pkgs.nix.name or "unknown"} \
                      --set NIX_PREFETCH_GIT_LEAVE_DOT_GIT "1"
              done
            '';
```

Wasn't sure if I should report this as a bug.

PR #198, opened 2015-02-21
Parallel bzip2

aycanirican:


hydra-module.nix
```diff
@@ -162,6 +162,7 @@ in
     users.extraUsers.hydra =
       { description = "Hydra";
         home = baseDir;
+        uid = 121;
```
edolstra: This should not be here.

edolstra:
Hm, this is a nice feature, but the number of cores to be used should be configurable. Otherwise a single request can put a huge load on the server (namely equal to the number of cores).

aycanirican:
@edolstra I just tried to implement it with a default number of threads to 1 as in bzip2's default case.

src/lib/Hydra/View/NixNAR.pm
```diff
@@ -12,8 +12,9 @@ sub process {
     $c->response->content_type('application/x-nix-archive'); # !!! check MIME type
 
     my $fh = new IO::Handle;
+    my $numThreads = $c->config->{'compress_num_threads'};
```
edolstra: There should also be a default here (`$c->config->{'compress_num_threads'} // 1`), otherwise it won't work for people who don't use `hydra-module.nix`.

domenkozar:
@aycanirican can you rebase?

aycanirican:
@domenkozar rebased.

domenkozar:
@aycanirican hmm, it's still showing 250+ commits.

PR #203, opened 2015-03-31
Allow to specify channel attribute path in jobsets.

aszlig:
This should make channels actually useful the same way as it is done for the official Nix(OS) channels, but without the need for [external scripts](https://github.com/NixOS/nixos-channel-scripts) or [rewriting-proxies like this](https://gist.github.com/aszlig/6948b38f3807d42da51b#file-webserver-nix-L26-L76) to create them.

When creating or editing jobsets, there is now a new channel field, where you can specify the attribute path leading to the Nix expression which then is delivered as the channel tarball.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -208,6 +208,36 @@ in rec {
         '';
   });
 
+  tests.channels = genAttrs' (system: let
+    testLib = import <nixpkgs/nixos/lib/testing.nix> { inherit system; };
+    hydra = builtins.getAttr system build;
+    perlLibs = pkgs.lib.concatStringsSep ":" [
+      "${hydra}/libexec/hydra/lib"
+      "${hydra.perlDeps}/lib/perl5/site_perl"
+    ];
+    runPerl = script:
+      "PERL5LIB=\\\"${perlLibs}\\\" NIX_REMOTE=daemon perl ${script}";
+  in testLib.simpleTest {
+    name = "channels";
+    machine = {
+      imports = [ (hydraServer hydra) ];
+      virtualisation.writableStore = true;
+      nix.useChroot = false;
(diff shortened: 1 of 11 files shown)


edolstra:
"Channel job attribute" seems a misnomer to me. If I understand it correctly, it's the name of the job that produces the channel Nix expressions, right? So "channel job name" would be more descriptive.

Also, why not just require the channel job to generate nixexprs.tar? Then Hydra doesn't have to tar/compress the Nix expressions. In fact, if the channel job announces its nixexprs.tar via $out/nix-support/hydra-build-products, it wouldn't even be necessary to specify the channel job in the web interface: Hydra could just find it automatically by looking for a build product with the right type.

BTW, note that jobset channels are currently kind of fundamentally broken, because they can contain builds from different evaluations. There is some code in JobsetEval.pm to provide a consistent channel, which Jobset.pm should probably redirect to.

aszlig:
Hm, you're right, channel job name is a better description.

As for the tarball: I had that already in the proxy, which redirects to the corresponding tarball job. But speaking of build products, I've got a better idea:

What about not specifying the channel job name altogether but specify the channel as a **build product**. That way it's not only easy to have more than one channel, but also move much of the configuration stuff to the nix expressions themselves. What do you think?

edolstra:
Sounds good. Regarding making a channel update dependent on a set of builds, we already have aggregate builds for that. So the channel could just be a build product of the aggregate build. Then channel updates will depend on the members of the aggregate.

aszlig:
@edolstra: Exactly that is what I have in mind :-)

aszlig:
Closed in favor of #211.

PR #211, opened 2015-04-27
Allow custom channel expressions.

aszlig:
This replaces #203 and not only ships custom Nix expressions but also allows to require certain jobs to succeed for a channel to upgrade.

So far, we already had a dispatcher for `/channel/latest`, which I rewrote but kept so far. Question here would be, whether we actually want those generic channels or replace it directly by the new approach.

We use `meta.isHydraChannel` to distinguish a regular build from a channel build, which is also stored in the DB and used to hide channel jobset from normal jobsets. Channel expressions are referenced by putting a `channel - $path` into `$out/nix-support/hydra-build-products`, where the second column is `-` for now until we have a need a way to name the expression tarball.

The main reason why I've stored it into the DB is to allow the channel to persist regardless of whether aggregate constituents failed or not.

Items left to implement:

 - [x] Implement "Channels" tab for jobsets.
 - [x] Deliver custom Nix expressions.
 - [ ] Replace `channel.png` with a better icon.
 - [ ] Add `channel` function to `releaseTools` ([WIP implementation](https://github.com/openlab-aux/vuizvui/blob/master/pkgs/build-support/channel.nix)).
 - [ ] Document `channel` function in release-notes for NixOS `15.06`.
 - [ ] Move channels to an extra tab in evaluation view.
 - [ ] Highlight failing/succeeding channels in jobset evaluations.


edolstra:
Looks good to me.

domenkozar:
@aszlig status on this? Looks really useful.

goodwillcoding:
I started writing my own, terrible version of this before I found this ...
this looks great ... any chance it can make it into hydra?

aszlig:
I already have this [running on my Hydra instance](https://headcounter.org/hydra/jobset/openlab/vuizvui#tabs-channels). It works fine but it needs a bit of polishing and I need to rebase it against current master.

Will do the rebase on Monday if I get to it.

aszlig:
Rebased against current master, sorry for the delay.

edolstra:
@aszlig Is this intended to be merged already? (Asking since only 2 out of 7 of your todo items are checked.)

aszlig:
@edolstra: The outstanding issues currently only concern the UI, so it can be merged. But I haven't yet updated my Hydra to post-`build-ng` (and thus the rebase of the commits here), so it's untested.

goodwillcoding:
@edolstra do you think this will be merged any time soon?

src/sql/upgrade-42.sql
```diff
@@ -0,0 +1 @@
+alter table Builds add column isChannel integer not null default 0;
```
edolstra: This has `not null` but the definition in `hydra.sql` doesn't.

edolstra: BTW we should probably use `boolean` for these columns, since we don't support SQLite anymore. The only problem is that SQLite is still used to generate the DBIx::Class files, but we can just start a temporary PostgreSQL instance for that.

src/lib/Hydra/Controller/Channel.pm
```diff
@@ -0,0 +1,83 @@
...
+        { 'eval.hasnewbuilds' => 1
+        , project => $projectName
+        , jobset => $jobsetName
+        , job => $channelName
+        , buildstatus => 0
+        },
+        { rows => 1, order_by => "eval.id desc"
+        , join => { jobsetevalmembers => 'eval' }
+        }
+    );
```
edolstra: Shouldn't this filter out unfinished evals? Otherwise you can get a channel update whose `nixexprs.tar.xz` contains expressions that haven't finished building yet.

aszlig: Isn't that already handled by the condition on `eval.hasNewBuilds`? On my Hydra instance I get only channel updates on finished evals, so it seems to work fine.

src/sql/upgrade-42.sql
aszlig: Fixed the comment about not nulls in 92e85f05e2879957ee395ce17428ae965d5e53f4.

aszlig: As for the second comment, I think that's better to do in a different branch/pull request, because we have quite a lot of integer columns that are used as booleans.

src/lib/Hydra/Controller/Channel.pm
edolstra: No, `hasNewBuilds` only indicates whether the eval is different from the last one. It's set by `hydra-evaluator`, so it can be set even when none of the new builds have been performed.

src/sql/upgrade-42.sql
edolstra: Yeah, I agree.

src/lib/Hydra/Controller/Channel.pm
aszlig: Okay, looked into that in detail... my original idea was that a particular channel doesn't need to be part of a finished evaluation as long as all builds needed for that particular channel are finished. For my use-case, this is what I want it to be, but if you prefer only fully built evals, but if you want I could add `finished => 1` to the conditions.

edolstra: So I guess you have all the builds for the channel as an explicit dependency of the channel job?

domenkozar: One thing to note here (I haven't had time to look into source), with current channels we can "travel back in time" if one build is restarted, as the job is then considered not finished and it goes back and uses older finished job as latests. Has this implementation gone around that?

edolstra: Channels going back in time was a problem of the channel mirror scripts which live outside of Hydra (which is fixed now). I imagine this will only go back in time if you restart the channel build (but you presumably wouldn't have any reason to do that, since it succeeded).

PR #216, opened 2015-05-17
Allow running on i686-linux

Warbo:
Hydra won't build via the instructions at https://nixos.org/wiki/Installing_hydra_as_nixos_module#Letting_nix_handle_the_git_repository without this change.

Specifically where it says "# or i686-linux if appropriate", there is an attribute not found error for i686.

With this change, and after setting up the "hydra" postgres user's password manually, I have Hydra up and running on my i686 laptop without issue.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -6,7 +6,7 @@ let
 
   pkgs = import <nixpkgs> {};
 
-  genAttrs' = pkgs.lib.genAttrs [ "x86_64-linux" ];
+  genAttrs' = pkgs.lib.genAttrs [ "x86_64-linux" "i686-linux" ];
 
   hydraServer = hydraPkg:
     { config, pkgs, ... }:


peti:
@shlevy, won't this change creates additional `i686` build jobs on `hydra.nixos.org` (which Eelco had previously disabled in 0fd6bb4e3f89649e8ce08e7bd5f44e95c2d5fb52)?

rbvermaa:
@peti @edolstra mentioned that nobody cares about i686-linux, which it turns out to be wrong, as some people do care it seems. Shouldn't be a big issue to keep it right? 

@edolstra why was it removed again? Was something not building at the time?

edolstra:
It's mostly that I don't want to have to maintain Hydra on more platforms than necessary (i.e. x86_64).

Warbo:
@peti I certainly don't want to burden anyone. If this causes technical complications, eg. extra builds on hydra.nixos.org, then I have no problem with reverting it. In which case, perhaps a comment should be added next to that line, since I could imagine others sending similar patches in the future; eg. `# Enable other architectures using this list, but upstream only supports x86_64`

@edolstra If there aren't any such complications, perhaps it would be better to add a statement to the documentation that only x86_64 is "officially" supported, then the code could be left in a "self service" state for those of us on other systems?

edolstra:
@Warbo I'm fine with building it. I probably disabled it at a time when it was failing.

PR #217, opened 2015-05-29
hydra-module: Don't assume hydra is running on localhost.

shlevy:
In the long run we should add retry logic to hydra-init...


shlevy:
@edolstra @rbvermaa

edolstra:
This was already done in 8ea05cc7f2349f18d4188fba9386c39bc30aadf0.

shlevy:
When is that going to be merged?

PR #226, opened 2015-07-31
buildInputToString: Pass along the input's type and urr

shlevy:


copumpkin:
urr!

shlevy:
Hey, I got it right in the code!

copumpkin:
urr, okay :smile:

edolstra:
Hm, shouldn't we set `type` to `derivation` to allow the input to be used as a derivation?

shlevy:
No strong opinions there, if we want to use `inputType` instead I can make that change.

edolstra:
Yeah, having a separate `inputType` attribute is probably better.

shlevy:
OK, updated

PR #228, opened 2015-08-03
All args to matchless lambda

shlevy:
If the hydra expression is a lambda with no formals, just pass in all args.

This makes it possible to have a shared `release.nix` for multiple projects with similar structures that vary in their inputs.

### DIFF
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -172,11 +172,31 @@ static void findJobsWrapped(EvalState & state, JSONObject & top,
     }
 
     else if (v.type == tLambda && v.lambda.fun->matchAttrs) {
-        Bindings & tmp(*state.allocBindings(0));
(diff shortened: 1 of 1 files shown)


edolstra:
I'm not sure if this is a good idea, because it increases the gap between hydra-eval-jobs and nix-build/nix-instantiate. It should be possible to test a Hydra job locally by doing "nix-build release.nix -A jobname", but if hydra-eval-jobs has special magic features, that no longer works. In fact, I'd like to get rid of hydra-eval-jobs at some point, since really the only reason it existed in the first place is to support input alternatives, which are now nearly removed.

Maybe this could be done only for lambdas that match an attribute set, e.g. `inputs @ { ... }`? That could probably be implemented consistently for hydra-eval-jobs and nix-instantiate. Or another possibility: have hydra-evaluator pass the complete set of inputs as an argument `inputs` (i.e. `--arg inputs '{ input1 = ...; inputs2 = ...; ... }'`). Then you can just write `{ inputs }: ...` in `release.nix`.

shlevy:
Closing in favor of #229

PR #229, opened 2015-08-04
hydra-eval-jobs: Pass all inputs as 'inputs' arg.

shlevy:
If there is no input named 'inputs', hydra-eval-jobs now passes in a set
of lists, where each attribute corresponds to an input defined in the
jobset specification and each list element is a different input alt, as
an argument named 'inputs'.

Among other things, this allows for generic hydra expressions to be
shared amongst projects with similar structures but different sets of
specific inputs.


edolstra:
Can you drop the "lists" part and only include the first input alt? Input alts are on the way out (you can't even configure them in the web interface anymore.)

shlevy:
@edolstra done

PR #231, opened 2015-08-10
Pass along drvPath and outputName for inputs that are previous builds.

shlevy:
This allows importing the .drv and getting the same store paths as if the
input had been passed in as nix expressions defining a proper derivation.

### DIFF
--- a/src/lib/Hydra/Helper/AddBuilds.pm
+++ b/src/lib/Hydra/Helper/AddBuilds.pm
@@ -114,7 +114,7 @@ sub fetchInputBuild {
         , outputName => $mainOutput->name
         };
     if (isValidPath($prevBuild->drvPath)) {
-        $input->{drvPath} = $prevBuild->drvPath;
+        $result->{drvPath} = $prevBuild->drvPath;
     }
 
     return $result;
--- a/src/lib/Hydra/Helper/AddBuilds.pm
+++ b/src/lib/Hydra/Helper/AddBuilds.pm
@@ -288,7 +288,7 @@ sub buildInputToString {
             (defined $input->{shortRev} ? "; shortRev = \"" . $input->{shortRev} . "\"" : "") .
             (defined $input->{version} ? "; version = \"" . $input->{version} . "\"" : "") .
             (defined $input->{outputName} ? "; outputName = \"" . $input->{outputName} . "\"" : "") .
-            (defined $input->{drvPath} ? "; drvPath = " . $input->{drvPath} . "" : "") .
+            (defined $input->{drvPath} ? "; drvPath = builtins.storePath " . $input->{drvPath} . "" : "") .
             ";}";
     }
     return $result;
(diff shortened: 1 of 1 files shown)


src/lib/Hydra/Helper/AddBuilds.pm
```diff
@@ -105,10 +105,14 @@ sub fetchInputBuild {
...
 
+    my $mainOutput = getMainOutput($prevBuild);
+
     return
-        { storePath => getMainOutput($prevBuild)->path
+        { storePath => $mainOutput->path
         , id => $prevBuild->id
         , version => $version
+        , outputName => $mainOutput->name
+        , drvPath => $prevBuild->drvPath
```
edolstra: This could have been garbage-collected, right? (I think Hydra only keeps .drvs as roots until the build has finished.)

src/lib/Hydra/Helper/AddBuilds.pm
```diff
@@ -279,6 +283,8 @@ sub buildInputToString {
             (defined $input->{gitTag} ? "; gitTag = \"" . $input->{gitTag} . "\"" : "") .
             (defined $input->{shortRev} ? "; shortRev = \"" . $input->{shortRev} . "\"" : "") .
             (defined $input->{version} ? "; version = \"" . $input->{version} . "\"" : "") .
+            (defined $input->{outputName} ? "; outputName = \"" . $input->{outputName} . "\"" : "") .
+            (defined $input->{drvPath} ? "; drvPath = " . $input->{drvPath} . "" : "") .
```
edolstra: This should be `drvPath = builtins.storePath ...`, I think.

src/lib/Hydra/Helper/AddBuilds.pm
shlevy: Oh, if that's the case this is less useful. Previously I've been adding a build that generates a nix expression that imports the drv (taken as an input source via `unsafeDiscardOutputDependency`) as part of the jobset definition, which doesn't have this problem, but this would have been nicer. Any objection to keeping the drv as a root as long as the build is?

src/lib/Hydra/Helper/AddBuilds.pm
shlevy: Will fix if above comment is addressed positively.

src/lib/Hydra/Helper/AddBuilds.pm
edolstra: On second thought, the derivations *are* kept if `gc-keep-derivations` is set in `nix.conf`, which happens to be the default. So this should be fine. Maybe you can make passing `drvPath` conditional on `isValidPath($prevBuild->drvPath)`.

shlevy: Does `isValidPath` add a temp root?

edolstra: No.

shlevy:
Fixed

PR #232, opened 2015-08-10
Allow . in input names

shlevy:
With the new inputs argument, this can now be useful


shlevy:
@edolstra Any objections? Will merge in a few days if not.

edolstra:
Allowing dots doesn't sound like a good idea due to the ambiguity with attribute names.

PR #249, opened 2015-10-31
Fix legacy channel URLs

aszlig:
Regression introduced by 1fdc258de0d7f97af428de34e8be019fca3db251.

The commit introduced a channel/custom PathPart which uses the new custom channel expressions, but I forgot to remove CaptureArgs, so the URL really is channel/latest/ignored-value.

Can't test this right now, so I'd appreciate if someone could test it :-)


edolstra:
It was almost right, it actually needs to have `CaptureArgs(0)`, otherwise other actions can't be chained on top of it.

PR #253, opened 2015-11-12
Add help flag to nix-prefetch-git

andrewrynhard:
Believe it or not, this flag would have saved me a good half hour of my day.

### DIFF
--- a/src/script/nix-prefetch-git
+++ b/src/script/nix-prefetch-git
@@ -20,6 +20,21 @@ else
   leaveDotGit=true
 fi
 
+usage(){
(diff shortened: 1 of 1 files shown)


edolstra:
Hm, I think it would be better to remove nix-prefetch-* from the Hydra repository and use the `nix-prefetch-scripts` package from Nixpkgs instead.

andrewrynhard:
Ah yes. That is my mistake. I forgot I installed that package. I will make a PR over there.

domenkozar:
Let's close this.

PR #264, opened 2016-01-08
fix a sendEmail problem

rick68:
I want to receive a mail when hydra have done something.
`sendEmail()` method does not tag `Message-ID` in header, that can cause to gmail reject the email.


domenkozar:
According to https://en.wikipedia.org/wiki/Message-ID, the message-id should be unique, that's not the case for what you set it here.

grahamc:
I closed it because I'm pretty sure this is not the right solution to the problem.

PR #269, opened 2016-02-06
statsd: add a chance to set hostname and port in hydra.conf

matejc:
For those that have statsd on other host than localhost. This pull request adds to hydra.conf two extra optional entries named `statsd_host` and `statsd_port`

These lines are the first that I have written in perl, I hope that it is acceptable :)

I have tested this in https://github.com/matejc/hydra-stack (Hydra inside docker with extra builders for distributed builds)


grahamc:
I'm closing this as superseded by #886, which includes tests and docs, but has a slightly different configuration format.

PR #273, opened 2016-02-29
Add slack plugin.

shlevy:
Respects `<slack>` blocks in the hydra config, with attributes:

* jobs: a regexp matching the job name (in the format project:jobset:job)
* url: The URL to a slack incoming webhook
* force: If true, always send messages. Otherwise, only when the build status changes

Multiple `<slack>` blocks are allowed

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -20,6 +20,7 @@ let
     { NIX_REMOTE = "daemon";
       SSL_CERT_FILE = "/etc/ssl/certs/ca-certificates.crt";
       OPENSSL_X509_CERT_FILE = "/etc/ssl/certs/ca-certificates.crt"; # FIXME: remove on NixOS >= 15.07
+      PERL_LWP_SSL_CA_FILE = "/etc/ssl/certs/ca-certificates.crt";
       PGPASSFILE = "${baseDir}/pgpass";
     } // hydraEnv // cfg.extraEnv;
 
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -20,7 +20,6 @@ let
     { NIX_REMOTE = "daemon";
       SSL_CERT_FILE = "/etc/ssl/certs/ca-certificates.crt";
       OPENSSL_X509_CERT_FILE = "/etc/ssl/certs/ca-certificates.crt"; # FIXME: remove on NixOS >= 15.07
-      PERL_LWP_SSL_CA_FILE = "/etc/ssl/certs/ca-certificates.crt";
       PGPASSFILE = "${baseDir}/pgpass";
     } // hydraEnv // cfg.extraEnv;
 
--- a/src/lib/Hydra/Plugin/SlackNotification.pm
+++ b/src/lib/Hydra/Plugin/SlackNotification.pm
@@ -48,10 +48,10 @@ sub buildFinished {
 
         my $imgBase = "http://hydra.nixos.org";
         my $img =
-            $build->buildstatus == 0 ? "$imgBase/static/images/checkmark_16.png" :
-            $build->buildstatus == 2 ? "$imgBase/static/images/dependency_16.png" :
-            $build->buildstatus == 4 ? "$imgBase/static/images/cancelled_16.png" :
-            "$imgBase/static/images/error_16.png";
+            $build->buildstatus == 0 ? "$imgBase/static/images/checkmark_256.png" :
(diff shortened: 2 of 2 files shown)


shlevy:
@rickynils

rickynils:
I've tested this and it works perfectly. I rebased it on 993647d1e3b43f1f9b7dc2ebce889b475d156bb9, though, since I couldn't bump nix to latest in my setup.

@shlevy A bonus feature would be if the webhook url could be read from a file at runtime, to not leak any secrets to the nix store.

src/lib/Hydra/Plugin/SlackNotification.pm
```diff
@@ -0,0 +1,93 @@
...
+    my ($authors, $nrCommits) = getResponsibleAuthors($build, $self->{plugins});
+
+    # Send a message to each room.
+    foreach my $url (keys %channels) {
+        my $channel = $channels{$url};
+        my @deps = grep { $_->id != $build->id } @{$channel->{builds}};
+
+        my $imgBase = "http://hydra.nixos.org";
+        my $img =
+            $build->buildstatus == 0 ? "$imgBase/static/images/checkmark_16.png" :
```
rickynils: These images looks very bad since they are scaled up a lot in the slack interface, especially on high-DPI screens. You should probably go for the 256 variants instead.

shlevy:
@rickynils `Config::General` supports includes: http://search.cpan.org/dist/Config-General/General.pm#INCLUDES

rickynils:
@shlevy Ah, thanks, I'll try that out

src/lib/Hydra/Plugin/SlackNotification.pm
shlevy: Pushed, except cancelled only has 128. @edolstra, why is that?

hydra-module.nix
```diff
@@ -20,6 +20,7 @@ let
     { NIX_REMOTE = "daemon";
       SSL_CERT_FILE = "/etc/ssl/certs/ca-certificates.crt";
       OPENSSL_X509_CERT_FILE = "/etc/ssl/certs/ca-certificates.crt"; # FIXME: remove on NixOS >= 15.07
+      PERL_LWP_SSL_CA_FILE = "/etc/ssl/certs/ca-certificates.crt";
```
edolstra: Hm, I've never needed to set this variable. And I was just using LWP with https today.

shlevy: I got an error without it, no idea why it would be any different.

shlevy:
@edolstra @rbvermaa any objections?

edolstra:
Looks good except for the `PERL_LWP_SSL_CA_FILE` thing.

shlevy:
> Feb 29 13:18:29 shlevy-laptop hydra-queue-runner[12208]: sending to https://hooks.slack.com/services/hook-url-elided
Feb 29 13:18:29 shlevy-laptop hydra-queue-runner[12208]: 500 Can't verify SSL peers without knowing which Certificate Authorities to trust: Can't verify SSL peers without kno
Feb 29 13:18:29 shlevy-laptop hydra-queue-runner[12208]: This problem can be fixed by either setting the PERL_LWP_SSL_CA_FILE
Feb 29 13:18:29 shlevy-laptop hydra-queue-runner[12208]: envirionment variable or by installing the Mozilla::CA module.
Feb 29 13:18:29 shlevy-laptop hydra-queue-runner[12208]: To disable verification of SSL peers set the PERL_LWP_SSL_VERIFY_HOSTNAME
Feb 29 13:18:29 shlevy-laptop hydra-queue-runner[12208]: envirionment variable to 0.  If you do this you can't be sure that you
Feb 29 13:18:29 shlevy-laptop hydra-queue-runner[12208]: communicate with the expected peer.

edolstra:
This works for me:
```
$ nix-shell -p perl perlPackages.LWP perlPackages.LWPProtocolHttps
$ perl -e 'use LWP::UserAgent; my $ua = LWP::UserAgent->new; print $ua->get("https://nixos.org")->decoded_content;'
```
That's on 15.09 with `SSL_CERT_FILE` set btw. This may be a 16.03 regression.

shlevy:
Fails for me on `16.03.git.40c586bM (Emu)` (the M is unrelated to this).

edolstra:
Looks like we lost a patch in LWPProtocolHttps. I'll put it back.

PR #280, opened 2016-03-12
Add a plugin to interact with the github status API.

shlevy:
Mutliple `<githubstatus>` sections are possible:
* jobs: regexp for jobs to match
* inputs: the input which corresponds to the github repo/rev whose
  status we want to report. Can be repeated
* authorization: Verbatim contents of the Authorization header. See
  https://developer.github.com/v3/#authentication.

### DIFF
--- a/src/hydra-queue-runner/builder.cc
+++ b/src/hydra-queue-runner/builder.cc
@@ -84,9 +84,16 @@ State::StepResult State::doBuildStep(nix::ref<Store> destStore, Step::ptr step,
             return sMaybeCancelled;
         }
 
-        for (auto build2 : dependents)
-            if (build2->drvPath == step->drvPath) { build = build2; break; }
-
+        for (auto build2 : dependents) {
+            if (build2->drvPath == step->drvPath) {
+              build = build2;
+              {
+                  auto notificationSenderQueue_(notificationSenderQueue.lock());
+                  notificationSenderQueue_->push(NotificationItem{NotificationItem::Type::Started, build->id});
+              }
+              notificationSenderWakeup.notify_one();
+            }
+        }
         if (!build) build = *dependents.begin();
 
         printMsg(lvlInfo, format("performing step ‘%1%’ on ‘%2%’ (needed by build %3% and %4% others)")
@@ -249,7 +256,7 @@ State::StepResult State::doBuildStep(nix::ref<Store> destStore, Step::ptr step,
         for (auto id : buildIDs) {
             {
                 auto notificationSenderQueue_(notificationSenderQueue.lock());
-                notificationSenderQueue_->push(NotificationItem(id, std::vector<BuildID>()));
+                notificationSenderQueue_->push(NotificationItem{NotificationItem::Type::Finished, id});
             }
             notificationSenderWakeup.notify_one();
         }
@@ -369,7 +376,7 @@ State::StepResult State::doBuildStep(nix::ref<Store> destStore, Step::ptr step,
         /* Send notification about this build and its dependents. */
         {
             auto notificationSenderQueue_(notificationSenderQueue.lock());
-            notificationSenderQueue_->push(NotificationItem(build->id, dependentIDs));
+            notificationSenderQueue_->push(NotificationItem{NotificationItem::Type::Finished, build->id, dependentIDs});
         }
         notificationSenderWakeup.notify_one();
 
--- a/src/hydra-queue-runner/hydra-queue-runner.cc
+++ b/src/hydra-queue-runner/hydra-queue-runner.cc
@@ -482,11 +482,11 @@ void State::notificationSender()
                 notificationSenderQueue_->pop();
             }
 
-            printMsg(lvlChatty, format("sending notification about build %1%") % item.first);
+            printMsg(lvlChatty, format("sending notification about build %1%") % item.id);
 
             Pid pid = startProcess([&]() {
-                Strings argv({"hydra-notify", "build", std::to_string(item.first)});
-                for (auto id : item.second)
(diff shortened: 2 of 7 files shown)


shlevy:
@rickynils

Ericson2314:
Excellent! This should bring us one step closer to building new PRs with hydra.nixos.org?

shlevy:
Yep. If this and #277 are merged, I have in mind how the last piece should look.

Ericson2314:
:)

domenkozar:
This looks really useful - could we get some docs for the plugin? @edolstra should we start a section in hydra manual for plugins?

shlevy:
Happy to write up docs if @edolstra wants this merged

rickynils:
Nice work!

I might have missed something, but looking at the code, it seems the status context is hard-coded to `continuous-integration/hydra`?

So if we have multiple jobs running for the same set of inputs (which is the usual situation), will we not get multiple status reports with identical contexts (but with possibly different states) for the same repo/rev? I think the context actually should be the `project:jobset:job` id, possibly with some additional Hydra-specific prefix. Or the context could be configured manually in the `<githubstatus>` block, but then we would have to have one block per job.

That way, you could configure a GitHub protected branch to allow only commits that have passed a specific set of Hydra jobs. With the current implementation, you could only really decide if a commit has passed _any_ Hydra job, right?

shlevy:
Ah, I had in mind that the `jobs` regexp would be specific enough to ensure only a single build, but you're right that that makes more sense. Any reason to prefer `project:jobset:job` instead of `buildId`?

rickynils:
`buildId` is unique per build, not per job, isn't it? That wouldn't make sense for GitHub statuses, since it would change all the time. So you couldn't refer to it when you configure a protected branch in GitHub.

Using a specific `jobs` expression, the behavior I want could be encoded, but you would still be limited to a single status (`continuous-integration/hydra`). You wouldn't be able to have more fine-grained checks on the GitHub side. Maybe it wouldn't be such a severe limitation in practice, but it doesn't feel right.

An acceptable alternative to sending `project:jobset:job` would be to be able to configure the status name in the `<githubstatus>` block. You would need one block per status then, but I guess that is alright.

shlevy:
There doesn't seem to be any drawback to `project:jobset:job`, though if we don't have `buildId` then there might be a problem where an update of some other input that causes a rebuild (say, an update of nixpkgs that changes some dep of the project you're building) could cause multiple updates with the same context to the same commit.

rickynils:
Having `buildId` in the status context would make the status update useless, just saying :)

However, you are entirely correct that this feature doesn't blend well with multiple inputs. A Hydra job is usually not building/testing a single commit, it is building specific combinations of commits (from different repos). So, in the general case this plugin is probably not meaningful... And the status report should maybe be sent from a job-specific post-build hook instead. Which is being discussed in https://github.com/NixOS/hydra/pull/281

shlevy:
Hmm I'm still not clear on why the build ID would be useless. So the context changes from one build to the next, *within* the same build it will be the same (so first you'll see a 'pending' message, then a 'success' or 'failure' one). Can you explain more the workflow that requires the context be the same for all builds of the same job?

shlevy:
Oh, I guess it's the github protected branch feature, I see.

shlevy:
Well, if you're using protected branches, you probably want all the other inputs pegged at specific revisions anyway, right?

rickynils:
Yeah, that's right. And in that case this plugin would actually work just fine (if adjusted to send something job-constant as status context).

shlevy:
OK, so for manually reviewed PRs we want the build number in the context (so that we can see how the change fares against all possible input changes) and for automatic use cases like protected branches where we can't use a regexp to identify the relevant context we drop the build number. So does a config switch to go from `continuous-integration/hydra-project:jobset:job:build` to `continuous-integration/hydra-project:jobset:job` work for you?

rickynils:
@shlevy Yes, I think such a switch would do the trick.

(Btw, I've e-mailed you more details on my use case that I didn't have the energy to clean up for public consumption)

domenkozar:
@shlevy if you address the status context, I'll test this and report back.

shlevy:
Rebased and added an `excludeBuildFromContext` setting. Compare https://github.com/shlevy/hydra-github-status-test/pull/2 and https://github.com/shlevy/hydra-github-status-test/pull/3

shlevy:
@edolstra ping
@rickynils FYI

Ericson2314:
That demo is mouthwatering!

domenkozar:
@shlevy can you post an example configuration (minus correct credentials)?

shlevy:
```
<githubstatus>
  jobs = test:pr:build
  inputs = src
  authorization = Basic notgivingyoumypasswordosorry
  excludeBuildFromContext = 1
</githubstatus>
```
`jobs` can be a regexp. `inputs` can be specified multiple times. This example matches only the `build` job of the `pr` jobset of the `test` project, sets a status associated with the revision of the `src` input when the build starts and when it finishes, and doesn't include the build's ID in the context of the status.

edolstra:
`excludeBuildFromContext` should be something like `display_build_id`. Otherwise the user has to guess what it means to exclude a build from a "context" (what context?).

shlevy:
`context` has a specific meaning for github statuses, and including the build ID should be the default IMO. And either way the build ID is displayed, it's just that it's not included in, well, the context :grin: . If you still think `display_build_id` is best I will switch to that.

robinp:
To clarify: `inputs` should be set to the jobset input name which triggers the source checkout (of type `Git checkout`). I was a bit confused, thinking it should be set to a path inside the repo.

PR #281, opened 2016-03-12
Add a plugin to commit changes to a git repo.

shlevy:
First matching `<gitcommit>` block is used:
- jobs: Jobs regexp
- repo: git repo
- branch: git branch

If a build of a matching job creates
$out/build-support/hydra-git-commit.patch, that patch is applied to
the specified branch, committed, and pushed back.

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -284,7 +284,7 @@ in
       { wantedBy = [ "multi-user.target" ];
         requires = [ "hydra-init.service" ];
         after = [ "hydra-init.service" "network.target" ];
-        path = [ pkgs.nettools pkgs.ssmtp ];
+        path = [ pkgs.nettools pkgs.ssmtp pkgs.patch ];
         environment = env // {
           PGPASSFILE = "${baseDir}/pgpass-queue-runner"; # grrr
           IN_SYSTEMD = "1"; # to get log severity levels
--- a/release.nix
+++ b/release.nix
@@ -128,6 +128,7 @@ rec {
             FileSlurp
             IOCompress
             IPCRun
+            IPCSystemSimple
             JSONXS
             LWP
             LWPProtocolHttps
--- a/src/lib/Hydra/Plugin/GitCommit.pm
+++ b/src/lib/Hydra/Plugin/GitCommit.pm
@@ -1,0 +1,46 @@
+package Hydra::Plugin::GitCommit;
+
+use strict;
+use parent 'Hydra::Plugin';
+use Cwd;
+use Hydra::Helper::CatalystUtils;
+use autodie qw( system );
+
+sub buildFinished {
+    my ($self, $build, $dependents) = @_;
+    return unless $build->buildstatus == 0;
+    my $cfg = $self->{config}->{gitcommit};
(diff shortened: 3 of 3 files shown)


shlevy:
@rickynils

Ericson2314:
Granted, I don't know the details of hydra at all, but I'm a bit skeptical of this approach. I rather have hydra merge the commit with master, then test, then push (each step conditionally on the last succeeding of course). That way, the test run is associated with the exact commit that would be pushed to master.

Also, it would be nice to configure to git credentials hydra uses, analogous to the handling of other tokens, so that hydra's commits can be signed, etc. But this is a nit as that's easily done as follow up work.

shlevy:
This is unrelated to PRs. The specific use case is for projects that have their inputs pegged in a paths file (see e.g. https://github.com/rickynils/nix-path or https://github.com/zalora/deployix/tree/master/nix-exec/sequence-pathspecs), hydra will run tests with the latest version of each dependency and, if everything works, update the paths file.

Ericson2314:
Oh OK. I'd argue however that use-case is more related to PRs than might seem, and is conceptually two jobs. One job is triggered on dependency updates, and creates the commit updating the downstream project. Then the second job tests that commit, along with normal tentative changes / PRs.

That said, the first job in my conception still needs something like this, and the whole point of plugins is that they are optional.

rickynils:
@shlevy Elegant solution!

edolstra:
I think it might be better to have a generic "run shell command" plugin for this sort of thing, e.g.
```
<on-success>
  job = foo:bar:xyzzy
  command = git push production $src_commit_hash:refs/heads/master
</on-success>
```
That way we don't have to hard-code a lot of policy into the plugin.

shlevy:
Hmm, so this morning I had a thought that maybe we could have a plugin that just executes $out/build-support/post-build or whatever, but the `on-success` idea seems so generic as to be basically replicating the `postBuild` functionality, except in shell instead of perl.

rickynils:
A more generic approach is nice, yes. I don't know which of `<on-success>` and Shea's suggestion about `$out/build-support/post-build` that's best. With `<on-success>` it should be possible to implement Shea's suggestion anyway (because I assume the script would have access to the output too?). And in any case, the script would need to have access to info about all inputs (`$src_commit_hash` wouldn't do, since there usually are more than one input).

shlevy:
I still maintain that `on-success` is the equivalent of the already existing plugin mechanism, especially if we make it general enough to access all sorts of relevant info about the build etc. I'm not clear on why "encode policy in an optional plugin" is worse than "encode policy in an optional binary", especially since plugins can be overridden by simply specifying an augmented perl path.

shlevy:
@edolstra Any further thoughts on this?

expipiplus1:
One caveat with putting things in `$out/build-support/post-build` is that these builds now depend on anything this `post-build` script depends upon. It would be possible to have a job which just contains `post-build` and depends on the actual build, but that might lead to clutter.

expipiplus1:
This might also fall victim to #342 unless `KEEP_DOTGIT` is specified.

shlevy:
@edolstra Any further thoughts on this?

PR #286, opened 2016-03-23
hydra-queue-runner: Handle missing nix-support directory

rickynils:
This fixes the "is not a directory"-exceptions that 02190b0fefe7b5739c0eac403264c07d15ae3635 introduced


edolstra:
Thanks. I've fixed the root cause in Nix instead.

PR #292, opened 2016-04-12
Add eval test for specific git revision

shlevy:
Wasn't completely sure how to specify a git input pegged to a specific rev, so I added a test to make sure it was working.


shlevy:
@rickynils Note that this test passes, but fails if e.g. I change the commit message (and thus the commit hash)

PR #307, opened 2016-05-12
Trivial fixes (readme, comment, missing dependency in release.nix)

groxxda:
Just 3 one-line changes so I thought they can live in one PR.

Please let me know if you want seperate PRs (either for these commit or in the future)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 To start hacking on Hydra, run:
 
+    $ bootstrap
     $ nix-shell
     $ ./configure $configureFlags --prefix=/opt/hydra
     $ make
--- a/release.nix
+++ b/release.nix
@@ -118,7 +118,7 @@ rec {
         [ makeWrapper autoconf automake libtool unzip nukeReferences pkgconfig sqlite libpqxx
           gitAndTools.topGit mercurial darcs subversion bazaar openssl bzip2 libxslt
           guile # optional, for Guile + Guix support
-          perlDeps perl
+          perlDeps perl nix
           postgresql92 # for running the tests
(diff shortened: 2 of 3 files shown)


domenkozar:
cc @edolstra

README.md
```diff
@@ -1,5 +1,6 @@
 To start hacking on Hydra, run:
 
+    $ bootstrap
```
domenkozar: It would be better to run `$configurePhase`, because this is part of `preConfigure` in release.nix

release.nix
```diff
@@ -118,7 +118,7 @@ rec {
         [ makeWrapper autoconf automake libtool unzip nukeReferences pkgconfig sqlite libpqxx
           gitAndTools.topGit mercurial darcs subversion bazaar openssl bzip2 libxslt
           guile # optional, for Guile + Guix support
-          perlDeps perl
+          perlDeps perl nix
```
domenkozar: Why is Nix needed?

groxxda: I probably also ran into #319 
I think new versions of hydra will need nix anyway since s3binarycachestore was moved to nix: https://github.com/NixOS/nix/commit/d155d8015578c43953e4a9d1867e49c0b71534d7

README.md
groxxda: Good point! Should we call bootstrap from release.nix or remove the bootstrap script? I don't like having the autoreconf call duplicated :)

domenkozar: Best to call it from `release.nix`

domenkozar:
Merging as this is an improvement already.

PR #308, opened 2016-05-12
WIP: catch up with nix changes

groxxda:
I think the first 3 are pretty obvious.

Usage of store-api instead of direct instantiation of the different stores should also be what we want here.

Replacing the store_mode and binary_cache_{dir,s3_bucket,private_key_file} parameters with a single URI parameter gives us the freedom of leveraging any store that's supported by nix without additional changes to hydra. I'm just not sure about the deprecation: whether we should handle the old parameters gracefully (for example convert them to an URI and inform the user that he should change his configuration accordingly). But since hydra is far from stable the breaking change could be ok. I'll update the documentation if this is accepted as-is.

Last commit (wip): I suggest removing the dumpStats stuff from queue-runner, putting it inside the Stores and using the visitor-pattern to collect it to get rid of the dependency from queue-runner -> *-store.

We will also have to bump the required nix version in release.nix 

With these changes hydra builds against nix master but the tests fail because of nix/src/libstore/import-export.cc:123 `assert(!accessor)`

@edolstra please tell me if my work is interfering with yours :wink:


domenkozar:
This is pretty exciting, would love to be able to upgrade Hydra to test declarative projects feature.

edolstra:
Thanks, I've merged this. I'll fix the accessor issue later today.

PR #315, opened 2016-05-19
Pbzip2 local

aycanirican:


aycanirican:
@domenkozar looks ok? Btw we may want to use gnu parallel to parallelize bzip2 but gnu parallel may introduce new dependencies.

domenkozar:
What's the difference to pbzip2?

aycanirican:
I accidentally pushed a merge commit hence created a new one. Also I added multiple cpu support by default.

domenkozar:
@edolstra this one is nice since it should speed up private hydra downloads. 

Current situtation:

2016-05-26-192448_990x97_scrot

domenkozar:
I wonder if we should just switch to `pixz` and use `xz`. I'm using this patch now with `--fast` since it's a waste to have 1Gbit connection and then waste resources on CPU

domenkozar:
According to https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/, gzip is the fastest (CPU wise),  so I'm using it now. My use case is a very fast line (1gbit+) and really just don't want to spend CPU cycles as a bottleneck.

lukego:
@aycanirican  Consider using `--loadavg` argument to `pbzip2`? That may reduce the risk of having too many pbzip2 processes (e.g. when compressing multiple large files) that would create contention for the scheduler and slow down other processes on the machine.

domenkozar:
@aycanirican I've ported your efforts to use `xz` compression over `bzip2`, see https://github.com/NixOS/hydra/pull/315 for details. Thanks for this efforts!

aycanirican:
Sorry, I haven't push forward to improve this PR due to lack of time. Thank you @domenkozar. I'm closing this.

PR #320, opened 2016-05-26
hydra-module: fix useSubstitutes configuration option

abbradar:
Without this I get:
```
error: value is a function while a Boolean was expected
```


domenkozar:
Thanks!

PR #326, opened 2016-06-12
WIP Add GithubAPIInput

expipiplus1:
This commit defines a new input type for performing queries against the
GitHub API.

The JSON results of the query is passed to nix expressions as a string.

An example query to get information on pull requests could be
`/repos/username/repositoryname/pulls`.

### DIFF
--- a/src/lib/Hydra/Helper/AddBuilds.pm
+++ b/src/lib/Hydra/Helper/AddBuilds.pm
@@ -294,8 +294,26 @@ sub inputsToArgs {
             if scalar @{$inputInfo->{$input}} == 1
                && defined $inputInfo->{$input}->[0]->{storePath};
         foreach my $alt (@{$inputInfo->{$input}}) {
-            if ($alt->{type} eq "string" || $alt->{type} eq "githubAPI") {
+            if ($alt->{type} eq "string") {
                 push @res, "--argstr", $input, $alt->{value};
+            }
+            elsif ($alt->{type} eq "githubAPI") {
+                my $template = "githubAPIXXXXX";
+                my ($fh, $filename) = File::Temp::tempfile( $template, TMPDIR => 1, SUFFIX => '.json');
+                print $fh $alt->{value};
+                close $fh;
+
+                # Add file to store
+                my $adder = "nix-store";
+                my @cmd = ($adder, "--add", $filename);
+                (my $res, my $storePath, my $stderr) = captureStdoutStderr(21600, @cmd);
+                die "$adder returned " . ($res & 127 ? "signal $res" : "exit code " . ($res >> 8))
+                    . " while adding github input to store:\n" . ($stderr ? decode("utf-8", $stderr) : "(no output)\n")
+                    if $res;
+                unlink $filename or print STDERR "WARNING: unable to remove temporary file $filename\n";
+
+                $storePath =~ s/^\s+|\s+$//g;
+                push @res, "--arg", $input, "builtins.readFile $storePath";
             }
             elsif ($alt->{type} eq "boolean") {
                 push @res, "--arg", $input, booleanToString($exprType, $alt->{value});
@@ -334,7 +352,8 @@ sub evalJobs {
 
     my $evaluator = ($exprType eq "guile") ? "hydra-eval-guile-jobs" : "hydra-eval-jobs";
 
-    my @cmd = ($evaluator, $nixExprFullPath, "--gc-roots-dir", getGCRootsDir, "-j", 1, inputsToArgs($inputInfo, $exprType));
+    my @args = inputsToArgs($inputInfo, $exprType);
+    my @cmd = ($evaluator, $nixExprFullPath, "--gc-roots-dir", getGCRootsDir, "-j", 1, @args);
 
     if (defined $ENV{'HYDRA_DEBUG'}) {
         sub escape {
--- a/src/lib/Hydra/Helper/AddBuilds.pm
+++ b/src/lib/Hydra/Helper/AddBuilds.pm
@@ -313,7 +313,7 @@ sub inputsToArgs {
                 unlink $filename or print STDERR "WARNING: unable to remove temporary file $filename\n";
 
                 $storePath =~ s/^\s+|\s+$//g;
-                push @res, "--arg", $input, "builtins.readFile $storePath";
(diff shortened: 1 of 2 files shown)


expipiplus1:
I've never used perl before, so I'm sure there's lots of unidiomatic code in here.

This is very rough at the moment, I'd be very happy to tidy it up if people are interested in merging it eventually.

expipiplus1:
Combined with #277 this allows a hacky way of building pull requests.

The `spec.json` here queries github for information on any pull requests and creates a jobset for each active pull request.
https://github.com/expipiplus1/hydra-deployment

See https://github.com/expipiplus1/teeth/pull/4

This method is not perfect, it relies on hydra polling github for new pull requests, and it has a bit of trouble when builds have already been cached that I've yet to investigate.

aycanirican:
+1 for this PR since we're already using it

expipiplus1:
@aycanirican I'm interested in hearing how you're using this!

shlevy:
Definitely interested in this, sorry I just saw it! Can you update this to pass a set corresponding to the JSON rather than the raw JSON string?

expipiplus1:
Is there a function in perl to do that conversion. Another issue is that the json string might represent a set or a list. Although nix isn't strongly typed it seems a shame to pass it a value of unknown type. Then again that will have to be handled somewhere.

shlevy:
Eh, I guess just pass `builtins.fromJSON <json-string>` should be fine.

shlevy:
(so not `--argstr`, but `--arg`)

shlevy:
@expipiplus1 still working on this?

expipiplus1:
@shlevy I've not done anything more on it recently.

The correct solution would be to write the JSON out to a file, load that as the input to the build and clean up the file after, see https://github.com/NixOS/hydra/issues/387

aycanirican:
@expipiplus1 could you please rebase since most of the functionality is moved to `src/hydra-eval-jobset`.

Also please consider adding "?per_page=<some_high_integer>" to the end of the `$url` since github api is limited to give 30 items by default.

expipiplus1:
@aycanirican:

 I'm afraid that most of my CI-based free time is going into getting Hercules off the ground. 

The line to change to get that behaviour would be line 24 of GithubAPIInput.pm.

Perhaps this PR should be closed since it no longer applies and I'm unlikely to port the code to hydra-eval-jobset.

Sorry for the bad news.

@shlevy:

 I have updated this branch with the code to put the github input into temp files to be read with `builtins.readFile`. No GC roots are added for these files, so there does exist a race condition! It seems to work well for us though as the files are used immediately and the GC isn't run very often :|

src/lib/Hydra/Helper/AddBuilds.pm
```diff
@@ -297,6 +297,24 @@ sub inputsToArgs {
...
+                my $adder = "nix-store";
+                my @cmd = ($adder, "--add", $filename);
+                (my $res, my $storePath, my $stderr) = captureStdoutStderr(21600, @cmd);
+                die "$adder returned " . ($res & 127 ? "signal $res" : "exit code " . ($res >> 8))
+                    . " while adding github input to store:\n" . ($stderr ? decode("utf-8", $stderr) : "(no output)\n")
+                    if $res;
+                unlink $filename or print STDERR "WARNING: unable to remove temporary file $filename\n";
+
+                $storePath =~ s/^\s+|\s+$//g;
+                push @res, "--arg", $input, "builtins.readFile $storePath";
```
shlevy: Can we make this `builtins.fromJSON (builtins.readFile $storePath)`? The nixexpr shouldn't care that the data it's passed was serialized as JSON at some point.

src/lib/Hydra/Helper/AddBuilds.pm
```diff
@@ -334,7 +352,8 @@ sub evalJobs {
 
     my $evaluator = ($exprType eq "guile") ? "hydra-eval-guile-jobs" : "hydra-eval-jobs";
 
-    my @cmd = ($evaluator, $nixExprFullPath, "--gc-roots-dir", getGCRootsDir, "-j", 1, inputsToArgs($inputInfo, $exprType));
+    my @args = inputsToArgs($inputInfo, $exprType);
+    my @cmd = ($evaluator, $nixExprFullPath, "--gc-roots-dir", getGCRootsDir, "-j", 1, @args);
```
shlevy: Why this change?

src/lib/Hydra/Plugin/GithubAPIInput.pm
```diff
@@ -0,0 +1,36 @@
...
+    $inputTypes->{'githubAPI'} = 'GitHub API Query';
+}
+
+sub fetchInput {
+    my ($self, $type, $name, $value) = @_;
+
+    return undef if $type ne "githubAPI";
+
+    # Use the GithubStatus plugin's authorization.
+    # TODO: move this into a more cleanly shared field.
```
shlevy: Can you do this before we merge?

shlevy:
Ah, I see you're shifting to hercules. OK! Should we close this?

src/lib/Hydra/Helper/AddBuilds.pm
expipiplus1: Ah, I used to pass a list of temporary file allocations so had to split this code, you're right, it is a redundant change after I stopped doing that.

src/lib/Hydra/Plugin/GithubAPIInput.pm
expipiplus1: vacuously yes!

src/lib/Hydra/Helper/AddBuilds.pm
expipiplus1: Done

expipiplus1:
@shlevy like @aycanirican pointed out this doesn't apply cleanly and might require some non-trivial work to get going again :(

I'm not planning on doing anything more on this branch other than bugfixes for our usage until we switch to Hercules, so unless someone else wants to continue that (I'd be more than happy to advise/help) we should close this.

Thanks for the review anyway :)

shlevy:
OK! Where's the hercules repo, I'd like to trackwhat's going on there :smile:

aycanirican:
@shlevy https://github.com/hercules-ci/hercules

PR #327, opened 2016-06-12
Insert a github token into the git URI when able

### LINKED ISSUE #324: Allow fetching private github repositories with an authorization token
It would be nice to have a single place in the configuration where a github token could be declared rather than having to specify fetch urls with `username:password` in them, and tokens elsewhere.

expipiplus1:
If the user has specified a github token in the config and the git clone
address is under github.com and the scheme is https then the token will
be used for the fetch. This allows fetching private repositories without
including a username and password in the URI.

There is some bikeshedding to be done over when to insert the token. At
the moment it's done as early as possbile which exposes the token in
logs and elsewhere.

The token is read from the 'authorization' field in the '<githubstatus>'
block. It must be of the format: 'token <40 character hex string>'.

Closes #324

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -102,6 +102,7 @@ rec {
             TestMore
             TextDiff
             TextTable
+            URI
             XMLSimple
             nix git boehmgc
           ];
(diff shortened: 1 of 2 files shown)


expipiplus1:
23d8942 modifies nix-prefetch-git to insert a token into the URI when checking out submodules.

This involves some code duplicated between `GitInput.pm` and `nix-prefetch-git`. It would be really nice to get rid of `nix-prefetch-git` and do everything in (slightly more sane) Perl.

PR #330, opened 2016-06-13
Don't set url rev and expHash unconditionally

### LINKED ISSUE #329: nix-prefetch-git handles it's arguments poorly
`nix-prefetch-git --leave-dotgit` will actually pass `--leave-dotgit` to `git remote add` as `url` is unconditionally set to `$1` in the third line of the script.

expipiplus1:
These are set properly by the argument parser.

Fixes #329

### DIFF
--- a/src/script/nix-prefetch-git
+++ b/src/script/nix-prefetch-git
@@ -1,8 +1,8 @@
 #! /bin/sh -e
 
-url=$1
-rev=$2
-expHash=$3
+url=
+rev=
+expHash=
 hashType=$NIX_HASH_ALGO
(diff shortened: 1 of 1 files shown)


domenkozar:
The correct fix would be https://github.com/NixOS/hydra/issues/346 - sorry for trouble

expipiplus1:
Is it possible to merge this in the meantime?

expipiplus1:
Seeing as #346 has not been closed yet, can this be merged? Although it's not the correct fix it's better than the current behaviour and will be removed when #346 is resolved.

PR #337, opened 2016-06-19
hydra-queue-runner: fix missing includes after gcc-5.4

vcunat:


domenkozar:
@vcunat is this backwards compatible?

vcunat:
Yes, I'm convinced of it, though I haven't tested that. I just fixed some parts that were non-conforming and only accepted due to gcc implementation (until 5.4).

aszlig:
Oh, somehow missed this and fixed it myself in 53ee2d0630a04a115561fe1b218e5b9adb857b31. I'm still not sure how `powf()` and `roundf()` could even end up there globally.

aszlig:
@vcunat: Did you by any chance find the changelog/changeset of GCC 5.4 that's relevant to this. I'm just curious to know, so if you haven't the world won't break apart :-)

vcunat:
Closed, it seems exactly equivalent change.

Standard headers can inside themselves `#include` other standard headers, but the users shouldn't rely on that. (AFAIK in C/++ there's no good way to use one header from another one without exposing it.) There's [no news at all for 5.3 -> 5.4](https://gcc.gnu.org/gcc-5/changes.html), and I don't know about particular commits. Several months before that we had another math-related strictness change that required patching some packages.

PR #338, opened 2016-06-19
Enable apache style includes in the config file

expipiplus1:
This enables the include syntax `include filename`. At the moment only `<<include filename>>` is allowed.

This also seems to enable absolute paths as filenames, but I can't find any documentation saying that.


expipiplus1:
Is there anything else I need to do to this before it's ready to merge?

PR #340, opened 2016-06-21
Trigger ssh like urls from github pushes

expipiplus1:
URLs like "git@github.com:owner/repo.git

This commit removes the need for a forward slash between 'github.com' and
'owner'.


expipiplus1:
Thanks!

PR #347, opened 2016-06-27
Build fixes

bennofs:
This PR contains 3 commits that together make the following two commands "just work":

    $ nix-build
    $ nix-shell

It hardcodes the nix revision that hydra needs to be built against.


domenkozar:
cc @edolstra

domenkozar:
Right, now that Hydra module is in NixOS, you can't include the one in this repo..

default.nix
```diff
@@ -0,0 +1,150 @@
+{ pkgs ? (import <nixpkgs> {})
+, version ? builtins.readFile ./version + ".HEAD"
+, src ? ./.
+}:
+
+let
+  # Build a version of nix that works with current hydra master
+  # (Currently, revision 1a714952732c56b4735f65dea49d406aacc7c595 works)
+  nixSrc = pkgs.fetchgit {
```
copumpkin: Why not `fetchFromGitHub`? Far more efficient!

bennofs: no particular reason. I'll fix it

aszlig:
Hm, I'd keep the `default.nix` as minimal as possible and only pass a fixed Nix version to the `release.nix` (like I've done in 1d6e087969f9f69b4a3685c5b0a3a953fb25831a, but not to the extent of actually providing the fixed source), because the way you're doing it it's not easily possible to build with a different Nix version without changing `default.nix`.

If you want I can cherry-pick your other changes to #353 and add a `default.nix` with a fixed Nix version.

bennofs:
@aszlig sounds good

default.nix
```diff
@@ -0,0 +1,150 @@
...
+  # Hydra currently requires a fork of aws-sdk-cpp with a few custom patches.
+  aws-sdk-cpp = pkgs.lib.overrideDerivation (pkgs.aws-sdk-cpp.override {
+    apis = ["s3"];
+    customMemoryManagement = false;
+  }) (attrs: {
+    src = pkgs.fetchFromGitHub {
+      owner = "edolstra";
+      repo = "aws-sdk-cpp";
+      rev = "local";
+      sha256 = "1vhgsxkhpai9a7dk38q4r239l6dsz2jvl8hii24c194lsga3g84h";
```
aszlig: Hm, that branch is rather stale (apart from that I'd use the commit rev instead of the branch name, because the latter could change), why exactly does it require a fork?

aszlig: My bad, that was from the original `release.nix`. Sorry for the noise.

bennofs:
Replaced by #353, closing

PR #353, opened 2016-07-04
Make it easier to spin up a development environment

aszlig:
So far the instructions in the `README` for spinning up a development environment were quite terse and also not to its full extent, because in the end the production environment is running PostgreSQL and not SQLite.

With these changes, we now have the ability to quickly spin up a local PostgreSQL cluster within the `inst/database` directory which is started and shut down upon entering/exiting a `nix-shell`.

Using `setup-dev-env` we now have a function inside the Nix shell, which should do all that work on the users behalf.

The main motivation behind this change was that I usually need to be mentally prepared (like: "gosh, I need to set up a dev environment to properly work on Hydra... maybe tomorrow...") for it to go through all these steps and wanted to make it easier for new developers to dive in.

Cc: @edolstra, @domenkozar

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,11 @@ This will run the following commands:
 
 The `setup-database` command is used for setting up a temporary database living
 in the `inst/database` subdirectory of the project root. It also sets up the
-required environment variables `HYDRA_HOME`, `HYDRA_DATA` and `HYDRA_DBI`.
+required environment variables `HYDRA_HOME`, `HYDRA_DATA` and `HYDRA_DBI` as
+well as the `PG*` environment variables so that if you want to start a database
+shell it's a matter of simply running:
+
+    $ pgsql
 
 Or, if you just want to build from source (on `x86_64-linux`):
 
--- a/README.md
+++ b/README.md
@@ -1,11 +1,19 @@
-To start hacking on Hydra, run:
+To quickly start hacking on Hydra, run:
+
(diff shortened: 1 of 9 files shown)


bjornfor:
Seems there is some churn in the shell-hook that can be squashed, but the changes look very useful. Nice!

aszlig:
Please don't merge yet, found a few bugs.

@bjornfor: Which churn are you referring to exactly?

bjornfor:
@aszlig: The setup-hook is first added inside a nix expression, then it is moved to a separate file, then it gets indentation fixes, then some error fixes... I just thought it'd be better to just have one commit "here's the setup-hook, where it should be and with proper error checking" instead of having the intermediate steps. But, it's nitpicking!

aszlig:
Bug fixed, tested also against a clean working dir.

aszlig:
Hm, need to implement some kind of locking mechanism to make sure that if you start several shells the first one doesn't tear down the database on exit.

aszlig:
Locking (in this case using semaphores) implemented and tested :-)

gilligan:
It's a bit unfortunate that so much work went into this but it still ultimately didn't lead to anything concrete - now that #759 has been merged i think I can go ahead and close this one.

aszlig:
@gilligan: Well it did work at some point back then, but it was just rotting here waiting to be reviewed/merged, but since there's now a merged solution that's even better :-)

PR #355, opened 2016-07-07
Remove SQLite and support PostgreSQL only

aszlig:
We already have a few stored procedures and other incompatibilities between the two database backends and after asking @edolstra it's clear that we don't want to support SQLite anymore.

Unfortunately there are a few problems for an easy switch, mainly that if we don't use SQLite anymore for generating the DBIx schema, we will end up with a schema with all-lowercase table names, because PostgreSQL [folds these names to lower-case](https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS).

Another issue with this identifier folding is already something which shows up in the code base: While in the SQL files every identifier is `CamelCased`, the actual database _and_ the code referring to the identifiers is not.

So to avoid confusion and also make naming of these identifiers more consistent I changed every `camelCased` identifier to use `under_scores` instead.

This is somewhat related to #353, because that PR makes it easier to spin up a temporary PostgreSQL cluster.

Cc: @edolstra


edolstra:
I'd really prefer not to change the entire schema (giving a huge amount of code churn) just because the moniker mapper doesn't work properly. After all, the mapper is customizable right?

aszlig:
@edolstra: The mapper is customizable, but we'd still have a major inconsistency when it comes to identifier naming, because the code would refer to camelCased names while the database *and* schema would map them to lower-case names. So every time you work with database identifiers you'd have to look up the right name of the mapper *and* the SQL schema in case of raw SQL.

Also we have a lot of code which bypasses the mapper (not only C++ code but also some Perl code).

aszlig:
Rebased against current master.

domenkozar:
Given that it's just a rename (so it's reversible in any case), I'd say merge this as it removes tech debt with sqlite :)

grahamc:
I spent a few hours today trying to get a dev hydra install working, and came to the conclusion that sqlite support is actually broken. Getting this support out of here would be much appreciated by the next person who sets up a dev env :)

PR #358, opened 2016-07-11
Add plugin for updating symlink on job completion

expipiplus1:
I've written a small plugin which can create a symlink pointing to the latest output of a job.

It also touches a file as lots of filesystem watchers don't like watching symlinks :/

I'd appreciate some feedback :)

This can be used with a section in the config like this:

```
      <latestoutput>
        jobs = hoogle-instance.default.hoogle
        path = /nix/var/nix/gcroots/hydra/hoogle
        touch = /tmp/hoogle-updated
      </latestoutput>
```

I'm using this at the moment to run a hoogle server with the latest version of all our code on.

### DIFF
--- a/src/lib/Hydra/Plugin/LatestOutputLink.pm
+++ b/src/lib/Hydra/Plugin/LatestOutputLink.pm
@@ -1,0 +1,61 @@
+package Hydra::Plugin::LatestOutputLink;
+
+use strict;
+use parent 'Hydra::Plugin';
+use HTTP::Request;
+use JSON;
+use LWP::UserAgent;
+use Hydra::Helper::CatalystUtils;
+use Data::Dump;
+
+sub buildFinished {
+  my ($self, $build, $dependents) = @_;
+
+  # TODO: Lookup what these statuses are, these are just cribbed from
+  # S3Backup
+  return unless $build->buildstatus == 0 or $build->buildstatus == 6;
+
+  my $jobName = showJobName $build;
+  my $job = $build->job;
+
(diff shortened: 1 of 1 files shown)


src/lib/Hydra/Plugin/LatestOutputLink.pm
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    my $output = $buildoutputs[0];
+    my $outputPath = $output->path;
+    my $linkLocation = $config->{path};
+    my $linkTarget = $outputPath;
+
+    print STDERR "INFO: Creating symlink at $linkLocation with target $linkTarget\n";
+    if (-e $linkLocation) { unlink ($linkLocation); }
+    my $linkCreated = symlink($linkTarget, $linkLocation);
```
aszlig: What about using `nix-env -p $linkLocation --set $linkTarget` here instead? This would also allow for rollbacks and would ensure that we have a gcroot.

expipiplus1:
@aszlig I've made that change.

expipiplus1:
The fact that only two lines in this change are actually specific to what this plugin is doing makes me wonder if there's a way to abstract over all the rest.

Some kind of 'on build completion script' as described in #281 might make sense.

grahamc:
This looks pretty cool. I wonder, though, if the runcommand plugin is satisfactory?

expipiplus1:
@grahamc Perhaps, I've not looked at this for ages.

gilligan:
Since there was no further response and the runCommand indeed seems like a good candidate i’ll close this one.

PR #370, opened 2016-07-29
Add a nix expression to build the manual

ericsagnes:
This add a nix expression to build the manual with `nix-build` in a similar fashion to `nixpkgs`.

```
$ nix-build doc/manual -A manual
$ nix-build doc/manual -A manualPDF
```

### DIFF
--- a/doc/manual/default.nix
+++ b/doc/manual/default.nix
@@ -1,0 +1,53 @@
+let
+  pkgs = import <nixpkgs> { };
+  lib = pkgs.lib;
+  sources = lib.sourceFilesBySuffices ./. [".xml"];
+in
+pkgs.stdenv.mkDerivation {
+  name = "hydra-manual";
+
+  buildInputs = with pkgs; [ pandoc libxml2 libxslt zip ];
+
+  xsltFlags = ''
+    --param section.autolabel 1
+    --param section.label.includes.component.label 1
+    --param html.stylesheet 'style.css'
+    --param xref.with.number.and.title 1
+    --param toc.section.depth 3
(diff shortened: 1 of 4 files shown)


domenkozar:
Note that this duplicates `doc/manual/Makefile.am` which should then be replaced. Let's wait what @edolstra thinks about this.

I'm +1 having a separate build

ericsagnes:
update: added a PDF manual like NixOS

edolstra:
I don't really see the need for a separate manual job. The manual is already available as part of the `tarball` job: hydra.nixos.org/job/nix/master/tarball/latest/download-by-type/doc/manual

ericsagnes:
@edolstra Sorry, I should have been more clear with the motivation of this PR.

I am starting to use Hydra and was thinking to contribute to the manual.
Having a nix expression to build the manual is an easy way to check the syntax and the rendering.

This also introduce the validation of the docbook (there was a few validation error in the current version, see the first commit of the PR for details) and the generation of a PDF version.

If you think this is redundant, I see no problem in closing the PR.
But I personally think that any thing that can facilitate contributions to the Hydra documentation should be considered as current state is sub optimal.

domenkozar:
@edolstra would it be sufficient if we build manual separately and symlink it for the tarball job? That would allow for easier contributions and it wouldn't duplicate the code for building it.

domenkozar:
Note: I'll start contributing some docs soon and this would make it simpler from my perspective.

grahamc:
Closing: hydra's manual is markdown / mdbook now.

Issue #279, opened 2016-03-12
Get rid of jobset alts


shlevy:
They seem to complicate the code to basically no purpose nowadays (with `genAttrs` and nix jobset inputs especially).

domenkozar:
What are alts?

shlevy:
Exactly :grin: I think the idea is to reevaluate the same release.nix with multiple versions of an input, originally needed to support multiple systems? But now we can pass arbitrary nix expressions as inputs, so you can just pass them as a list.

domenkozar:
Where do you even specify alts? :D

shlevy:
Hmm, I'm not sure you even can anymore.

domenkozar:
Cool, let's do it :)

edolstra:
This is basically a dupe of #60.

domenkozar:
See #375

PR #375, opened 2016-08-04
Replace jobset_input_alts with a JSONB field

aszlig:
This builds upon #355 (thus the commits within this PR will be less cluttered after #355 is merged, otherwise please start at [d00399badac55a9a5edeac6ae21cfedfbb54c325 with reviewing](https://github.com/aszlig/hydra/compare/remove-sqlite...remove-input-alts)) and uses a single [`JSONB`](https://www.postgresql.org/docs/9.4/static/datatype-json.html) field to allow for multiple values for a specific jobset input.

So far we already had multiple values, but those were stringly typed by either splitting them of or using [regular expressions like this](https://github.com/NixOS/hydra/blob/dd32033657fc7d6a755c2feae1714148ee43fc7e/src/lib/Hydra/Helper/AddBuilds.pm#L29-L45) to tear apart the string into multiple values.

The result is something like this in the edit view of the jobset:

image

And this is how it looks like in the `Configuration` tab of the jobset:

image

Fixes: #279
Cc: @edolstra


shlevy:
Why do we need multiple values at all anymore?

aszlig:
@shlevy: In this particular case it's for setting properties on Git inputs to handle pull requests. And apart from that, we already have multiple values, for example for branches/deepClone, ids, job(set) matching and so on. So do you have a better solution to address this (except for EAV of course)?

aszlig:
Rebased against current version of #355.

shlevy:
Oh, I thought you were just reimplementing nput alts. OK, that's fine.

aszlig:
@shlevy: Missed that project edit/create has an input as well, so I've fixed it and generalized the code used for jobset inputs to be re-used for declarative jobsets.

PR #377, opened 2016-08-12
Use pixz instead of bzip2 for compressing nars

domenkozar:
According to following two benchmarks:

- https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/
- http://catchchallenger.first-world.info/wiki/Quick_Benchmark:_Gzip_vs_Bzip2_vs_LZMA_vs_XZ_vs_LZ4_vs_LZO

xz has better compression ratio than bzip2 at lowest compression rate.

https://github.com/vasi/pixz has been chosen as it can scale compressing
over multiple cores linearly.

We're using this in snabblab for a month now and it has improved CPU
wise the main Hydra server.

cc @edolstra

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -225,6 +225,7 @@ in
         base_uri ${cfg.hydraURL}
         notification_sender ${cfg.notificationSender}
         max_servers 25
+        compress_num_threads 0
         ${optionalString (cfg.logo != null) ''
           hydra_logo ${cfg.logo}
         ''}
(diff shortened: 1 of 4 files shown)


ctheune:
I just stumbled over this as we upgraded our Hydra, but our nix is a bit too old for this. I'm running our hydra without your change for now. The newer Nix has the support for both bzip and xz. Ideally I would love if hydra would allow supporting both, too. However, looking at the protocol, this likely means that we sniff the user agent and decide on that basis ... would that be something you could support?

ctheune:
Please forget about my previous comment for now. I think we may have barked up the wrong tree here ...

PR #378, opened 2016-08-12
Give highest CPU priority to evaluator right after kernel

domenkozar:
We've noticed that when Hydra serves a lot of nars, they hog CPU usage a lot for compression.

This PR gives evalutor the highest priority so builds can get in the pipeline ASAP.

See https://github.com/snabblab/snabblab-nixos/commit/07c3749057b9d28af11149f999aada115e2146d6 for downstream commit.

cc @edolstra

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -345,6 +345,8 @@ in
           { ExecStart = "@${cfg.package}/bin/hydra-evaluator hydra-evaluator";
             User = "hydra";
             Restart = "always";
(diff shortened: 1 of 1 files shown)


edolstra:
IMHO this is is a policy decision that belongs in the user's system configuration, not in the generic Hydra module. For example, I would probably use cgroup CPUShares rather than niceness.

domenkozar:
Makes sense, this works for us now so we'll keep it this way.

PR #382, opened 2016-08-17
Send BuildFinished notifications on cached build results.

### LINKED ISSUE #342: Post build hooks are not triggered when a build has been cached.
## Steps to reproduce using the Slack plugin as an example
- Create a repo and hydra instance set up to post to a Slack channel
- Push an initial change to the repo
- Observe that the build has changed to success and hydra posts to Slack
- Push a change to break the build
- Observe that the build has changed to failure and hydra posts to Slack
- Use `git revert HEAD` to undo the last change and push
- _Observe that Slack does not post about the build becoming successful to Slack_
## What I think is happening

The hash of the build inputs is the same since we're not using KEEP_DOT_GIT, hydra uses the prebuilt derivation so the `buildFinished` plugin hook never fires, so nothing gets posted to slack.

CC @shlevy

shlevy:
Fixes #342

### DIFF
--- a/src/hydra-queue-runner/queue-monitor.cc
+++ b/src/hydra-queue-runner/queue-monitor.cc
@@ -192,6 +192,8 @@ bool State::getQueuedBuilds(Connection & conn, ref<Store> localStore,
                 nrBuildsDone++;
             }
 
+            enqueueNotificationItem({NotificationItem::Type::BuildFinished, build->id});
+
             return;
         }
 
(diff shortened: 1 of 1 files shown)


shlevy:
Not yet tested (not even compile tested), @expipiplus1  can you try this?

domenkozar:
Wouldn't it be better to have a separate hook? One might want to differentiate the two.

expipiplus1:
I'm going on vacation until the 30th so won't be able to test this before then. Please ping me in #nix-hydra then if you'd still like me to try this out.

shlevy:
@domenkozar You can check `build->isCachedBuild`

domenkozar:
@edolstra objections?

shlevy:
@expipiplus1 Did you have a chance to test this?

expipiplus1:
@shlevy yeah, it's been working well for us, thanks!

shlevy:
@edolstra Can we merge this in?

PR #407, opened 2016-10-24
Update build images to be more friendly to users with color blindness.

### LINKED ISSUE #406: Improve accessibility of Hydra
With the current iconset of Hydra, I have a very difficult time deciphering the build statuses. It is difficult for me to discern:
-  ? and the check-mark icons (similar color, small interior symbol inside the circle)
- The different "X" and "!" icons
- sometimes, even the check (pass) and X (failed) -- see the second two types of color blindness.

Here is an example of how users with different types of color-blindness perceive Hydra build icons: 

color-blind

Note that when viewing a build matrix, the icons are _much_ smaller.

I'd propose changing the icons to:
1. Not have a colored circle around the symbol
2. Use a larger symbol

Perhaps also using different symbols for different build failures.

Here is such an example of proposal items 1 and 2:
proposed

Note that in this proposed version, there is also screen-reader specific indicators of build status, and on-hover tool-tips.

Before I put forth the effort to get it in to hydra proper, I'd like to know that the change would be accepted in the first place. May I have some feedback regarding that? If this change is unwanted, I would rather spend the effort implementing this change as a greasemonkey script.

Thank you!
Graham

grahamc:
Closes #406 

This is how Hydra works for users with color blindness:

color-blind
---

Now:

pending
project-view

The symbols are from EmojiOne.com which are CC-BY 4.0 and require no attribution for non-commercial / open source use.

I wasn't able to trigger actual builds though due to numerous problems, this mean I wasn't able to exercise the full change -- ie: the failure modes. I'd lovehelp either getting my machine to test that, or someone else testing that ... or just merging anyway.

It is difficult for me to use Hydra.


grahamc:
Here are all of them:

all

edolstra:
What's the performance impact of using SVG? Are browsers smart enough to render each icon only once?

grahamc:
@edolstra I did a test and put each icon on a page 600 times (a total of 3,000 icons, to simulate a big build matrix.) Firefox and Chromium seem to render and process each icon once, and then reuse for the remainder 599 display spots.

shlevy:
This seems like an obvious win for me, if @edolstra hasn't gotten back in a few days ping me and I'll merge it.

edolstra:
CC BY 4.0 requires attribution right? ("You must give appropriate credit, provide a link to the license, and indicate if changes were made.") Maybe add an example link like those listed in http://emojione.com/licensing/#attribution to README.md.

groxxda:
@edolstra [source](http://emojione.com/licensing/#attribution):
> Personal or non-commercial use of our emoji do not require attribution.

Using unicode is of course also an option, but then the page will not look uniform across OS/browsers.

grahamc:
Eelco Dolstra <notifications@github.com> writes:

> CC BY 4.0 requires attribution right? ("You must give appropriate credit, provide a link to the license, and indicate if changes were made.") Maybe add an example link like those listed in http://emojione.com/licensing/#attribution to README.md.


I thought we didn't because of:

    Personal or non-commercial use of our emoji do not require
    attribution.

but I've updated the PR to include the attribution.

edolstra:
@grahamc Thanks!

PR #416, opened 2016-10-29
src/sql/upgrade-50.sql: fix sql syntax

k0001:
Her I fix some minor SQL syntax issues that were causing migrations to fail.


domenkozar:
@edolstra

edolstra:
Hm, what's the error? This seems to work fine for me (after 41a020e3237fe487e3b5bdacf7506f2da2a2f862).

k0001:
I don't have the error at hand now, because I have already ran the SQL with
this fix, but basically it was complaining about SQL syntax being wrong.
This was with PostgreSQL 9.4. @edolstra that commit you mentioned doesn't
solve the issue (I was already running
2b790ceefa4d3b92acbee745b1ce8deaba1291a3, which is newer than the commit
you mentioned, I when I encountered this issue).

edolstra:
I'm running 9.5, maybe that explains it. Note that if you change `upgrade-*.sql`, you should change `hydra.sql` as well.

basvandijk:
Note I also got the syntax error on `upgrade-50.sql` with postgresql-9.4. After switching to postgresql-9.5 the error disappeared.

k0001:
I just added the changes to `hydra.sql` as well.

PR #424, opened 2016-11-25
Fetch jquery/ui locally

### LINKED ISSUE #402: Add configure option to load all .js files locally
Some files are retrieved from google, such as: https://ajax.googleapis.com/ajax/libs/jquery/1.8.3/jquery.min.js

It'd be nice to have a configuration option to load all files locally. This has IMHO four advantages:
- No "leaking" of i.e. company internal URLs to external js cdns
- No instant compromise of the site in case the CDN is compromized
- No dependency on a working internet connection for the site to work
- "More nix style": An externally included js file can change independently of the nix expression that builds hydra.

montag451:
Fixed #402


domenkozar:
Thanks!

bjornfor:
FYI, jquery(-ui) is packaged in nixpkgs, it doesn't have to be committed to hydra source tree.

PR #432, opened 2016-12-11
Only set buildMachinesFiles when nix.buildMachines is defined

### LINKED ISSUE #430: For a self-contained instance, nix.buildMachines must be set but this isn't documented
```
2:02 PM <ocharles_> Hey - just updated to the Hydra in release-16.09, but now my Hydra server apparently has no machines (the /admin/machines route shows nothing) and no builds are happening. How do I set up that machine itself to be a build server?
2:03 PM <clever> ocharles_: it always needs a slave in /etc/nix/machines
2:03 PM <ocharles_> Ok, can that be populated with something in configuration.nix?
2:03 PM <ocharles_> I have no /etc/nix/machines file at the moment
2:04 PM <clever> ocharles_: its configured via nix.buildMachines in configuration.nix
2:04 PM <ocharles_> Ah, thanks!
2:06 PM <ocharles_> perfect, thanks clever! I'll open an issue to get this added to the hydra manual
```

I had to have add the following to the machines `configuration.nix`:

```nix
      nix.buildMachines = [
        { hostName = "localhost";
          maxJobs = "1";
          system = "x86_64-linux";
        }
      ];
```

There is no mention of this in https://nixos.org/hydra/manual/, but I think there should be.

basvandijk:
This should fix #430.

We should set `buildMachinesFiles` to `[ "/etc/nix/machines" ]`only when `nix.buildMachines` is set to a non-empty list because that [causes](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/misc/nix-daemon.nix#L341) that file to be generated. 

This should also be applied to [<nixpkgs/nixos/modules/services/continuous-integration/hydra/default.nix>](
https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/continuous-integration/hydra/default.nix#L169).

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -165,7 +165,7 @@ in
 
       buildMachinesFiles = mkOption {
         type = types.listOf types.path;
-        default = [ "/etc/nix/machines" ];
+        default = optional (config.nix.buildMachines != []) "/etc/nix/machines";
         example = [ "/etc/nix/machines" "/var/lib/hydra/provisioner/machines" ];
         description = "List of files containing build machines.";
       };


domenkozar:
@edolstra

joepie91:
Any reason this is not being reviewed/merged? Just ran into this issue with a self-contained Hydra installation, and it seems like a relatively simple change to apply.

grahamc:
One reason might be hydra.nixos.org's build machines file is managed manually, and this would change that behavior some.

basvandijk:
In the process of working on https://github.com/NixOS/hydra/pull/665 I was bitten by this again. I spend the best part of an hour figuring out why my jobs weren't being build until I remembered this PR.

If `hydra.nixos.org` is manually managing `/etc/nix/machine` wouldn't it make more sense to explicitly set `services.hydra.buildMachinesFiles = [ "/etc/nix/machines" ]` on `hydra.nixos.org`  and not rely on the default?

Issue #433, opened 2016-12-16
Failure to actually run the builds in the queue


FPtje:
# Problem description
Our private hydra build server (with @basvandijk) has had its database cleaned and currently holds one project with one jobset containing one (fresh) job. This job is put inside the Queue (as seen in `Status > Queue`), but it is never actually built.

When looking at the job's `Summary` page, the `Status` shows `Scheduled to be built`, part of evaluation 1. The build steps tab is _empty_. The build dependencies tab, though, shows a whole bunch of dependencies. 

The `Status > Latest steps` page is empty, as is the `Status > Latest builds`. The `Status > Running builds` says there are no running builds. The machines in `Machine status` are shown as Idle.

# Background
The issue has started since hydra was updated a month or so ago. In a debugging session (together with @basvandijk), the version was upgraded to revision `de55303197d997c4fc5`, where the issue still occurs. Jobs just wouldn't build. Only when actually running `nix-store --realize <drv path>` does hydra mark the job as finished.

Somewhat related to this issue is #431, in which @basvandijk describes some error with a Raspberry Pi 3 build server. He thought the Raspberry Pi stuff might cause Hydra to act this weird, but the one job currently active has nothing to do with the Raspberry Pi, which should exclude it as a cause.

# Debug information
With all `lvlChat` print messages changed to `lvlInfo` (because we're too lazy to figure out how to set verbose mode), the following is printed in the journal when the configuration is switched (and hydra is started):

```
hydra-queue-runner[9884]: adding new machine ‘root@rpibuild3’
hydra-queue-runner[9884]: checking the queue for builds > 0...
hydra-queue-runner[9884]: loading build 1 (lumi:staging:test-stalling-net)
nix-daemon[1629]: accepted connection from pid 9884, user hydra-queue-runner (trusted)
[...]

PR #434, opened 2016-12-17
Set nix.package to the same version that hydra is using

basvandijk:
In the process of debugging #433 @FPtje and I stumbled upon an interesting problem. Our hydra machine is deployed with a nixos-16.09 revision of nixpkgs. This means that its nix db has schema version 7. The machine uses hydra-HEAD however which [uses](https://github.com/NixOS/hydra/blob/master/release.nix#L57) a newer version of the nix tools. When the nix tools that hydra uses touch the DB they will upgrade its schema to version 10. However since the nix-daemon running on the machine only knows version 7 it will give errors.

The solution is for the hydra-module to set `nix.package` to the version it's using itself.

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -217,6 +217,8 @@ in
 
     nix.trustedUsers = [ "hydra-queue-runner" ];
 
+    nix.package = (import ./nix.nix pkgs).nix;
+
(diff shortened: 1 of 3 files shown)


domenkozar:
@edolstra what do you think about this? I was always curious how we can mix&match the two

dasJ:
Probably not needed anymore with Flakes

PR #451, opened 2017-03-20
hydra-eval-jobs: Disable the build hook.

shlevy:
For some reason this causes an unexpected EOF when doing import-from-derivation.


domenkozar:
That could also explain https://github.com/NixOS/nixos-org-configurations/commit/a122f418797713d519aadf02e677fce0dc1cb446

shlevy:
@edolstra Thoughts on this?

aycanirican:
This solved my "unexpected EOF" problem (we have switched to IFD recently). Thank you.

Issue #254, opened 2015-11-16
nixexprs.tar.bz2 contains broken channel/default.nix with else<EOF>


calbrecht:
At this very moment this happens for https://hydra.mayflower.de/jobset/nixos/hydra-jobs/channel/latest
you may verify this with:

```
channel=https://hydra.mayflower.de/jobset/nixos/hydra-jobs/channel/latest
curl $channel/nixexprs.tar.bz2 | tar -j -x channel/default.nix -f - -O | tail
```

though this has happened for http://hydra.nixos.org/jobset/gnu/emacs-trunk/channel/latest before as well.

One thing they both have in common (i guess) is that in configuration there is only one system defined.

https://github.com/mayflower/hydra-jobs/blob/master/default.nix

```
supportedSystems ? [ "x86_64-linux" ]
```

and http://hydra.nixos.org/jobset/gnu/emacs-trunk#tabs-configuration

```
system  String value    "x86_64-linux"
```

next guess is that in combination with failing jobs within a jobset this leads to a broken default.nix which ends without newline with else

calbrecht:
This is the perl script that creates the tar https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/View/NixExprs.pm

domenkozar:
Is this still happening?

calbrecht:
Yes, just tried the two cmd lines from description right now:
    
    channel=https://hydra.mayflower.de/jobset/nixos/hydra-jobs/channel/latest
    curl $channel/nixexprs.tar.bz2 | tar -j -x channel/default.nix -f - -O | tail

PR #455, opened 2017-03-31
NixExprs.pm: encode default.nix as utf-8 to fix missing chars/data.

dtzWill:
Fixes the problem described in #254 for me. \o/


dtzWill:
Thanks!

PR #466, opened 2017-04-14
Allow to configure the timeout value for the GitInput plugin in different places.

cyraxjoe:
1. From the hydra configuration file.

    The configuration is loaded from the `git-input` block.
    Currently only the `timeout` variable is been looked up in the file.
```
    <git-input>
      # general timeout
      timeout = 400

      <project:jobset:input-name>
        # specific timeout for a particular input
        timeout = 400
      </project:jobset:input-name>

    </git-input>
```

2. As an argument in the input value after the repo url and branch (and after the `deepClone` if is defined)
   `timeout=<value>`

The preference on which value is used:

 1. input value
 2. Inside a block `<project_name:jobset_name:input_name>` in the `<git-input>` block
 3. `timeout` inside the `<git-input>` block
 4. Default value of 600 seconds. (original hard-coded value)

The code is generalized for more values to be configured, it might be too much for a single value on a single plugin.

### DIFF
--- a/src/lib/Hydra/Plugin/GitInput.pm
+++ b/src/lib/Hydra/Plugin/GitInput.pm
@@ -49,7 +49,7 @@ sub _printIfDebug {
     print STDERR "GitInput: $msg" if $ENV{'HYDRA_DEBUG'};
 }
 
-=item _pluginConfig($main_config, $input_name)
+=item _pluginConfig($main_config, $project_name, $jobset_name, $input_name)
 
 Read the configuration from the main hydra config file.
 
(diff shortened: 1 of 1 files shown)


cyraxjoe:
Any interest in this PR? Potential improvements?

edolstra:
Can input names actually have spaces? That seems wrong to me.

Wouldn't a single config option for setting the global Git timeout suffice? I don't know if per-input configurability is all that useful. However, if we do want that, then probably the input name should include the project and jobset names as wel, to disambiguate between same-named inputs in different jobsets. (E.g. `<project:jobset:input1>`.)

cyraxjoe:
Thanks for the feedback @edolstra!

You're right, it shouldn't have spaces, I added that note because I was thinking on the Config-General perl module, but it doesn't make sense in this context. 

About the ability of specifying the value per input, I personally use that to make sure I only allow that much of time to a particular input that I know has a lot of static assets and it will fail for any reasonable timeout for the average repo, if you have a strong feeling against that much granularity I can remove that part.

cyraxjoe:
@edolstra included the feedback in the last commit, let me know if you have any additional comments/concerns.

Thanks.

cyraxjoe:
ping

cyraxjoe:
Any improvements that I could make to the PR to get this merged?

goodwillcoding:
This looks great. Any chance this will make it in @edolstra @domenkozar ?

edolstra:
Thanks, merged!

PR #471, opened 2017-04-20
Fix perl search path; avoid reference to build path

### LINKED ISSUE #460: PERL5PATH refers to build path, causes breakage while building hydra on hydra
The answer is probably "well don't build things on the hydra node, it has enough to worry about" but I hit this when using nixos-rebuild on my hydra instance.

The logs fill with "permission denied" for stict.pm and others, causing evaluations to fail with "error 13".

I'm not sure if these are added to make tests work, but it seems like they shouldn't be included in the final product.

spacefrogg:
Removes the dependency from the temporary build dir. Should fix #460


dtzWill:
:+1:

PR #472, opened 2017-04-20
Add sensible defaults to some numeric columns in SQL

spacefrogg:
The following numbers a calculated with in the web frontend, they
 should, thus, have default values.

 BuildSteps.timesBuilt is compared against 0 in `built.tt`.

 BuildProducts.fileSize is converted from bytes to megabytes in
 `product-list.tt`.


dtzWill:
Is this still relevant? If so please consider resolving the conflicts.  Thanks!

PR #474, opened 2017-04-21
Make all content of rotated table header clickable

zraexy:
Currently if you have multiple rotated table headers in succession, the top part of the `<div>` that holds the content (the link) gets covered up by the next header.


zraexy:
Would it be possible to get this merged?

grahamc:
Tested, LGTM

PR #491, opened 2017-06-27
hydra-module.nix: restart daemons on config change

domenkozar:
TODO: port to nixpkgs as well


edolstra:
BTW, I thought `hydra-queue-runner` had `restartIfChanged = false` to prevent restarting builds during reconfigurations, but that turns out to be hydra.nixos.org specific (https://github.com/NixOS/nixos-org-configurations/blob/master/delft/hydra.nix#L58).

PR #492, opened 2017-07-04
Add BitBucket plugin

Chakerbh:
This plugin will post to the build status system in BitBucket. In order
to use it you need to add to **ExtraConfig**
```
<bitbucket>
username = bitbucket_username
password = bitbucket_password
</bitbucket>
```
You can use an [application password]( https://blog.bitbucket.org/2016/06/06/app-passwords-bitbucket-cloud/) instead of your password. Or you could also use team name and API key.


edolstra:
If I understand this correctly, this will post a build status to every jobset input in Bitbucket. For jobsets that have lots of different Bitbucket inputs, this is probably not desirable - you only want to notify the "main" input, and not spam the dependencies. Maybe the `emailResponsible` field can be borrowed for this purpose? I.e. only do jobset inputs that have the `emailResponsible` field enabled.

PR #497, opened 2017-07-26
queue-monitor: never move lastBuildId forward without processing jobs.

dtzWill:
Touchup for fix made in dc5e0b120aac06f1c46bcf0296d14775d72c2354.

Notifications about jobs added shouldn't cause lastBuildId to advance, only to re-scan for possibly missed jobs.

lastBuildId should only advance due to jobs being processed, so take the minimum of the two instead.


dtzWill:
Switch to `std::stoul` and such is an attempt to be slightly nicer about the integer types but things will probably go bad if we get an unexpected value here anyway :).

edolstra:
Thanks, good catch!

PR #498, opened 2017-07-26
Hydra status datadog timeboard

AmineChikhaoui:


dd-dashboard.nix
```diff
@@ -0,0 +1,161 @@
+{
+  host
+, appKey
+, apiKey
+, ...
+}:
+{
+  resources.datadogTimeboards.dash = {
+    inherit appKey apiKey;
+    description = "Hydra build farm status (hydra.nixos.org)";
```
edolstra: I guess `hydra.nixos.org` should be `${host}`.

AmineChikhaoui: Depends on the hostname, in this case chef I think

PR #516, opened 2017-11-04
hydra-update-jobset cli utility

globin:
For the hydra PR testing I need a CLI utility to create and update them as there is no other API to that.
This might need some clean-up, I have no idea of perl and props to @LnL7 for the initial implementation.

- nixpkgs PR at: https://github.com/NixOS/nixpkgs/pull/31255
- nixborg code at: https://github.com/mayflower/nixborg (to be moved to the NixOS org if implementation is accepted)
- RFC upcoming


globin:
ping @edolstra

PR #520, opened 2017-12-05
hide disabled jobsets by default.

AmineChikhaoui:


dtzWill:
oh, YAY! Nice QoL improvement :+1:

Issue #436, opened 2016-12-28
Automatic rescheduling of jobs known to be prone to spurious failures


shlevy:
Many of our NixOS tests fail randomly quite regularly (without exaggeration, at least 9 times out of 10 that I check the `tested` job, at least one constituent has failed that succeeds after a restart). While ideally we would remove this impurity, until/unless that happens it would be great to be able to mark those jobs as prone to spurious failure and, rather than marking them as failed and failing dependents upon a single failure, rescheduling them (possibly waiting for a time when load is reduced) and only marking them as failed after repeated failure.

shlevy:
Closing this to clear up my open issues list, please open a new one if interested in pursuing this.

Issue #354, opened 2016-07-06
Option to treat build failure as transient 


expipiplus1:
It would be nice to be able to communicate with hydra from a build that it should be attempted again after some time has passed.

This probably has limited utility for perfectly pure builds. For builds which have time dependent impurities (a build tool requiring locking a stupid license) this would be super useful.

One way I can think of doing this would be for hydra to keep build failed builds around to examine a `hydra-retry-in` file which could contain a number of seconds after which to try again. There might be a better way of doing this though.

copumpkin:
Can you elaborate on what time-dependent impurities you have? Are you not running Hydra with sandboxed builds?

expipiplus1:
@copumpkin I'm compiling things with the Xilinx Vivado tools.

At the moment I'm doing the perfectly horrible `PATH=$PATH:/opt/Xilinx/Vivado/2016.2/bin` in `buildPhase` This obviously precludes sandboxed builds.

Vivado looks for a license file at `$XILINXD_LICENSE_FILE` or grabs a floating license and holds onto it until it's finished building. Hydra's (very desirable) propensity for multiple builds in flight at once has the potential to cause builds to fail because of a transient lack of a license.

I suppose that another solution to this would be to keep retrying a build for a set period of time.

Or another solution could be some kind of queue for builds wanting the license. This is probably not too bad given how impure this is already.

expipiplus1:
I think that given the lack of interest this is likely to get some kind of impure queue or retrying system might be best to avoid cluttering hydra's source.

domenkozar:
Could be you'd set a retry attribute on the derivation. There're some of those already existing, but they're not documented anywhere.

PR #523, opened 2017-12-13
Support builds that can abort themselves

rickynils:
If a build creates the file `/nix-support/aborted` and returns exit
status 0, Hydra will mark the build as aborted, and schedule it for
retry (just like it does for  builds that are manually aborted).
This way, if a build can identify a transient error, it can instruct
Hydra to try the build again.

I believe this could fix the issues #436 and #354.


edolstra:
The problem with this approach is that Nix *won't* in fact retry the build because its output path already exists in the Nix store. (Using multiple build machine you might get lucky and the retry might be scheduled on another machine where the output path doesn't exist.) Also, dependent builds added to / read from the queue can still proceed because the output path exists.

grahamc:
Closing since this should probably be a feature of Nix itself, and not an in-band "failing successfully" signal.

PR #526, opened 2017-12-30
Add a restart-jobs role

grahamc:
Frequently users want Hydra access just to restart jobs. However,
prior to this commit the only way to grant that access was by giving
them full Admin access which isn't necessarily what we want to do.

By having a restart-jobs role, we can grant this privilege to users
who are known to the community and want to help, but aren't long-time
members.

~I haven't tested this commit, but as a very novice Perl user, it looks good to me...~ @globin has tested this PR.

### DIFF
--- a/src/root/user.tt
+++ b/src/root/user.tt
@@ -80,6 +80,7 @@
           <select multiple="multiple" name="roles" class="span3" [% IF !c.check_user_roles('admin') %]disabled="disabled"[% END %]>
             [% INCLUDE roleoption role="admin" %]
             [% INCLUDE roleoption role="create-projects" %]
+            [% INCLUDE roleoption role="restart-jobs" %]
           </select>
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


src/root/user.tt
```diff
@@ -80,6 +80,7 @@
           <select multiple="multiple" name="roles" class="span3" [% IF !c.check_user_roles('admin') %]disabled="disabled"[% END %]>
             [% INCLUDE roleoption role="admin" %]
             [% INCLUDE roleoption role="create-projects" %]
+            [% INCLUDE roleoption role="restart-jobs" %]
```
grahamc: I _believe_ this role will be created on save: https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Controller/User.pm#L217-L218 and also I see nowhere in the code which explicitly adds the existing roles to the DB.

edolstra:
Thanks @grahamc!

Issue #440, opened 2017-01-06
Hydra JSON API


rzetterberg:
By demand on IRC, here's a quick overview of the Hydra JSON API created by looking at the [implementation](https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Controller/API.pm) and [tests](https://github.com/NixOS/hydra/blob/master/tests/api-test.pl).

When writing this I called each endpoint, analyzed the results and created an explanation. (I used the Emacs tool [restclient](https://github.com/pashky/restclient.el), if you want the restclient file to try out the API interactively in emacs, see the bottom of this issue.)

## Listing projects

To get a list of the hydra projects you simply request the root of the site and demand `application/json` as the content type:

```
curl -H "Content-Type: application/json" https://hydra.nixos.org
```

This will give you a JSON document where the root element is a list of objects that represents the projects.

Here's an shortened example of this output:

```json
[
  {
    "displayname": "cabal2nix",
    "description": "Convert Cabal files into Nix build instructions",
    "releases": [],
    "enabled": 0,
    "name": "cabal2nix",
    "owner": "simons@cryp.to",
    "hidden": 1,
    "jobsets": [
      "master"
    ]
  },
  {
    "displayname": "Hydra",
    "description": "Hydra, the Nix-based continuous build system",
    "releases": [],
    "enabled": 1,
    "name": "hydra",
    "owner": "eelco",
    "hidden": 0,
    "jobsets": [
      "hydra-master",
      "hydra-ant-logger-trunk",
      "master",
      "build-ng"
    ]
  }
]
```

## Getting project details

If you want to get the details of a specific project you call:

```
curl -H "Content-Type: application/json" https://hydra.nixos.org/project/hydra
```

This will give you a JSON document where the root element is an object that represents the requested project.

Note: This object is the same object you get in the project list.

Here's an example:

```json
{
  "displayname": "Hydra",
  "name": "hydra",
  "releases": [],
  "description": "Hydra, the Nix-based continuous build system",
  "enabled": 1,
  "owner": "eelco",
  "hidden": 0,
  "jobsets": [
    "hydra-master",
    "hydra-ant-logger-trunk",
    "master",
    "build-ng"
  ]
}
```

## Getting jobset details

If you want to get the details of a specific jobset you call:

```
curl -H "Content-Type: application/json" https://hydra.nixos.org/jobset/hydra/build-ng
```

This will give you a JSON document where the root element is an object that represents the detais of the requested jobset.

[...]

PR #529, opened 2018-01-14
feat: add /eval/<id>/builds endpoint

bennofs:
This endpoint allows efficient retrieval of all the builds in an
evaluation, without making a request for each single build.

### DIFF
--- a/src/lib/Hydra/Controller/JobsetEval.pm
+++ b/src/lib/Hydra/Controller/JobsetEval.pm
@@ -264,4 +264,15 @@ sub store_paths : Chained('evalChain') PathPart('store-paths') Args(0) {
 }
 
 
+# Return full info about all the builds in this evaluation.
+sub all_builds : Chained('evalChain') PathPart('builds') Args(0) {
+    my ($self, $c) = @_;
+    my @builds = $c->stash->{eval}->builds;
+    $self->status_ok(
+        $c,
+        entity => [@builds],
+    );
+}
+
+
 1;


bennofs:
Any feedback on this one? I am happy to pursue a different approach if this change is not wanted / should be implemented in a different way.

dtzWill:
I like this!

Can this produce JSON like api in #440? If not, do you know how difficult it would be to support that?

Hardly a barrier for entry but such an endpoint seems especially compelling when talking about thousands of builds -- many more than a human would ever issue requests for individually.

bennofs:
@dtzWill yes the format should be the same as that API. It should return exactly the same as querying the builds of the evaluation and then getting the build details for each build with the JSON api. This new endpoint is only necessary to avoid making 40k requests to get the info for  a single evaluation.

bennofs:
If this would result in too much load on the hydra server, we could also provide a cached version of this is part of a channel (just like store-paths)

Mic92:
cc @edolstra

Mic92:
@bennofs mhm, if we cannot get this in, maybe a travis cron job in nix-index would do.

Uploading built index files to the github release tab is not too hard: https://github.com/Mic92/cntr/blob/master/.travis.yml#L39

grahamc:
The way this works right now is not wonderful for hydra.nixos.org, taking almost 9 minutes and producing 80M of data:

```
[grahamc@Morbo:~]$ curl --header "Accept:application/json" https://hydra.nixos.org/eval/.../builds > latest-eval-builds
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 75.9M  100 75.9M    0     0   149k      0  0:08:40  0:08:40 --:--:-- 8008k
```

I'm not sure what to do about that, but just for what its worth.

grahamc:
Also the data includes magic numbers, like build status numbers: https://github.com/NixOS/hydra/blob/master/src/sql/hydra.sql#L202-L215

Mic92:
What information is actually needed from builds for `nix-index` to work?
Compression would also save 90%

```
$ du latest-eval-builds.xz
7.0M    latest-eval-builds.xz
```

PR #539, opened 2018-02-19
hydra-module: use new style of postgresql db/user creation

sorki:


sorki:
Fixed in a71e6301d222e265d755337576667e1b6c24a56c

PR #541, opened 2018-02-23
release.nix: Don't force gcc6

dtzWill:
When hydra's dependencies are built with gcc7
(particularly nix) using gcc6 doesn't work
as its libstdc++ is too old.

Using gcc6 was done in e0b2921ff2f714fd243a4ef3357c5d01fffba1a5
back in 2016 -- probably to use a /newer/-than-default gcc.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -103,8 +103,6 @@ rec {
 
       src = if shell then null else hydraSrc;
 
-      stdenv = overrideCC stdenv gcc6;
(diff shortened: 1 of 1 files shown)


dtzWill:
https://github.com/NixOS/hydra/commit/c0fac5287254b203a5ea26ffffa07a09e465bbc3 .

PR #552, opened 2018-03-28
nix-prefetch-git: Fix git submodule checkout

joranvar:
When fetching git submodules, nix-prefetch-git would fail on line 160 with "cd: too many arguments".  I'm not sure if the `git submodule status` output has changed, or whether init_submodules() is called in different situations where it does work.  But this change works for me (the submodule is checked out correctly), so it may benefit someone else as well.

**Separate git submodule dir from `git describe` part when parsing**

**Ignore first character of `git submodule status` output**
The submodule may possibly already be initialized, thus lacking the
initially expected dash (`-`).

### DIFF
--- a/src/script/nix-prefetch-git
+++ b/src/script/nix-prefetch-git
@@ -142,8 +142,8 @@ init_submodules(){
     git submodule status |
     while read l; do
         # checkout each submodule
-        local hash=$(echo $l | sed 's,^.\([0-9a-f]*\) \(.*\) (.*)$,\1,');
-        local dir=$(echo $l | sed 's,^.\([0-9a-f]*\) \(.*\) (.*)$,\2,');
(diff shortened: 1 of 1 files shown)


aycanirican:
@joranvar This worked like a charm, thank you. cc @edolstra

wavewave:
:+1:

joranvar:
I ran into an issue today where the complete commit hash was needed.  As the skipped character may not be just white space (it may get trimmed off by bash), I constrained the skipping to just the expected non-hash characters.  Hope it still works for you!

jokogr:
I have tested this PR, too, and it seems to be working properly, thanks!

kquick:
With all due respect to this solution from @joranvar, I have just submitted https://github.com/NixOS/hydra/pull/570 which is a superset of this PR, fixing some other issues in nix-prefetch-get by taking a bit of a different approach.  I would appreciate confirmation from anyone that tested this PR (#552) that #570 also resolves the issues they encountered.   Thanks!

PR #561, opened 2018-05-17
Update Perl dependencies

lopsided98:
Add missing Perl dependencies. This change is copied from NixOS/nixpkgs#39314

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -62,6 +62,7 @@ rec {
             CatalystViewDownload
             CatalystViewJSON
             CatalystViewTT
+            CatalystXRoleApplicator
             CatalystXScriptServerStarman
             CryptRandPasswd
             DBDPg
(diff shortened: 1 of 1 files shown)


dtzWill:
:+1: thanks!

Adding these shouldn't cause problems when building against "older" nixpkgs right, since it's a matter of explicitly listing existing dependencies already used (instead of assuming they're pulled in implicitly by other things)? (similar to include-what-you-use re:headers and such)

lopsided98:
Both [`CatalystXRoleApplicator`](https://github.com/NixOS/hydra/blob/master/src/lib/Hydra.pm#L22) and [`JSONAny`](https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Controller/API.pm#L11) are direct dependencies of Hydra and were added to nixpkgs in 2013. I assume something else used to depend on them, but no longer does.

lopsided98:
Already done by https://github.com/NixOS/hydra/commit/3ef253c65f03cc7ede70ca753cd654eb6f771635.

PR #569, opened 2018-06-29
Show evaluation errors

kquick:
Display evaluation errors on the project summary page and also on the Status:Latest Evaluations page.

In the original form, processing of a jobset can be affected by evaluation errors, but there is no indication of an evaluation error unless you visit the specific jobset page.  This modification makes these evaluation errors visible on top-level summaries (in the same places that build failures are shown) so that there is visible indication of a problem that should be fixed.

If there are no errors, there should be no difference in the display from the current output.  This only shows additional information when there are evaluation errors.

### DIFF
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -341,10 +341,20 @@ sub evals :Local Args(0) {
 
     my $evals = $c->model('DB::JobsetEvals');
 
+    my @failEvals = [$c->model('DB::Jobsets')
+                       ->search({ "me.enabled" => 1, "me.hidden" => 0,
+				  "project.enabled" => 1, "project.hidden" => 0 })
+		       ->search({ -or => [ errormsg => { '!=' => '' },
+					   fetcherrormsg => { '!=' => '' }
+				      ]},
+				{ order_by => ["project", "name"], join=>["project"] }
+		     )];
+
     $c->stash->{page} = $page;
     $c->stash->{resultsPerPage} = $resultsPerPage;
     $c->stash->{total} = $evals->search({hasnewbuilds => 1})->count;
-    $c->stash->{evals} = getEvals($self, $c, $evals, ($page - 1) * $resultsPerPage, $resultsPerPage)
+    $c->stash->{evals} = getEvals($self, $c, $evals, ($page - 1) * $resultsPerPage, $resultsPerPage);
+    $c->stash->{evalfailJobsets} = $_ foreach @failEvals;
 }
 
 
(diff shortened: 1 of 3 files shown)


src/lib/Hydra/Controller/Root.pm
```diff
@@ -341,10 +341,20 @@ sub evals :Local Args(0) {
 
     my $evals = $c->model('DB::JobsetEvals');
 
+    my @failEvals = [$c->model('DB::Jobsets')
+                       ->search({ "me.enabled" => 1, "me.hidden" => 0,
```
nlewo: There are some indentation issues here.
Also, it may be better to do only one `search` call, as it is implemented in the `search` method below (while I'm really not sure about this:/).

kquick: I'm not sure quite what you mean about indentation "issues".  There are multiple dereferences, and each reference is somewhat long, so this separates each dereference to a separate line. The code works just fine (I run it locally).  Was your concern about functionality or just appearance, and did you have a specific recommendation regarding appearance?

In my experience it is typical for ORMs to perform local optimization of the generated SQL statement, and multiple refinements like this are often recommended as a way to specialize a base query.  Additionally the SQL server performs optimization on the query, so I don't expect this will result in any performance problems, especially when used with postgres.

grahamc:
I feel a bit scared about adding this to the load. These pages already do pretty heavy pageloads on hydra.nixos.org -- adding more, heavy queries seem spooky.

grahamc:
I feel a bit scared about adding this to the load. These pages already do pretty heavy pageloads on hydra.nixos.org -- adding more, heavy queries seem spooky.

src/lib/Hydra/Controller/Root.pm
nlewo: Ok, forget my comment about indentation and thx for your answer!

nlewo:
@grahamc Actually, one commit (https://github.com/NixOS/hydra/pull/569/commits/e7a7bd58bc154af4dee5468a6af67e2dbd5a85de) of this PR should not increase the load and is quite nice (it shows evaluation errors from the project view).
I'm using this commit on my Hydra instance and it is really useful.
What about merging this commit?

Regarding the second commit, if we really want this feature, we could add a configuration variable to enable this feature. It would be disable by default and people wanting this feature could explicitly enable it.

edolstra:
Thanks, I've cherry-picked commit e7a7bd5.

gilligan:
Considering the concerns on the one hand, and the fact that parts were cherry picked i’ll close this PR.

Thank you for the contribution.

PR #571, opened 2018-06-30
Additional helpful information in error messages.

kquick:


src/lib/Hydra/Helper/Nix.pm
```diff
@@ -425,7 +425,7 @@ sub run {
 sub grab {
     my (%args) = @_;
     my $res = run(%args, grabStderr => 0);
-    die "command `@{$args{cmd}}' failed with exit status $res->{status}" if $res->{status};
+    die "command `@{$args{cmd}}' failed with exit status $res->{status} in $args{dir}" if $res->{status};
```
grahamc: is `dir` guaranteed to be set?

dasJ: From the same file: `return grab(cmd => ["nix", "cat-store", "--store", getStoreUri(), "$path"]);`

kquick: Thanks for catching this.  I don't normally work in Perl, so I missed the hindmost check in https://github.com/NixOS/hydra/blob/1e804d1b8a1a258c36d85a8e9afca42ea1539fb9/src/lib/Hydra/Helper/Nix.pm#L398

PR #575, opened 2018-07-15
jobset view: jump back to the proper tab when displaying all failed jobsets

Ma27:
When I press "n builds omitted" I get back to the first tab of a jobset.
This is extremely counter-intuitive, instead this notice should link to
the currently opened tab.


Ma27:
@edolstra this bug can be very confusing. Would be cool if you could merge this:)

edolstra:
Thanks!

PR #577, opened 2018-07-16
Fix perl dependencies

Taneb:
I've not really used perl in anger so I might have done something wrong here, but I had to do this to get `hydra-server` working.


Taneb:
This has been fixed in 3ef253c65f03cc7ede70ca753cd654eb6f771635 and 6d693b6b2a028269a029b679a81adea41b431044, I think

Issue #580, opened 2018-07-21
Build product metadata missing


lheckemann:
In https://hydra.nixos.org/build/77875732, the metadata (size, hashes) for the build products is missing, although downloading them works fine. I'm not sure why this is, but I suspect that it's because the files in the derivation are symlinks.

The metadata should be accessible behind the "Details" button for each build product, and shown in the JSON as well (`curl -H 'Accept: application/json' https://hydra.nixos.org/build/77875732`).

PR #581, opened 2018-07-21
[on hold] Queue runner symlinks

lheckemann:
Adds a test demonstrating #580 and hopefully fixes the issue fully too.


lheckemann:
This isn't sufficient to make netboot files immediately available because `nix cat-store` also can't deal with symlinks. Since that requires the same sort of logic, it would make sense to factor the symlink resolution code out into libstore and use it both from here and from cat-store, so this may be considered as on hold for now.

PR #588, opened 2018-08-13
Consumable system features

### LINKED ISSUE #585: Consumable system features
Hydra clone of https://github.com/NixOS/nix/issues/2307, since Hydra has its own scheduler independent of nix.

I've already started working on this, but I'm not terribly well versed in C++, so definitely feel free to preempt me on this ^^

ledettwy:
Adds support for consumable system features.

Remote build machines can specify a supported or mandatory feature as
consumable with a "feature:n" notation, where n is the quantity of the
feature that the machine provides.  "feature:0" and "feature" are both
interpreted as a non-consumable (infinite) feature.

Derivations can specify how much of a feature they consume with a
similar "feature:n" notation, where n is the quantity of the feature
that the derivation consumes.  "feature" is interpreted as "feature:1"
(i.e., consumes one of the feature), and "feature:0" is interpreted as
still requiring the feature but not consuming any of it.

Closes #585

I tested this locally with a variety of different feature combinations and it behaves correctly.

Also, I am not a C++ programmer. A lot of this may very well be written in a non-optimal or non-idiomatic way. If there's a better way to write any of the code I've got here, please let me know.

Design notes are at https://github.com/NixOS/nix/issues/2307, feedback / discussion on the design should happen there.

There's an obvious corresponding implementation in NixOS/nix, but that will probably have to be authored by someone who's more familiar with the codebase than I am.

### DIFF
--- a/src/hydra-queue-runner/dispatcher.cc
+++ b/src/hydra-queue-runner/dispatcher.cc
@@ -340,6 +340,9 @@ State::MachineReservation::MachineReservation(State & state, Step::ptr step, Mac
 {
     machine->state->currentJobs++;
 
+    for (auto & f : step->requiredSystemFeatures)
+        machine->state->featuresConsumption[f.first] += f.second;
+
     {
         auto machineTypes_(state.machineTypes.lock());
         (*machineTypes_)[step->systemType].running++;
@@ -353,6 +356,9 @@ State::MachineReservation::~MachineReservation()
     assert(prev);
     if (prev == 1)
         machine->state->idleSince = time(0);
+
+    for (auto & f : step->requiredSystemFeatures)
+        machine->state->featuresConsumption[f.first] -= f.second;
 
     {
         auto machineTypes_(state.machineTypes.lock());
--- a/src/hydra-queue-runner/hydra-queue-runner.cc
+++ b/src/hydra-queue-runner/hydra-queue-runner.cc
@@ -1,5 +1,6 @@
 #include <iostream>
 #include <thread>
+#include <climits>
 
 #include <sys/types.h>
 #include <sys/stat.h>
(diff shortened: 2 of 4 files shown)


ledettwy:
don't know how to add people for reviewing... so just tagging @edolstra @shlevy @rbvermaa

Intent is that this would also make its way into Nix eventually... either when I have time or probably @edolstra has time.  We should make sure that the notation especially wouldn't break anything in Nix before the hydra part gets merged (assuming people actually think this is useful enough to merge :^))

Any objections to the hydra part getting merged before the corresponding Nix patch is ready? This shouldn't break anything in terms of backwards compatibility... Nix already doesn't allow ":" characters in system feature names, so AFAIK no one is using it already incidentally for some other meaning.

Taneb:
I'd be very happy if this could get reviewed, I've been looking for pretty much exactly like this change for work.

nlewo:
@ledettwy Thanks a lot for this!

I've rebased this PR onto master and also added more status reporting in this branch:
 https://github.com/nlewo/hydra/tree/consumable-feature

Thanks to consumable features, my NixOS tests can specify how many CPUs they use. Hydra will schedule them on build slaves. Without this feature, these tests often "timeouts" because there is not enough CPUs to run them.

There is still an issue on the Nix side since the Nix configuration must contain all feature amounts:/ For example, in a NixOS test deviation, I have:
```
stdenv.mkDerivation {
  ...
  ...
   requiredSystemFeatures = [ "kvm" "nixos-test" "cpu:2" ];
}
```
The `/etc/nix/machines` file contains:
```
localhost x86_64-linux,builtin - 8 1 kvm,nixos-test,big-parallel,cpu:8
```
And the `/etc/nix/config.nix` file:
```
system-features = kvm nixos-test cpu:2
```

nlewo:
FYI, I finally added the support to specify consumables feature by repeating the feature keyword in d6d31acd1d166533c585455a1a82054fe64278f6. For instance `requireSystemFeatures = [ "kvm" "kvm" ]` is equivalent to `requireSystemFeatures = [ "kvm:2" ]`. 
The former is ignored by Nix so you can still build such kind of derivations locally. To build derivation with `kvm:2`, we would need to add the `system-feature` `kvm:2` to your local Nix configuration (since commit 1e7b8de).

To have a correct support of consumable features in Hydra, I think we have to patch Nix to either support consumable features, or ignore them... because right now, it's a bit hacky!

PkmX:
+1 for this feature. We also need this for controlling scheduling tests onto physical hardware that only allow one job at a time.

PR #590, opened 2018-08-21
Allow for precisely one instance of RunCommand plugin

Taneb:
The current code expects `$self->{config}->{runcommand}` to be an array if it exists at all. This isn't the case when there's precisely one `runcommand`


edolstra:
Thanks!

Issue #512, opened 2017-10-10
With outside PRs, githubstatus POSTs the wrong repo


taktoa:
Suppose you set up a Hydra declarative jobset to build PRs using the `githubpulls` input on a GitHub repo named `alice/foo`. You then enable the Hydra `githubstatus` plugin with the intention of making it possible for contributors to see the build status easily. This requires you to give it a GitHub auth token with `repo:status` permission on an account that has push access to `alice/foo`. Now suppose a pull request comes in from `bob/foo`. Currently, the githubstatus will POST the commit status to `alice/foo` instead of `bob/foo`, and since `alice/foo` doesn't have that commit, it seems to fail.

Example: commit statuses are visible on local commits [here](https://github.com/zenhack/haskell-capnp/commits/master), but are not visible on this [pull request](https://github.com/zenhack/haskell-capnp/pull/14) despite a [jobset](https://hydra.angeldsis.com/jobset/haskell-capnp/zenhack-master-pr-14) existing.

I suspect this issue was not noticed because many of the users of `githubstatus` are making feature branches _on_ `alice/foo` and then doing a pull request against `master`, but this feature is pretty much essential to making Hydra usable for everyday open source projects on GitHub (as opposed to big industrial projects).

domenkozar:
cc @cleverca22 @shlevy

shlevy:
The repo is determined by the relevant jobsetinput: https://github.com/NixOS/hydra/blob/45b138373bf4b5b1ed677d16d3ee172284fe0dc0/src/lib/Hydra/Plugin/GithubStatus.pm#L59

shlevy:
In fact, I'm now wondering if the problem isn't the exact opposite of what you suggest: You don't have permission to set status on `bob`, but if you instead tried setting status on `alice` it would work.

taktoa:
That sounds plausible; I haven't looked too deeply into the issue, so the explanation given in the initial post is nothing more than a hypothesis based on the observed symptoms.

shlevy:
My guess is adding something to the GithubStatus config to hard-code owner and repo would do the trick.

domenkozar:
Afaik most CIs merge the PR and then post the status on the merge commit build. Otherwise, you might not be able to attach the status on `$rev` since it's not part of the repo. Hopefully, I'm mistaken :)

shlevy:
[...]

PR #592, opened 2018-09-04
Add GitlabPulls input plugin

nlewo:


nlewo:
cc @shlevy

Shados:
@nlewo in light of #512, I would suggest changing the plugin slightly to something like [this](https://gist.github.com/Shados/34939268a6d30589d017b16a93abb088).

This also includes support for using a personal access token (PAT) in order to access protected or private repositories. Using the *target* repository URL along with the merge-request ref instead of the source repository url and branch is necessary to avoid running into issues if the source repository is not actually accessible to the user Hydra is authenticating as.

I'm running this modified version of the plugin internally at work, and haven't had any issues thus far.

nlewo:
@Shados Thanks for your patch. I applied it and it works well.
@edolstra Could you please take a look?

Ericson2314:
I have used this and can confirm is does the thing. Please merge!

Ericson2314:
Thanks @shlevy!

PR #594, opened 2018-09-10
Set notification status to done for dependent builds.

kquick:
When a build completes with error, the dependent builds are updated
with errors as well, and notifications are sent for all these builds,
but the DB wasn't setting the notification completion indication for
the dependent builds.  One effect of this is that on hydra restart,
notifications would *again* be sent for all dependent build failures,
causing duplicate notifications, many of which were often very old.

### DIFF
--- a/src/hydra-queue-runner/builder.cc
+++ b/src/hydra-queue-runner/builder.cc
@@ -429,11 +429,14 @@ State::StepResult State::doBuildStep(nix::ref<Store> destStore,
                     if (build2->finishedInDB) continue;
                     printMsg(lvlError, format("marking build %1% as failed") % build2->id);
                     txn.parameterized
-                        ("update Builds set finished = 1, buildStatus = $2, startTime = $3, stopTime = $4, isCachedBuild = $5, notificationPendingSince = $4 where id = $1 and finished = 0")
+                        ("update Builds set finished = 1, buildStatus = $2, startTime = $3, stopTime = $4, isCachedBuild = $6, notificationPendingSince = $5 where id = $1 and finished = 0")
                         (build2->id)
                         ((int) (build2->drvPath != step->drvPath && result.buildStatus() == bsFailed ? bsDepFailed : result.buildStatus()))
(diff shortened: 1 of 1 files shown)


grahamc:
I'm not really able to determine if this is a proper fix, or more of a hack?

kquick:
Thank you for taking a look at this PR, @grahamc. This is intended to be a proper fix.  

The code following this block will send the notifications for this build and all of its dependents:
https://github.com/NixOS/hydra/blob/1e804d1b8a1a258c36d85a8e9afca42ea1539fb9/src/hydra-queue-runner/builder.cc#L466-L471

This modification ensures that the `notificationPendingSince` is set to zero (the zero value indicates that the notification has been sent).  If the value is non-zero, then the code below will see that there is a pending notification and repeat the notification attempts done in 466-471 above, causing duplicate notifications:
https://github.com/NixOS/hydra/blob/1e804d1b8a1a258c36d85a8e9afca42ea1539fb9/src/hydra-queue-runner/hydra-queue-runner.cc#L850-L861

This fix is based on our local running hydra and the complaints of users, which this fix resolved.  The alternative would be to remove lines 466-471 in builder.cc and let the hydra-queue-runner.cc handle the notifications, but it does seem to communicate the intent more directly to instigate those notifications in 466-471.

grahamc:
I have a new appreciation for this PR after reviewing the notification work. I'm not sure this PR is the right thing: it seems to me that the right behavior is:

1. notify that the top level build finished, with the dependent IDs
2. also send a unique notification per build ID

then let hydra-notify handle the per-build marking as notified. If this isn't the right behavior, then maybe hydra-notify should run the handlers per ID in the payload, and not just pass the extra IDs as an additional argument.

kquick:
Independently addressed in https://github.com/NixOS/hydra/commit/4417f9f260b17cf2d4098df2a6a4b858856b76ea

grahamc:
Interesting. Okay :) :shrug:

PR #603, opened 2018-10-18
Added job/PROJECT/JOBSET/JOB/latest-outpath endpoint.

kquick:
Returns a redirect to the first output path of the latest successful
build of the specified job. This can be useful when the output
provides HTML documentation (e.g. generated documentation for a build
such as Haskell haddock output).  By providing a "constant" redirect
link, the hydra machine can also serve the documentation it builds via
an href in a static page.

### DIFF
--- a/src/lib/Hydra/Base/Controller/ListBuilds.pm
+++ b/src/lib/Hydra/Base/Controller/ListBuilds.pm
@@ -58,6 +58,20 @@ sub latest : Chained('get_builds') PathPart('latest') {
 }
 
 
+# Redirect to the latest successful build's first output path (useful
+# if the output is HTML (e.g. haddock documentation).
+sub latest_outpath : Chained('get_builds') PathPart('latest-outpath') {
+    my ($self, $c, @rest) = @_;
+
+    my $latest = $c->stash->{allBuilds}->find(
+        { finished => 1, buildstatus => 0 }, { order_by => ["id DESC"], rows => 1 });
+
+    notFound($c, "There is no successful build to redirect to.") unless defined $latest;
+
+    $c->res->redirect(($latest->buildoutputs)[0]->path);
+}
+
+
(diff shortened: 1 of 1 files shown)


aycanirican:
@kquick this is cool, thank you.

nomeata:
I am confused about this PR. Isn’t this functionality already in hyda?

kquick:
@nomeata Not as far as I could tell... do you know of a URL that provides this?

nomeata:
Oh, I see, this is output paths, not declared built artifacts like https://github.com/NixOS/hydra/blob/879333fd3ce481460fda74a19e26d5e16ff008f8/src/lib/Hydra/Base/Controller/ListBuilds.pm#L57

Nevermind then.

basvandijk:
I tried this PR but I get the following:

When requesting the `latest-outpath` URL:

https://my.hydra.com//job/my-project/my-jobset/my-job/latest-outpath

it redirects me to:

https://my.hydra.com//nix/store/n3zsgkc28nw1b8bddinww50d9bb55kcp-my-output

which leads to a 404 because I guess hydra doesn't know how to serve `/nix/store/...` paths.

Do I need to setup something extra to serve `/nix/store/...` paths?

kquick:
Yes, thanks for the reminder on this, Bas.

I'm using `services.httpd` to proxy for hydra on the same machine as well as provide paths for haddock docs and hoogle services generated by hydra (and thus this patch to allow serving of the "latest").  To allow the above redirection, I have the following clause:

    services.httpd = {
        servedDirs = [
            { dir = "/nix/store";
              urlPath = "/nix/store/";
            };
        ];
        ...
    };

Although it should be possible to serve the file directly via hydra, I'm not sure that's a good idea because it essentially bypasses a lot of the controls that a webserver proxy can provide.

PR #622, opened 2018-12-12
Add sha256hash to input attrs for eval.

kquick:
This makes the sha256 hash of the input available to the evaluation process, which is useful in cases where this evaluation generates a fetchgit for a subsequent build.


grahamc:
Can you share an example usage?

minijackson:
I'm needing this. My usecase is that I want to generate a channel that follows multiple git repositories. Because the packages need to be reproducible from source, I need a `fetchgit` for each package in the channel. So what I've figured I should do is in the `patchPhase` of my channel job, I generate the arguments for each of my package's `src`'s `fetchgit`. Since this is Hydra is making this pure, I must deduce these parameters from Hydra's args. (Please tell me if the wording is confusing)

kquick:
Whoops, forgot about this one.  The use case I had was very similar to what @minijackson describes although it was related to git fetching an input without submodules to drive cabal2nix generation of nix descriptions for haskell packages obtained from git repos.

Since that time however we shifted strategy and no longer directly need this input, but it still seems like this could be generally useful.  It's not secret information and it's already gathered by the eval process... it's just a matter of passing this as part of the input information to the job.

PR #629, opened 2018-12-29
hydra-eval-jobs: fix maintainer resolution

Ma27:
Some time ago the data structure for maintainer descriptions in
`nixpkgs` changed from a simple attr set with maintainer emails as
values to an attribute set where the maintainer' nick is associated to
an attribute set with email, GitHub handle and full name.

Hydra can either parse a Nix list or fetches `shortName` from the
associated attribute set (which is used for `meta.licenses` as each
value in it contains a `shortName`). This behavior needs to be
replicated for maintainers to retrieve the emails for `hydra-notify`.

This change is backwards-compatible since `queryMetaStrings` is still
able to understand lists, so old versions of `nixpkgs` or packages using
the old maintainer data structure remain usable.

### DIFF
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -29,7 +29,7 @@ static void findJobs(EvalState & state, JSONObject & top,
     Bindings & autoArgs, Value & v, const string & attrPath);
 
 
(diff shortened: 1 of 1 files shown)


grahamc:
Thanks, verified it works :) https://hydra.nixos.org/build/90818518#tabs-details

Issue #393, opened 2016-09-14
Friendlier times in Hydra


copumpkin:
The current timestamps all across the UI are lacking time zone indicators so it's hard to know when things actually happen. To up friendliness further, it would be nice to use relative times (with some sort of precision falloff as they get bigger) so I can learn that the last evaluation was e.g., "3 hours ago" (a tooltip would tell me exact time with timezone) and I don't have to mentally subtract times.

PR #631, opened 2019-01-22
Enhances user-facing dates

samueldr:
With these changes, dates will be automatically switched to the user's timezone with JavaScript. Users without JavaScript enabled will have the previous behaviour, with the addition of a `title` which holds a fuller date, which also shows the timezone name. Additionally, relative times will always show more details in the title.

The user-facing dates, when enhanced, also present an enhanced title test, here's an example:

```
Adjusted to local time (-05:00)
Other timezones:
  UTC: 2018-12-02 01:39:40
  As Hydra reported: 2018-12-01 20:39:40 (EST)
```

The first line shows the user's timezone's offset, and also shows the date in UTC and hydra-local timezone.

##### Motivations

I have been burned by the hydra.nixos.org instance, assuming the dates were UTC, since there was no timezone label at all, and no setting available.

With this change, it is harder for an end-user to make bad assumptions about dates, EXCEPT for users which were used to timezone offsets. This is helped by introducing an underline to the dates, hoping the users then notice the cursor change and the title text.

Additionally, iconography could be added to try an better indicate the change, but I fear this would be too much.

Finally, relative dates haven't been touched. It could be possible to hook a live-updater to the relative times, but I don't want to introduce other third-party code. Making it ourselves could work, but for the investment of time (not that much) I'd prefer a clear go-ahead.

* * *

I think this fully addresses #393 cc @copumpkin


LnL7:
Works on my machine™.

hydra-dates

samueldr:
Not quite, the previous "works on my machine™" fails. The *UTC* field of the tooltip mutated the date object; this field was a late addition to the project. [The fix was trivial](https://gist.github.com/49747b8b52b654af21c26d133504dc7b).

With the fix:

image

PR #632, opened 2019-01-22
hydra-queue-runner: better error message if nix-store can not be started

nlewo:
The hydra-queue-runner opens a connection to the builder. If the
builder is 'localhost' it starts `nix-store`, otherwise it starts
'ssh'.

Currently, if the hydra-queue-runner can not start `nix-store` (not in
the PATH for instance), the error message is:

  cannot connect to ‘localhost’: error: cannot start ssh: No such file
  or directory

This is not useful since ssh is actually not started:/

With this patch the error message is now:

  cannot connect to ‘localhost’: error: cannot start nix-store: No such file
  or directory

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -29,6 +29,7 @@ static void append(Strings & dst, const Strings & src)
 
 static void openConnection(Machine::ptr machine, Path tmpDir, int stderrFD, Child & child)
 {
+    string pgmName;
     Pipe to, from;
     to.create();
     from.create();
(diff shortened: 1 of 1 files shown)


src/hydra-queue-runner/build-remote.cc
```diff
@@ -66,7 +70,7 @@ static void openConnection(Machine::ptr machine, Path tmpDir, int stderrFD, Chil
 
         execvp(argv.front().c_str(), (char * *) stringsToCharPtrs(argv).data()); // FIXME: remove cast
 
-        throw SysError("cannot start ssh");
+        throw SysError(format("cannot start %1%") % pgmName);
```
edolstra: BTW it's shorter to write `throw SysError("cannot start %s", pgmName)`.

nlewo: Thx, applied.

PR #607, opened 2018-10-31
Add JSON search API endpoint 

nlewo:
I really don't know how to let Catalyst exports the result of a query as JSON. So, I've tried a way... but it could be completely wrong:/

PR #636, opened 2019-02-14
Serialize data as JSON with `Accept: application/json`

Ma27:
Similar to #607. According to the Catalyst[1] docs it's possible to
specify a data structure that is supposed to be serialized when
requesting i.e. a JSON response.

[1] https://metacpan.org/pod/Catalyst::Controller::REST#status_ok


grahamc:
Tested via `curl -H "Accept: application/json" https://hydra.nixos.org/machines`, lgtm, thank you!

PR #646, opened 2019-03-22
Microsoft Teams notification plugin

mray-infor:
Adds a new notification plugin for Microsoft Teams, for those of us who aren't on Slack.

I refactored the existing SlackNotification.pm into a pseudo-abstract class called RichMessengerNotification.pm (I don't think Perl has "real" abstract classes, but I could be wrong), then had both SlackNotification.pm and the new MicrosoftTeamsNotification.pm inherit from it.

I have no prior experience with Perl, _definitely_ no experience with Perl inheritance, and no clear idea of how to test this thing. Any advice would be appreciated.

### DIFF
--- a/src/lib/Hydra/Plugin/MicrosoftTeamsNotification.pm
+++ b/src/lib/Hydra/Plugin/MicrosoftTeamsNotification.pm
@@ -1,4 +1,4 @@
-package Hydra::Plugin::SlackNotification;
+package Hydra::Plugin::MicrosoftTeamsNotification;
 
 use strict;
 use parent 'Hydra::Plugin';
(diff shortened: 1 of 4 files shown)


grahamc:
I wonder if for simplicity and clarity, it might be worth not trying to abstract them?

mray-infor:
@grahamc You mean, just duplicate 70% of the SlackNotification.pm code in MicrosoftTeamsNotification.pm? Or move the common code into some sort of imported utility class instead of trying to use inheritance?

mray-infor:
I gave up on the fancy inheritance structure, since I'm having trouble even building Hydra to test the code to see what went wrong. Instead I just let the MSTeams plugin duplicate some code (ugh) from the Slack plugin. Though I've factored out a few common operations into a separate util file, to make it less horrible.

Mic92:
Running hydra should be easier since: https://github.com/NixOS/hydra/pull/759

PR #648, opened 2019-03-29
tests: change postgresql socket dir to /tmp

danbst:
In https://github.com/NixOS/nixpkgs/pull/57677 default postgresql socket directory was changed
to `/run/postgresql`, which doesn't exist (and can't be created) in
Nix build environment. We'll use /tmp as socket dir explicitly then.

Fixes build failure https://hydra.nixos.org/build/91221682

Cc @aszlig @edolstra

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ Makefile.in
 /inst
 hydra-config.h
 hydra-config.h.in
+result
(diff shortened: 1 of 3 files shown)


flokli:
That patch is happily referenced from nixpkgs since 8e18f4e4143184f667de3ffb3fc958f2e34b6e29.

It would be nice if could merge it to master and drop the patch from nixpkgs again, or is there anything blocking this from being merged?

PR #649, opened 2019-04-11
Create a gitlab status plugin

globin:
This plugin expects as inputs to a jobset the following:
 - gitlab_status_repo => Name of the repository input pointing to that
   status updates should be POST'ed, i.e. the jobset has a git input
   "nixexprs": "https://gitlab.example.com/project/nixexprs", in which
   case "gitlab_status_repo" would be "nixexprs".
 - gitlab_project_id => ID of the project in Gitlab, i.e. in the above
   case the ID in gitlab of "nixexprs"

cc @ericson2314 @grahamc 

See https://git.mayflower.de/eso/test-libs/pipelines/4794 and https://hydra-eso-test.yflower.de/jobset/test-libs/pkgs-mr-3 for an example.

### DIFF
--- a/src/lib/Hydra/Plugin/GitlabStatus.pm
+++ b/src/lib/Hydra/Plugin/GitlabStatus.pm
@@ -1,0 +1,88 @@
+package Hydra::Plugin::GitlabStatus;
+
+use strict;
+use parent 'Hydra::Plugin';
+use HTTP::Request;
+use JSON;
+use LWP::UserAgent;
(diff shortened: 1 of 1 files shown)


edolstra:
This plugin should probably be explicitly enabled by a hydra.conf option, because currently it's doing some non-trivial amount of work just to decide whether it needs to do anything (e.g. query `$build->jobsetevals`, look for a `gitlab_status_repo` input, and this for all dependents). That's quite a bit of overhead when we're not using this plugin.

PR #654, opened 2019-06-05
Allow to search builds by hash

nlewo:
Currently, a full store path has to be provided to search in
builds. This patch permits to search jobs with a output path or
derivation hash.

Usecase: we are building Docker images with Hydra. The tag of the
Docker image is the hash of the image output path. This patch would
allow us to find back the build job from the tag of a running
container image.

### DIFF
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -422,11 +422,11 @@ sub search :Local Args(0) {
 
     # Perform build search in separate queries to prevent seq scan on buildoutputs table.
     $c->stash->{builds} = [ $c->model('DB::Builds')->search(
-        { "buildoutputs.path" => trim($query) },
+        { "buildoutputs.path" => { ilike => "%$query%" } },
         { order_by => ["id desc"], join => ["buildoutputs"] } ) ];
 
     $c->stash->{buildsdrv} = [ $c->model('DB::Builds')->search(
-        { "drvpath" => trim($query) },
+        { "drvpath" => { ilike => "%$query%" } },
         { order_by => ["id desc"] } ) ];
 
     $c->stash->{resource} = { projects => $c->stash->{projects},
--- a/src/sql/hydra.sql
+++ b/src/sql/hydra.sql
@@ -688,3 +688,9 @@ create index IndexBuildsOnKeep on Builds(keep) where keep = 1;
 create index IndexJobsetEvalsOnJobsetId on JobsetEvals(project, jobset, id desc) where hasNewBuilds = 1;
 
 create index IndexBuildsOnNotificationPendingSince on Builds(notificationPendingSince) where notificationPendingSince is not null;
+
+#ifdef POSTGRESQL
+-- Provide an index used by LIKE operator on builds.drvpath (search query)
+CREATE EXTENSION pg_trgm;
+CREATE INDEX IndexTrgmBuildsOnDrvpath ON builds USING gin (drvpath gin_trgm_ops);
+#endif
--- a/src/sql/upgrade-57.sql
+++ b/src/sql/upgrade-57.sql
@@ -1,0 +1,2 @@
+CREATE EXTENSION pg_trgm;
+CREATE INDEX IndexTrgmBuildsOnDrvpath ON builds USING gin (drvpath gin_trgm_ops);


nlewo:
Note I really don't know what could be the performance impact on search queries on hydra.nixos.org... Of course, on my Hydra instances, it's negligible.

edolstra:
Yeah it seems that this could require a sequential scan on the `Builds` table. Maybe you can do an `explain analyze` to see what the query is doing?

nlewo:
I confirm using a regex to search by `drvPath` requires a seq scan of the `builds` table while it is not the case with `=`. Regarding the search by `path`, it seems to me it already requires a seq scan of the `builds` table.
```
postgres=# explain analyze select * from builds,buildoutputs where buildoutputs.path = '/nix/store/3q84b8jzsnckipx5gxqzd6wnvzyn0wml-trivial';
 Nested Loop  (cost=0.00..39.38 rows=600 width=560) (actual time=0.137..0.192 rows=1 loops=1)
   ->  Seq Scan on builds  (cost=0.00..11.50 rows=150 width=492) (actual time=0.013..0.021 rows=1 loops=1)
   ->  Materialize  (cost=0.00..20.39 rows=4 width=68) (actual time=0.031..0.054 rows=1 loops=1)
         ->  Seq Scan on buildoutputs  (cost=0.00..20.38 rows=4 width=68) (actual time=0.013..0.021 rows=1 loops=1)
               Filter: (path = '/nix/store/3q84b8jzsnckipx5gxqzd6wnvzyn0wml-trivial'::text)
```

nlewo:
@edolstra We could create indexes supporting `like` operator with the module https://www.postgresql.org/docs/9.3/pgtrgm.html. I locally tried it and I get the following:
```
postgres=# explain analyze select * from builds where drvpath like '%k3r71gz0gv16ld8rhcp2bb8gb5w1xc4b%';
                                                            QUERY PLAN                                                            
----------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on builds  (cost=124.00..128.01 rows=1 width=492) (actual time=0.065..0.084 rows=1 loops=1)
   Recheck Cond: (drvpath ~~ '%k3r71gz0gv16ld8rhcp2bb8gb5w1xc4b%'::text)
   ->  Bitmap Index Scan on trgm_idx_users_username  (cost=0.00..124.00 rows=1 width=0) (actual time=0.038..0.038 rows=1 loops=1)
         Index Cond: (drvpath ~~ '%k3r71gz0gv16ld8rhcp2bb8gb5w1xc4b%'::text)
 Total runtime: 0.146 ms
(5 rows)
```

What do you think?

nlewo:
I added the creation of this index in the commit 7935cff.

edolstra:
How much disk space does the trigram index need?

nlewo:
Hm, I didn't find clear doc about the disk usage of this kind of
index. So, I did some small experiments:

In the following, I ran Postgresql 11.3 on my laptop (4 cores, 16G RAM).
I inserted `10 000 000` rows containing `/nix/store/<UUID>` using the following script:
```
CREATE TABLE Builds (
    drvPath       text not null
);
INSERT INTO builds (drvpath)
  SELECT ('/nix/store/' || md5(random()::text))
  FROM generate_series(1, 10000000);
```

The size of /var/lib/postgresql/data is 1.8G. The query
```
select * from builds where drvpath LIKE '%ded6%';
```
took about 1.5s to execute.

Then I created the `pg_trgm` index. The creation of the index took
2m30s. The same query took 53ms.
The size of `/var/lib/postgresql/data` is 2.5G after index creation.

For a table containing `100 000 000` rows, results are
- initial `/var/lib/postgresql/data` size: 8.2G
- `/var/lib/postgresql/data` size with index: 15G
- time to create the index: 25m
- query time without index: 56s
- query time with index: 4s (400ms when the query is repeated)


I'm not sure what to conclude... It seems it consumes something close
to the size of the data it has to index (at least in the case I
considered).
How many rows does the `Builds` of  hydra.nixos.org table contain?

edolstra:
It has ~90 million rows. You can use `pg_indexes_size` to get the size of the indexes for a table. For `hydra.nixos.org` this is already 249 GiB, hence my worry about adding another potentially large index:
```
hydra=> select pg_indexes_size('builds');
 pg_indexes_size 
-----------------
    267906842624
(1 row)
```

nlewo:
For a table containing `100` million rows (with one column containing `/nix/store/<UUID>`), the index uses `6.5 GiB`:
```
postgres=# select * from builds limit 2;
                   drvpath                   
---------------------------------------------
 /nix/store/b45c80696fc5833abb53b16d03a50566
 /nix/store/1982122195e776931946fa78865f8805
(2 rows)

postgres=# select count(*) from builds;
   count   
-----------
 100000000
(1 row)

postgres=# select pg_indexes_size('builds');
 pg_indexes_size 
-----------------
      6968713216
(1 row)
```

edolstra:
Thanks, that sounds like a still acceptable increase.

nlewo:
Thx!

grahamc:
@nlewo when searching like this are you searching for partial hashes or full hashes?

PR #657, opened 2019-06-21
Create `pg_trgm` extension in the NixOS module

nlewo:
The creation of the `pg_trgm` extension needs superuser power. So,
this patch makes the extension creation in the Hydra NixOS module when
a local database is used.

If it is not possible to create this extension (remote database for
instance with nosuperuser), the creation of the `pg_trgm` index is
skipped (this index speedup queries on builds.drvpath) and warnings
are emitted:

    initialising the Hydra database schema...
    WARNING:  Can not create extension pg_trgm: permission denied to create extension "pg_trgm"
    WARNING:  HINT: Temporary provide superuser role to your Hydra Postgresql user and run the script src/sql/upgrade-57.sql
    WARNING:  The pg_trgm index on builds.drvpath has been skipped (slower complex queries on builds.drvpath)

This allows to keep smooth migrations: the migration process doesn't
require a manual step (but this manual step is recommended on big
remote databases).

@rbvermaa does it work for you?

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -273,6 +273,7 @@ in
               runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/createdb -O hydra hydra
               touch ${baseDir}/.db-created
             fi
+            echo "create extension if not exists pg_trgm" | runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/psql hydra
           ''}
 
           if [ ! -e ${cfg.gcRootsDir} ]; then
@@ -415,6 +416,8 @@ in
         hydra-users hydra-queue-runner hydra
         hydra-users hydra-www hydra
         hydra-users root hydra
+        # The postgres user is used to create the pg_trgm extension for the hydra database
+        hydra-users postgres postgres
       '';
 
     services.postgresql.authentication = optionalString haveLocalDB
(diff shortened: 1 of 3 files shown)


nlewo:
@rbvermaa @basvandijk does this PR fix your issue (#660) ?

basvandijk:
I haven't tried this on our company hydra but I can confirm this works with `nix-build release.nix -A tests.api.x86_64-linux` after applying https://github.com/NixOS/hydra/pull/658.

hydra-module.nix
```diff
@@ -273,6 +273,10 @@ in
               runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/createdb -O hydra hydra
               touch ${baseDir}/.db-created
             fi
+            # The extension pg_trgm needs to be created by a superuser.
+            echo "alter user hydra with superuser" | runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/psql
+            echo "create extension pg_trgm" | ${config.services.postgresql.package}/bin/psql -U hydra
```
edolstra: Could this be done via the `postgres` superuser? Temporarily making `hydra` a superuser seems a bit ugly.

nlewo: Yep, this is ugly:/
I added a mapping for the `postgres` user in the `pg_ident.conf`. The `postgres` superuser is now able to create the index for the `hydra` database, so we no longer need to provide superuser capabilities to the `hydra` user.

edolstra:
Thanks!

PR #658, opened 2019-06-21
test.api: use Hydra perl dependencies to run `api-test.pl`

nlewo:
The test was failing with:

    machine# Can't locate LWP/UserAgent.pm in @INC (you may need to install the LWP::UserAgent module)...

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -219,7 +219,7 @@ rec {
           $machine->waitForOpenPort("3000");
 
           # Run the API tests.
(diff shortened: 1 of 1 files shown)


basvandijk:
@nlewo thanks for this!
@knl and I just spend almost an hour coming up with the same fix.
Note to self: first look at existing PRs before embarking on a long debugging session.

nlewo:
@basvandijk hehe! you're welcome:)

PR #664, opened 2019-07-25
hydra-server: jobsets api endpoint returns more elements

nlewo:
Improve the jobset API output by returning attributes related to
evaluations. These informations can be useful to know if a jobset has
been successfully evaluated.


nlewo:
`nix-build release.nix  -A tests.api.x86_64-linux` passed.

FYI, these elements are used by a `hydra-cli jobset-wait` command to get the status of a particular jobset. A Travis job can then call this command to check a PR built by an Hydra instance.

edolstra:
Replaced by #672 which omits `errormsg`.

nlewo:
Actually, I need to know if an error occurred or not and I was using the `errormsg` attribute for this purpose. So, I address this need in the PR #673.

PR #668, opened 2019-08-12
Turn hydra-notify into a daemon

edolstra:
`hydra-notify` is now a daemon that receives events from `hydra-queue-runner` by listening to PostgreSQL notifications. This makes it a lot more efficient: not starting a `hydra-notify` process for each event avoids the ~1s startup time loading all the plugins. Also, it allows future plugins to be written in any language as long as they can listen to PostgreSQL notifications, and to run on different machines from the queue runner.

The PostgreSQL notifications are:

* `step_started`
* `step_finished`
* `build_started`
* `build_finished`
* `eval_started`
* `eval_failed`
* `eval_cached`
* `eval_added`

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -391,6 +391,8 @@ in
           { ExecStart = "@${cfg.package}/bin/hydra-notify hydra-notify";
             # FIXME: run this under a less privileged user?
             User = "hydra-queue-runner";
+            Restart = "always";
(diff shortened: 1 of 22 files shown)


AmineChikhaoui:
Awesome !

ajs124:
I take everything back I said. We just weren't running the new daemon, because we still had an older version of the module.

PR #669, opened 2019-08-13
Add GerritChanges input plugin

knedlsepp:
This plugin enables to fetch open changes from a gerrit server.
I've successfully used this to do automatic CI of gerrit topics.

This introduces a dependency on Net::SSH::Perl.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -100,6 +100,7 @@ rec {
             LWPProtocolHttps
             NetAmazonS3
             NetPrometheus
+            NetSSHPerl
             NetStatsd
             PadWalker
             Readonly
--- a/src/lib/Hydra/Plugin/GerritChanges.pm
+++ b/src/lib/Hydra/Plugin/GerritChanges.pm
@@ -1,0 +1,56 @@
+# This plugin allows to build gerrit changes.
+#
+# The declarative project spec.json file must contains an input such as
+#   "pulls": {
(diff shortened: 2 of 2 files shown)


mweinelt:
Hi! I'm wondering whether this is the latest version of this patch, and if you are still using it.

I would hope getting a plugin merged should have a fairly low barrier, since it is not a core part of hydra.

knedlsepp:
> Hi! I'm wondering whether this is the latest version of this patch, and if you are still using it.

Have been using it at #work for years now to generate jobsets for patchsets in Gerrit. I can check tomorrow but I think this is exactly what I'm using.

> I would hope getting a plugin merged should have a fairly low barrier, since it is not a core part of hydra.

I suppose there isn't many people around wanting to use both gerrit and hydra, so I understand that merging this was not on anybody's radar. 🤷
(Frankly I probably wouldn't use Gerrit myself if I had a choice. 😄)

knedlsepp:
I'm sorry @mweinelt I apparently misremembered. I am in fact not actively using this plugin. Instead I ended up (in 2020) downloading, processing and dumping that JSON file in Python (which I'm more familiar with than Perl) as a separate service and just configured my hydra jobsets to read in that local file.
So I think the right action here is to close this PR instead. Thanks however for considering!

PR #673, opened 2019-08-20
Add error boolean attribute to jobset API response

nlewo:
This attribute allows to know if an error occurred or not. Note we can
not use the errormsg attribute because it can be arbitrarily long and
is excluded from the jobset API response.


nlewo:
In this PR, I use `true` and `false` JSON values while `1` and `0` are used in the rest of the API to represent boolean (see `project.hidden` attribute for instance). Maybe we should use `integer` values for homogeneity, even if `boolean` values are better. WDYT?

gilligan:
LGTM.

> Maybe we should use integer values for homogeneity, even if boolean values are better. WDYT?

Personally i'd definitely prefer those values to be *boolean* values and i'd rather like to see the other occurrences of integers that should actually be booleans fixed.

src/lib/Hydra/Controller/API.pm
```diff
@@ -87,7 +87,8 @@ sub jobsetToHash {
         checkinterval => $jobset->checkinterval,
         triggertime => $jobset->triggertime,
         fetcherrormsg => $jobset->fetcherrormsg,
-        errortime => $jobset->errortime
+        errortime => $jobset->errortime,
+        successful => $jobset->errormsg eq "" ? JSON::true : JSON::false
```
grahamc: Does this query always fetch the entire error message every time? Did it before?

Does `errormsg` mean `stderr` output?

If so, the presence  of stderr output doesn't mean it failed which makes "successful" a misnomer.

nlewo: Actually the semantic of `successful` is not correct: when a fetch error occurs, we don't want to have a `successful` attribute set to `true`.

So, we propose `has_errormsg` attribute instead.

Issue #578, opened 2018-07-17
Can't delete some jobsets/projects: DBI exception


ledettwy:
I can't delete some of my jobsets that were previously declarative jobsets. This is the error that Hydra displays:
```
DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::Pg::st execute failed: ERROR: update or delete on table "builds" violates foreign key constraint "releasemembers_build_fkey" on table "releasemembers" DETAIL: Key (id)=(28206) is still referenced from table "releasemembers". [for Statement "DELETE FROM Builds WHERE ( project = ? )" with ParamValues: 1='myproject'] at /nix/store/n9zz43ixj7zvk7c62q3vp8r9m80cwv6x-hydra-2017-11-21/libexec/hydra/lib/Hydra/Controller/Project.pm line 82
```

My only option at this point is to just rename / disable / ignore the jobsets/projects and create new ones.

bjornfor:
I've got the same problem. I never used declarative jobsets though (AFAIR). Of my 5 jobsets, hydra managed to delete 1. The other jobs have this error. (I think this may be the final nail in the coffin for my private hydra.)

PR #677, opened 2019-09-09
Javascript libraries update

twhitehead:
Update to newer versions of the various javascript libraries used.

I undertook this as our security people flagged our SHARCNET/Compute Canada hydra server as a XSS security risk due to the use of jquery 1.x. I'm not certain I agree XSS concerns apply to hydra, but figure the update was worthwhile in its own right as the libraries had fallen multiple major versions behind.

You can check out our [hydra server](https://hydra.sharcnet.ca) to see a running version of these patches on top of the hydra server packaged in 19.03. I rebased that patch set onto master for this pull request. I haven't actually tried it, but the rebase was 100% clean except for a missing `c.uri_for` call that had been added.


edolstra:
A few comments from looking at https://hydra.sharcnet.ca:

* The pagination buttons at the bottom of https://hydra.sharcnet.ca/steps are rendered as lists.

* The header in the "Reproduce this build" popup looks wrong.

twhitehead:
Thanks. Fixed now. I also created you a temporary account on the server so you can fully dig through all the pages, dialogues, and such. Let me know if you find anything else.

twhitehead:
@edolstra expect you have been just busy with things, but it's been 4 months now, so just wondering what the thoughts are on this?

We've been running it on our server and haven't noticed any issues yet. Noticed a minor conflict had arisen against master, so just finished resolving that.

Thanks!

twhitehead:
Actually, I just finished breaking our hydra server by pushing this branch to it to double check my conflict resolution.

Looks like it is just a newer perl dependency (`Net/Prometheus.pm`) that is missing from my override preventing startup.

I probably won't get it resolved to tomorrow though.

twhitehead:
Updated the package dependencies for the newer version and now [our hydra server](https://hydra.sharcnet.ca) is back up for your perusal. Thanks.

twhitehead:
@edolstra any chance this could get merged if I fixed the latest set of conflicts that have arisen in the since I lasted fixed conflicts back in January?

It looks like there was a branch created back in 2013 to attempt to update to boostrap 3, so I was thinking it might have been something you were wanting.

src/root/Makefile.am
```diff
@@ -21,9 +23,14 @@ all:
...
 install-data-local: $(ZIPS)
 	mkdir -p $(hydradir)/static/js
 	cp -prvd $(srcdir)/static/js/* $(hydradir)/static/js
 	mkdir -p $(hydradir)/static/bootstrap
-	cp -prvd $(srcdir)/static/bootstrap/* $(hydradir)/static/bootstrap
+	cp -prvd $(srcdir)/static/$(basename $(BOOTSTRAP))/* $(hydradir)/static/bootstrap
+	mkdir -p $(hydradir)/static/fontawesome/css
+	cp -prvd $(srcdir)/static/$(basename $(FONTAWESOME))/css/all.css $(hydradir)/static/fontawesome/css
+	mkdir -p $(hydradir)/static/fontawesome/webfonts
+	cp -prvd $(srcdir)/static/$(basename $(FONTAWESOME))/webfonts/* $(hydradir)/static/fontawesome/webfonts
```
grahamc, line 36: It looks like this doesn't actually work when in a dev shell:

```
[nix-shell:~/projects/github.com/NixOS/hydra/src/root]$ make install-data-local
mkdir -p /nix/store/sj3rallg7arl4bp009j0ydgdzhkls590-hydra-0.1.20210318.5c8e6d7/libexec/hydra/root/static/js
mkdir: cannot create directory ‘/nix/store/sj3rallg7arl4bp009j0ydgdzhkls590-hydra-0.1.20210318.5c8e6d7’: Read-only file system
make: *** [Makefile:529: install-data-local] Error 1
```

like this make target is only used for install. I think this probably needs to be run as part of the `all` step... not sure, though... experimenting ...

twhitehead: From the output I would have guessed that this must always have failed as it looks to be the first line of the rule
```
	mkdir -p $(hydradir)/static/js
```
which is the same as it was before.

To be honest, I had always assumed none of the install targets were valid inside a dev shell unless you redirect`$out` to somewhere that can be written. Is this not the case?

grahamc:
I gave rebasing this a shot, and published my rebase here: https://github.com/grahamc/hydra/pull/new/twhitehead-javascript-updates. I took a look at this with @samueldr and we found a number of issues. Most of them are minor, but one is quite major.

### Most significantly: clicking tabs don't update the URL.

For example, visit https://hydra.nixos.org/build/137117854 and select Details/Inputs/etc. It'll update the URL, and you can deep-link to a page. Bootstrap 4 seems to eliminate that behavior. Ideally there is a way to bring it back without too much work, but I found https://webdesign.tutsplus.com/tutorials/how-to-add-deep-linking-to-the-bootstrap-4-tabs-component--cms-31180 and it is pretty nasty. I think this problem is significant enough to block this PR.

### Less important things that can be easily fixed

These can probably be each fixed by search and replace.

#### Button styling

Some button widgets are missing the buttoney display treatment.  Example from a build page:

image

#### Error / warning label styling

Some labels like warnings and errors are no longer looking like labels. Example from a project page:

image

#### Project page's default tab state isn't correct

On a project page like `/projects/nixos`, the default loaded tab isn't "selected":

image
 

### Bugs that are already bugs that I don't expect this PR to fix
...but are maybe a bit more pronounced with the new styles.

#### Zebra stripes on the project list and jobset list pages are mis-striped due to hidden jobsets #578 
image

vs:

image

##### More button & Jobs header on a jobset don't extend far enough

image


---

Overall, I think if we can fix the deep linking problem we could merge this. The new style is definitely roomier than the old style, but @samueldr pointed out that it is also much more touch-friendly. It took a few minutes to adjust to the new size, but I don't mind it and actually kind of like it by now.

twhitehead:
The [w3 site](https://w3programmings.com/add-deeplinking-in-the-bootstrap-nav-tabs/) has a bit of javascript to restore deep linking functionality.
```
var url = window.location.href;
if (url.indexOf("#") > 0){
     var activeTab = url.substring(url.indexOf("#") + 1);
     $('.nav[role="tablist"] a[href="#'+activeTab+'"]').tab('show');
}

$('a[role="tab"]').on("click", function() {
     var newUrl;
     const hash = $(this).attr("href");
     newUrl = url.split("#")[0] + hash;
     history.replaceState(null, null, newUrl);
});
```

I'm not terribly familiar with javascript, but my reading would be
1. the first bit handles the initial page load by checks the URL for `#` and then manually activates the associated tab if found, and 
2. the second bit handles the update by updating the URL everytime something in role `tab` is clicked.

samueldr:
This is not "the w3 site"... Sorry for the nit pick. It's just an arbitrary website full of stolen snippets and tutorials.

It comes, word for word AFAICT (the w3programming site breaks heavily with ad blocking) from this bootstrap issue:

 - https://github.com/twbs/bootstrap/issues/25220

This is not a good thing. This is re-implementing what was purposefully broken by bootstrap 4. The links already points to anchors. This may work in most browsers, but I wouldn't put any trust in patching on top of the breakage bootstrap 4 did.

I wouldn't feel confident in approving something that broke a fundamental behaviour of the web, even with a patch on top to try and reinstate it.

Furthermore `replaceState` is the wrong choice, as it replaces the current state instead of pushing on top of the navigation stack, like the native behaviour. (Though I'm observing tabs with bootstrap 2 seems to not push state, so it seems the implementation is the same...)

twhitehead:
@samueldr Thanks for the details and sorry for the bad link. I have to confess I am really out of my depth when it comes to this sort of thing. From your comments and the twbs/bootstrap#25220 issue, it looks like this would be more correct/cleaner
```
const anchor = window.location.hash;
$(`a[href="${anchor}"]`).tab('show')

$('a[role="tab"]').on("click", function() {
     history.pushState(null, null, url.split("#")[0] + $(this).attr("href"));
});
```

With regard to the fundamental behaviour of the the web, I'm not sure I understand what you are saying. I am guessing maybe you are saying the basic idea of using anchors to deep link tabs is breaking a fundamental behaviour of the web, and that is why it was deliberately removed in bootstrap 4. Is this it, and, if so, are you then saying we shouldn't have deep linking period?

samueldr:
Don't worry, the web is way way way more complex than it looks like. So being out of one's depth is not necessarily an issue :).

* * *

> With regard to the fundamental behaviour of the the web, I'm not sure I understand what you are saying. I am guessing maybe you are saying the basic idea of using anchors to deep link tabs is breaking a fundamental behaviour of the web, and that is why it was deliberately removed in bootstrap 4. Is this it, and, if so, are you then saying we shouldn't have deep linking period?

What I meant to say is that clicking on an anchor link should push the navigation to that anchor in the browser history, and scroll to the anchor. I guess scrolling to the anchor is undesirable with tabs navigation.

Though looking at how it works, now, I see that @edolstra basically implemented it. It looks like bootstrap 2 would, too, have broken the navigation the same way.

 - https://github.com/NixOS/hydra/blame/aeb3d2f44ce1841f45343db40c04f94dedec045c/src/root/static/js/common.js#L53-L71

So in the end I guess it's not an issue? uh.

(For the record, I'm not a fan of bootstrap, but I'm still surprised when bootstrap drops the ball...)

grahamc:
This option looks great, and resolves my largest concern:

```js
$(`a[href="${window.location.hash}"]`).tab('show')

$('a[role="tab"]').on("click", function() {
     history.pushState(null, null, url.split("#")[0] + $(this).attr("href"));
});
```

twhitehead:
Thanks guys. Interesting that the original behaviour appears to have been JavaScript too. In some sense then it was a bit I missed in my update. I'll try the rebase plus the javaScript on our server next week to see how it goes.

twhitehead:
Okay, I think I've got the deep link stuff working. It is live on our [hydra server](https://hydra.sharcnet.ca), so give it a try and let me know.

I'll force push an update. I went through the rebase and made some other tweaks on top of it too. So I'll add some comments about that.

If it all looks good, I'll squash the fixups and push it again.

src/root/edit-jobset.tt
```diff
@@ -113,9 +113,7 @@
   <div class="form-group row show-on-flake">
     <label class="col-form-label col-sm-3" for="editjobsetflakeref">Flake URI</label>
     <div class="col-sm-9">
-      <div class="input-group">
-        <input type="text" class="form-control" id="editjobsetflakeref" name="flakeref" [% HTML.attributes(value => jobset.flake) %]/>
-      </div>
+      <input type="text" class="form-control" id="editjobsetflakeref" name="flakeref" [% HTML.attributes(value => jobset.flake) %]/>
```
twhitehead, line 116: Don't think the `input-group` wrapper is required as it is only a single element.

src/root/edit-jobset.tt
```diff
@@ -92,10 +92,10 @@
   </div>
 
   <div class="form-group row">
-    <label class="col-sm-3" for="editjobsettype">Type</label>
+    <label class="col-form-label col-sm-3" for="editjobsettype">Type</label>
     <div class="btn-group btn-group-toggle col-sm-9" data-toggle="buttons">
       <label id="type-flake" class="btn btn-secondary[% IF jobset.type == 1 %] active[% END %]">
-        <input type="radio" name="type" value="1" [% IF jobset.type == 1 %]checked[% END %]>Flake</button>
+        <input type="radio" id="editjobsettype" name="type" value="1" [% IF jobset.type == 1 %]checked[% END %]>Flake</button>
```
twhitehead, line 98: Was missing `id` to tie back to label.

src/root/edit-jobset.tt
```diff
@@ -92,10 +92,10 @@
   </div>
 
   <div class="form-group row">
-    <label class="col-sm-3" for="editjobsettype">Type</label>
+    <label class="col-form-label col-sm-3" for="editjobsettype">Type</label>
```
twhitehead, line 95: Not sure if the `col-form-label` is required, but the others had it it, so added it here too.

src/root/edit-project.tt
```diff
@@ -6,7 +6,7 @@
   <div class="form-group row">
     <label class="col-sm-3" for="editprojectenabled">Enabled</label>
     <div class="col-sm-9">
-      <input type="checkbox" id="editprojectenabled" name="enabled" [% IF project.enabled %] checked="checked" [% END %]/>
+      <input type="checkbox" id="editprojectenabled" name="enabled" [% IF create || project.enabled %] checked="checked" [% END %]/>
```
twhitehead, line 9: The original had `create`, so added it back. Let me know if this was removed on purpose.

src/root/static/js/common.js
```diff
@@ -50,25 +50,13 @@ $(document).ready(function() {
 
     $(".hydra-popover").popover({});
 
-    $(function() {
-        if (window.location.hash) {
-            $(".nav-tabs a[href='" + window.location.hash + "']").tab('show');
-        }
-
-        /* If no tab is active, show the first one. */
```
twhitehead, line 58: Can't see what this was needed for, so just removed it.

src/root/static/js/common.js
```diff
@@ -46,29 +46,17 @@ $(document).ready(function() {
...
-        /* Ensure that pressing the back button on another page
-           navigates back to the previously selected tab on this
-           page. */
-        $('.nav-tabs').bind('show', function(e) {
-            var pattern = /#.+/gi;
-            var id = e.target.toString().match(pattern)[0];
-            history.replaceState(null, "", id);
-        });
+    $('.nav-tabs > .nav-item:not(.dropdown) a[href^="#"]').on('click', function() {
+        history.pushState(null, null, window.location.href.split("#")[0] + $(this).attr("href"));
```
samueldr, line 59: I see why `replaceState` must have been used. Easier to deal with than `pushState` when you have "simple" JavaScript like that.

On a page, going to a tab, then using the back navigation does not go to the tab the URL is now showing (unless you reload). It looks like it's "stealing" a step from going back, since it does nothing.
Since the current Hydra uses `replaceState`, let's use it, otherwise I'm betting we'll have a deluge of "now Hydra is eating back button" issues like I've seen happen on projects switching to the (in my opinion) more correct pushState. (The other options would have been to add the listener to history events.)

```suggestion
        history.replaceState(null, null, window.location.href.split("#")[0] + $(this).attr("href"));
```

src/root/static/js/common.js
```diff
@@ -46,29 +46,17 @@ $(document).ready(function() {
...
-        if (window.location.hash) {
-            $(".nav-tabs a[href='" + window.location.hash + "']").tab('show');
-        }
-
-        /* If no tab is active, show the first one. */
-        $(".nav-tabs").each(function() {
-            if ($("li.active", this).length > 0) return;
-            $("a", $(this).children("li:not(.dropdown)").first()).tab('show');
-        });
+    /* Deep emedding for tabs. */
```
samueldr, line 53: *emedding* isn't a word :).

Another way to identify this is:

```suggestion
    /* Activates tab according to URL anchor. */
```

samueldr:
I have seen a couple locations (live on your hydra instance) where `btn` elements had no "style" added to it. E.g. no `btn-secondary`.

This is because bootstrap 2 had a default style when none was specified. Which changed with more recent bootstrap releases.

 - https://hydra.sharcnet.ca/build/127789#tabs-summary

This list may not be exhaustive, I guess `grep`ping for `btn` and grepping against `btn-primary` and `btn-secondary` would help catch many of the leftovers.

samueldr:
Labels are now badges!

 - https://getbootstrap.com/docs/4.6/components/badge/
 - https://hydra.sharcnet.ca/project/ccpkgs

In the columns, the numbers use `label` classes.

## TODO

 - [ ] `label` → `badge`

samueldr:
Once you have those fixes in for the mis-styled items, I'll checkout and grep *too* to try and see if anything else is missing.

src/root/static/js/common.js
samueldr: This was used to open the first available tab in the `eval` view

 - https://hydra.sharcnet.ca/eval/575
 - https://hydra.nixos.org/eval/1656754

In that first link, the *Still succeeding jobs* tab should get activated.

src/root/static/js/common.js
twhitehead: Thanks a ton for all this feedback! :+1: I'll mark these as resolved as I knock them off.

src/root/Makefile.am
twhitehead: I'm assuming that  was the case, so I'm marking this as resolved. If not, please unmark. Thanks!

PS: Thanks very much for the review effort and especially the rebase work! Much appreciated.

twhitehead:
Okay. I believe I've addressed everything that has been raised and some additional items I found along the way too. I've made the changes as fixups so as to make it easy to see what is new.

The current version is live on [our hydra server](https://hydra.sharcnet.ca). If everything is good, let me know, and I'll squash it all down into the single boostrap update commit.

Thanks a ton to both of you for all the help and follow through on this. I had pretty much given up on it ever going anywhere, and now it is probably almost to the point of merger! :+1:

grahamc:
I tried this again today, and there is still a problem with it being unstyled at development time. After a git clone:

```
./bootstrap
configurePhase
make
foreman start
```

then visit http://127.0.0.1:63333/ and you'll see what I mean.

I get 404's loading:

* `/static/fontawesome/css/all.css`
* `/static/bootstrap/js/bootstrap.min.js`
* `/static/bootstrap/css/bootstrap.min.css`

I used `strace` to figure out what catalyst was trying to do here, and found it is looking for these files in `./src/root/static/`:

```
15:06:51 hydra-server.1 | stat("/home/grahamc/projects/github.com/NixOS/hydra/src/root", {st_mode=S_IFDIR|0755, st_size=51, ...}) = 0
15:06:51 hydra-server.1 | stat("/home/grahamc/projects/github.com/NixOS/hydra/src/root/static/images/debian2.png", 0x13ca4b8) = -1 ENOENT (No such file or directory)
```

so then I realized maybe the bootstrap .zip you're replacing named the directory `bootstrap` and indeed:

```
[nix-shell:~/projects/github.com/NixOS/hydra/src/root/oeu]$ unzip ../bootstrap-2.3.1.zip 
Archive:  ../bootstrap-2.3.1.zip
   creating: bootstrap/
   creating: bootstrap/css/
  inflating: bootstrap/css/bootstrap-responsive.css  
  inflating: bootstrap/css/bootstrap-responsive.min.css  
  inflating: bootstrap/css/bootstrap.css  
  inflating: bootstrap/css/bootstrap.min.css  
   creating: bootstrap/img/
  inflating: bootstrap/img/glyphicons-halflings-white.png  
  inflating: bootstrap/img/glyphicons-halflings.png  
   creating: bootstrap/js/
```

whereas the new one is extracting to a versioned directory. So I applied this patch to your branch:

```patch
--- a/src/root/Makefile.am
+++ b/src/root/Makefile.am
@@ -22,15 +22,16 @@ nobase_hydra_DATA = $(EXTRA_DIST)
 all:
        mkdir -p $(srcdir)/static/js
        unzip -u -d $(srcdir)/static $(BOOTSTRAP)
+       mv $(srcdir)/static/$(basename $(BOOTSTRAP)) $(srcdir)/static/bootstrap
        unzip -u -d $(srcdir)/static/js $(FLOT)
        unzip -u -d $(srcdir)/static $(FONTAWESOME)
+       mv $(srcdir)/static/$(basename $(FONTAWESOME)) $(srcdir)/static/fontawesome
 
 install-data-local: $(ZIPS)
        mkdir -p $(hydradir)/static/js
        cp -prvd $(srcdir)/static/js/* $(hydradir)/static/js
        mkdir -p $(hydradir)/static/bootstrap
-       cp -prvd $(srcdir)/static/$(basename $(BOOTSTRAP))/* $(hydradir)/static/bootstrap
-       mkdir -p $(hydradir)/static/fontawesome/css
-       cp -prvd $(srcdir)/static/$(basename $(FONTAWESOME))/css/all.css $(hydradir)/static/fontawesome/css
-       mkdir -p $(hydradir)/static/fontawesome/webfonts
-       cp -prvd $(srcdir)/static/$(basename $(FONTAWESOME))/webfonts/* $(hydradir)/static/fontawesome/webfonts
+       cp -prvd $(srcdir)/static/bootstrap/* $(hydradir)/static/bootstrap
+       mkdir -p $(hydradir)/static/fontawesome/{css,webfonts}
+       cp -prvd $(srcdir)/static/fontawesome/css/* $(hydradir)/static/fontawesome/css
+       cp -prvd $(srcdir)/static/fontawesome/webfonts/* $(hydradir)/static/fontawesome/webfonts
```

and this fixed the local development flow (after applying, run `configurePhase && make`.)

grahamc:
Note: the above patch is not tested with the `nix-build` and deploy process! I'm going to explore the UI now and see if I can find problems.

grahamc:
* deep linking like http://127.0.0.1:63333/eval/1657641#tabs-errors works
* buttons (log and Details) seem to be correct now
* the project page's default tab is correctly selected
* error/warning labels look good

I'd like @samueldr to take another look at this but I think it is ready once that Makefile patch is applied.

grahamc:
This fixes the table display issues:

```
--- a/src/root/common.tt
+++ b/src/root/common.tt
@@ -439,10 +439,7 @@ BLOCK renderEvals %]
         [% END %]
         <th rowspan="2" style="width: 10em">Date</th>
         <th rowspan="2">Input changes</th>
-        <th colspan="3">Jobs</th>
-        <th rowspan="2">
-            <img src="[% c.uri_for("/static/images/delta.svg") %]" height="12" width="12" title="Delta" alt="Delta" class="build-status" />
-        </th>
+        <th colspan="4">Jobs</th>
       </tr>
       <tr>
         <th style="width: 2em">
@@ -454,6 +451,9 @@ BLOCK renderEvals %]
         <th style="width: 2em">
             <img src="[% c.uri_for("/static/images/emojione-question-2754.svg") %]" height="12" width="12" title="Queued" alt="Queued" class="build-status" />
         </th>
+        <th style="width: 2em">
+            <img src="[% c.uri_for("/static/images/delta.svg") %]" height="12" width="12" title="Delta" alt="Delta" class="build-status" />
+        </th>
       </tr>
     </thead>
     <tbody>
@@ -503,7 +503,7 @@ BLOCK renderEvals %]
         </tr>
       [% END;
       IF linkToAll %]
-        <tr><td class="centered" colspan="5"><a href="[% linkToAll %]"><em>More...</em></a></td></tr>
+        <tr><td class="centered" colspan="7"><a href="[% linkToAll %]"><em>More...</em></a></td></tr>
       [% END %]
     </tbody>
   </table>
```

grahamc:
This patch makes the nav bar appear the same way, sorta centered, as on nixos.org:

```patch
--- a/src/root/layout.tt
+++ b/src/root/layout.tt
@@ -47,18 +47,20 @@
   <body>
 
     <nav class="navbar navbar-expand-md navbar-light bg-light">
-      <a class="navbar-brand" href="[% c.uri_for(c.controller('Root').action_for('index')) %]">
-        [% IF logo == "" %]
-          Hydra
-        [% ELSE %]
-          <img src="[% c.uri_for(logo) %]" alt="Hydra Logo" class="logo" />
-        [% END %]
-      </a>
-      <button type="button" class="navbar-toggler" data-toggle="collapse" data-target="#navbarSupportedContent">
-        <span class="navbar-toggler-icon"></span>
-      </button>
-      <div class="collapse navbar-collapse" id="navbarSupportedContent">
-        [% PROCESS topbar.tt %]
+      <div class="container">
+        <a class="navbar-brand" href="[% c.uri_for(c.controller('Root').action_for('index')) %]">
+          [% IF logo == "" %]
+            Hydra
+          [% ELSE %]
+            <img src="[% c.uri_for(logo) %]" alt="Hydra Logo" class="logo" />
+          [% END %]
+        </a>
+        <button type="button" class="navbar-toggler" data-toggle="collapse" data-target="#navbarSupportedContent">
+          <span class="navbar-toggler-icon"></span>
+        </button>
+        <div class="collapse navbar-collapse" id="navbarSupportedContent">
+          [% PROCESS topbar.tt %]
+        </div>
       </div>
     </nav>
```

samueldr:
### To be fixed still (I'll come up with the patches soon)

 - Alerts don't have a style selected https://hydra.nixos.org/jobset/mobile-nixos/unstable#tabs-channels
 - Search inputs are filling the full width *and* don't have any spacing

* * *

### Couldn't check:

 - Green successful login alert banner

* * *

### Two issues where bootstrap "broke" with bootstrap 4:

 - ["channel" page](https://hydra.sharcnet.ca/jobset/ccpkgs/ccpkgs-17.09/channel/latest) `<pre>` not in a well.
 -  - Logs aren't in a "well"
     - https://hydra.nixos.org/eval/1658884#tabs-errors
     - https://hydra.nixos.org/build/140330080/log

Looks like that they assume you want a totally naked `<pre>`...  Apparently [Cards](https://getbootstrap.com/docs/4.0/components/card/) are what should be used to produce what bootstrap 3 called wells... But that's totally not ergonomic. **And** there is no card that acts like the wells acted with a light background. For their documentation they added their own rules on `highlight` for the background.

I guess since beforehand all `<pre>` had a background implicitly we could just add our own rule on `<pre>`. Though it wouldn't have a red text and background for the evaluation errors. But is that really an issue?

* * *

### Similarly, broke somewhere between bootstrap 2 and 4

The spacing between tabs and the content is screwey *as "designed"* in boostrap 4.

image

AFAICT there is absolutely no built-in component with Bootstrap 4 that is made to fit after a tab navigation properly. Try to find it in **their** docs.

This is most infuriating. What an amateur project where a basic component provided by the framework is incomplete...

We could use something like that though:

```css
.tab-content > .tab-pane {
    padding-top: 1.5rem;
}
```

Which AFAICT 1.5rem is the correct spacing according to bootstrap 4's metrics. Though I might have looked at the wrong thing.

samueldr:
```patch
From fb3398dad339922142b904b543b4ce76591b4ee9 Mon Sep 17 00:00:00 2001
From: Samuel Dionne-Riel <samuel@dionne-riel.com>
Date: Wed, 31 Mar 2021 20:03:44 -0400
Subject: [PATCH] Style untstyled alerts

---
 src/root/jobset-channels-tab.tt | 2 +-
 src/root/jobset-eval.tt         | 2 +-
 src/root/jobset-jobs-tab.tt     | 4 ++--
 src/root/search.tt              | 2 +-
 4 files changed, 5 insertions(+), 5 deletions(-)

diff --git a/src/root/jobset-channels-tab.tt b/src/root/jobset-channels-tab.tt
index c003d70e..692f2682 100644
--- a/src/root/jobset-channels-tab.tt
+++ b/src/root/jobset-channels-tab.tt
@@ -2,7 +2,7 @@
 
 [% IF channels.size == 0 %]
 
-  <div class="alert">There are no channels available.</div>
+  <div class="alert alert-warning">There are no channels available.</div>
 
 [% ELSE %]
 
diff --git a/src/root/jobset-eval.tt b/src/root/jobset-eval.tt
index d444322f..6822ed13 100644
--- a/src/root/jobset-eval.tt
+++ b/src/root/jobset-eval.tt
@@ -29,7 +29,7 @@ project=otherEval.jobset.project.name jobset=otherEval.jobset.name %] evaluation
 c.uri_for(c.controller('JobsetEval').action_for('view'),
 [otherEval.id]) %]">[% otherEval.id %]</a>.</p>
 [% ELSE %]
-<div class="alert">Couldn't find an evaluation to compare to.</div>
+<div class="alert alert-danger">Couldn't find an evaluation to compare to.</div>
 [% END %]
 
 <form>
diff --git a/src/root/jobset-jobs-tab.tt b/src/root/jobset-jobs-tab.tt
index 8acc5ca2..e38234d2 100644
--- a/src/root/jobset-jobs-tab.tt
+++ b/src/root/jobset-jobs-tab.tt
@@ -36,12 +36,12 @@
 
 [% IF jobs.size == 0 %]
 
-  <div class="alert">There are no matching jobs.</div>
+  <div class="alert alert-warning">There are no matching jobs.</div>
 
 [% ELSE %]
 
   [% IF nrJobs > jobs.size %]
-    <div class="alert">Showing the first [% jobs.size %] jobs.  <a href="javascript:setFilter('filter=%')">Show all [% nrJobs %] jobs...</a></div>
+    <div class="alert alert-info">Showing the first [% jobs.size %] jobs.  <a href="javascript:setFilter('filter=%')">Show all [% nrJobs %] jobs...</a></div>
   [% END %]
 
   [% evalIds = evals.keys.nsort.reverse %]
diff --git a/src/root/search.tt b/src/root/search.tt
index d2b2dece..71b88e1c 100644
--- a/src/root/search.tt
+++ b/src/root/search.tt
@@ -81,7 +81,7 @@
 [% IF builds.size > 0 || buildsdrv.size > 0 ; matched = 1 ; END %]
 
 [% IF !matched %]
-  <div class="alert alert-warn">Sorry! Nothing matches your query.</div>
+  <div class="alert alert-warning">Sorry! Nothing matches your query.</div>
 [% END %]
 
 [% END %]
-- 
2.29.2
```

samueldr:
About the search fields filling the whole width... Is it an issue? No it probably isn't. What got to me the most was the lack of spacing.

grahamc:
@twhitehead I've taken the patches in this thread and moved them to a branch like I should have from the start :P https://github.com/NixOS/hydra/compare/master...grahamc:twhitehead-javascript-round-2

twhitehead:
Okay. I've merged in all these commits and restarted our hydra server with them.

It all looks good to me except I am seeing problems with the deep linking and top space to table contents commit that show up whenever you F5 refresh after selecting a tab or follow a direct deep link like [this one](https://hydra.sharcnet.ca/project/ccpkgs#tabs-configuration).

padding

From messing around with the chromium developer console/thing, it appears to be due to the `:target { ...; margin-top: -40px; ... }` line in the hydra.css file introduced in commit 
c2e5bec. Any thoughts on this? Is this Evil CSS hackery still required for something, or can it be entirely stripped out?

samueldr:
You can remove that rule entirely as the navbar isn't sticky anymore.

Good catch!

twhitehead:
With respect to the `<pre>` stuff, I would be tempted to not fight the system and just use the cards and their closest styling.

The [`.bg-light` background option](https://getbootstrap.com/docs/4.0/components/card/#background-and-color) looks pretty close for the light grey background, and we could just switch the evaluation error from the current dark red on light red to the standard white on red danger one.

That is just me though. I'm perfectly happy with whatever you guys want.

twhitehead:
> You can remove that rule entirely as the navbar isn't sticky anymore.

Should I strip out as much of that commit as possible, or just the `:target`?

samueldr:
Only the `:target` rules. The rest of the commit looks like it hitched a ride and solve other issues, mainly.

grahamc:
I'm excited for this PR, I think once this last bit is sorted we can go ahead and merge.

samueldr:
absolutely no one said:

> But why is the navbar not stickied while it was before?

Well, it could be. It's not a good idea. See bugs websites have to work around. Keeping it at the top rather than stickied makes a class of bugs and issues entirely vanish. Thanks for coming to my ted talk.

grahamc:
I didn't even notice the difference. =)

twhitehead:
Okay. I removed the `:target` rule and took a stab at wrapping all the `<pre>` tags in cards to give the same sort of thing we saw on the older server. I actually wound up putting the one-liners into `<code>` tags as they were already separated out by the cards and that gave them the same visual look as  on the original hydra server (wrapping and even margins all around).

I think the only big looking difference is that the error stuff is now white on darker red instead of darker red on light red. You can checkout [our hydra server](https://hydra.sharcnet.ca) as I've updated it to see what you think.

How would you guys want all these commits to merge them? Are you going to do a merge squash so it doesn't matter, or should I

1. squash everything into the bootstrap update commit, or
2. give better names to my fixup commits and leave it as is otherwise.

twhitehead:
By everything, I mean all the tweaks we have made to the bootstrap update commit. I would still leave the other javascript library update commits separate. I would of course list you guys as co-authors.

samueldr:
I wouldn't put the evaluation errors on red. This is because "Evaluation errors" is rather vague, and in Nixpkgs' case, things are fine even with the errors. So styling them the same exact way as other logs is probably better. @grahamc thoughts about the evaluation log?

grahamc:
I agree, the white on red is too intense. It seems like that style is optimized for very very short errors, but the evaluation error log should be optimized for reading.

w.r.t. commits: I'm inclined to do a squash and merge.

twhitehead:
I've gotten rid of the white on red, and expanded out the history on the two fixups commits I did, so that is everything I am aware of/wanted to do. I've updated our [hydra sever](https://hydra.sharcnet.ca) with the latest as before.

Unless anyone else has anything else outstanding on this, I think it is clear to squash merge. Thank both of you very much for all the feedback, commits, and just general work pushing this one though. :+1:

grahamc:
Thank you so much for coming back to this PR and getting it across the line, @twhitehead!

twhitehead:
> Thank you so much for coming back to this PR and getting it across the line, @twhitehead!

Not a problem at all. You guys made it super easy between the rebase and followup patches!  :grin:

PR #685, opened 2019-11-02
Implement maintainer search

Ma27:
As proposed during the release managers' talk on last NixCon, Hydra
should provide a feature to search for maintainers. I wanted to have
something like this for a long time as well and decided to give it a
try.

This change adds a second search field to the jobset-eval view that can
be used to search for maintainer's github names. Please note that I
intentionally didn't merge this together with the `Search by name`-form
since this form is only used to compare the builds against another eval.

By default Hydra now searches for github handles of maintainers (IMHO this
is more convenient than email addresses), however emails are
used fallbacks as the github handle isn't mandatory for a package
maintainer. The new values are stored in the `Builds` table at the new
column `maintainers_github`.

### DIFF
--- a/src/root/jobset-eval.tt
+++ b/src/root/jobset-eval.tt
@@ -37,6 +37,11 @@ c.uri_for(c.controller('JobsetEval').action_for('view'),
   <input name="full" type="hidden" [% HTML.attributes(value => full) %]/>
 </form>
 
+<form class="form-search">
+  <input name="maintainers" type="text" class="input-large search-query" placeholder="Search jobs by maintainers..." />
+  <input type="hidden" name="full" [% HTML.attributes(value => full) %] />
+</form>
+
 <ul class="nav nav-tabs">
   [% IF c.user_exists %]
     <li class="dropdown">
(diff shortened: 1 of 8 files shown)


src/root/jobset-eval.tt
```diff
@@ -37,6 +37,11 @@ c.uri_for(c.controller('JobsetEval').action_for('view'),
   <input name="full" type="hidden" [% HTML.attributes(value => full) %]/>
 </form>
 
+<form class="form-search">
```
Ma27: Those form-fields should be inlined, the current state is just a POC :)

Ma27:
Just converted this to a draft for now for the following reasons:

* I'd like to display maintainer's email and/or github handle (or both if possible). This may cause some changes in the DB schema.
* The two form-fields for searches should be unified (which requires some changes in the filter logic)
* When rebasing this onto the current master, the eval-list rendering doesn't seem to work which must be fixed before this is somehow mergable.

Ma27:
Continued in #743

PR #692, opened 2019-11-14
Add host tag to InfluxDB metrics

knl:
This should help us discern machines in environments with multiple Hydra deployments.

### DIFF
--- a/src/lib/Hydra/Plugin/InfluxDBNotification.pm
+++ b/src/lib/Hydra/Plugin/InfluxDBNotification.pm
@@ -3,9 +3,8 @@ package Hydra::Plugin::InfluxDBNotification;
 use strict;
 use parent 'Hydra::Plugin';
 use HTTP::Request;
-# use JSON;
 use LWP::UserAgent;
-# use Hydra::Helper::CatalystUtils;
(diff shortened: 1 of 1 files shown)


src/lib/Hydra/Plugin/InfluxDBNotification.pm
```diff
@@ -3,9 +3,8 @@ package Hydra::Plugin::InfluxDBNotification;
 use strict;
 use parent 'Hydra::Plugin';
 use HTTP::Request;
-# use JSON;
 use LWP::UserAgent;
-# use Hydra::Helper::CatalystUtils;
+use Sys::Hostname;
```
edolstra: In some other modules we're using `Sys::Hostname::Long`, would that be preferrable?

knl: Either works. `Sys::Hostname` is in the core, but I see that we also bring in `Sys::Hostname::Long`.

knl:
@edolstra @basvandijk please take another look, I changed the code to use `Sys::Hostname::Long`.

Mic92:
ping

PR #698, opened 2019-12-21
Allow error message from remote builder with error type indicator.

kquick:
Original code was unconditionally erasing any error messages received from a remote builder, which made diagnostics more difficult.  This change allows remote builder errors to be displayed in Hydra, including an indicator of the type of error that occurred along with the log message information.

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -340,28 +340,29 @@ void State::buildRemote(ref<Store> destStore,
                 case BuildResult::PermanentFailure:
                     result.stepStatus = bsFailed;
                     result.canCache = true;
-                    result.errorMsg = "";
                     break;
                 case BuildResult::InputRejected:
                 case BuildResult::OutputRejected:
                     result.stepStatus = bsFailed;
                     result.canCache = true;
+                    result.errorMsg += " {inp/out rejected}";
                     break;
                 case BuildResult::TransientFailure:
                     result.stepStatus = bsFailed;
                     result.canRetry = true;
-                    result.errorMsg = "";
                     break;
                 case BuildResult::TimedOut:
                     result.stepStatus = bsTimedOut;
-                    result.errorMsg = "";
+                    result.errorMsg += " {timed out}";
                     break;
                 case BuildResult::MiscFailure:
                     result.stepStatus = bsAborted;
                     result.canRetry = true;
+                    result.errorMsg += " {misc. failure}";
                     break;
                 case BuildResult::LogLimitExceeded:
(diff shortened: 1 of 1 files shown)


src/hydra-queue-runner/build-remote.cc
```diff
@@ -370,12 +371,15 @@ void State::buildRemote(ref<Store> destStore,
...
                     result.stepStatus = bsAborted;
+                    result.errorMsg += " {default/abort}";
                     break;
             }
-            if (result.stepStatus != bsSuccess) return;
-        }
 
-        result.errorMsg = "";
+            if (result.stepStatus != bsSuccess) {
+                result.errorMsg += " Log: " + chomp(readFile(result.logFile));
```
edolstra: We definitely should not store the log file in the database. It's already way to big.

src/hydra-queue-runner/build-remote.cc
```diff
@@ -340,28 +340,29 @@ void State::buildRemote(ref<Store> destStore,
                 case BuildResult::PermanentFailure:
                     result.stepStatus = bsFailed;
                     result.canCache = true;
-                    result.errorMsg = "";
                     break;
                 case BuildResult::InputRejected:
                 case BuildResult::OutputRejected:
                     result.stepStatus = bsFailed;
                     result.canCache = true;
+                    result.errorMsg += " {inp/out rejected}";
```
edolstra: Hm, it's not clear to me what this adds over the `stepStatus`, which is already stored in the database IIRC.

src/hydra-queue-runner/build-remote.cc
kquick: Good point.  Is it worth storing the tail end of the log (10 lines?) for informative purposes, or would you rather there was no log message information stored?

kquick: I investigated this further and have updated this PR.  There are two locations where this diagnostic information could be viewed: (1) on the hydra system directly via `journalctl`, and (2) on the `.../build/NUM#tabs-buildsteps` page.  I had been focused on (1), and hadn't considered (2).  For (2) there is already a mechanism for viewing the logfile, so I've updated to only show the logfile in (1) and not add this information to the database.

src/hydra-queue-runner/build-remote.cc
kquick: Added stepStatus information to logging output (not to database for web ui output) and removed this text where not covered by the stepStatus.

kquick:
@edolstra I've update the PR based on your comments.  Please review this again when you get a chance.  Thanks!

PR #702, opened 2020-01-12
Handle case where jobset has no defined errormsg for api/jobsets

kquick:
Fixes logged messages:

    {date} {hostname} hydra-server[PID]: Use of uninitialized value in string eq at /nix/store/.../libexec/hydra/lib/Hydra/Controller/API.pm line 79 <$read> line 383.

### DIFF
--- a/src/lib/Hydra/Controller/API.pm
+++ b/src/lib/Hydra/Controller/API.pm
@@ -88,7 +88,7 @@ sub jobsetToHash {
         triggertime => $jobset->triggertime,
         fetcherrormsg => $jobset->fetcherrormsg,
         errortime => $jobset->errortime,
-        haserrormsg => defined($jobset->errormsg) ? ($jobset->errormsg eq "" ? JSON::false : JSON::true) : JSON::false
+        haserrormsg => defined($jobset->errormsg) && $jobset->errormsg ne "" ? JSON::true : JSON::false
     };
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/Hydra/Controller/API.pm
```diff
@@ -88,7 +88,7 @@ sub jobsetToHash {
         triggertime => $jobset->triggertime,
         fetcherrormsg => $jobset->fetcherrormsg,
         errortime => $jobset->errortime,
-        haserrormsg => $jobset->errormsg eq "" ? JSON::false : JSON::true
+        haserrormsg => defined($jobset->errormsg) ? ($jobset->errormsg eq "" ? JSON::false : JSON::true) : JSON::false
```
basvandijk: I think it's clearer to collapse the nested condition into a single condition like:
```suggestion
        haserrormsg => defined($jobset->errormsg) && $jobset->errormsg ne "" ? JSON::true : JSON::false
```

nlewo:
@kquick Could you squash these commits?
Otherwise, lgtm.

kquick:
I'd prefer not to squash these commits, since I've already merged them into other branches and the result would be pretty impactful from an administrative perspective (... it would take a significant amount of effort to rebuild those other branches, which had other commits on them as well).  OK, @nlewo?

PR #703, opened 2020-01-12
Update libpqxx usage to move away from deprecated API interactions.

kquick:
The `parameterized()` API function in libpqxx is deprecated.  Update the code to move to the `exec_params` replacement interface.


edolstra:
Thanks!

PR #709, opened 2020-02-06
sql: Generate models from postgresql

grahamc:
Lowercasing is due to postgresql not having case-sensitive table names.
It always technically workde before, but those table names never
existed literally.

The switch to generating from postgresql is to handle an upcoming
addition of an auto-incrementign ID to the Jobset table. Sqlite doesn't
seem to be able to handle the table having an auto incrementing ID
field which isn't the primary key, but we can't change the primary
key trivially.

Since hydra doesn't support sqlite and hasn't for many year anyway,
it is easier to just generate from pgsql directly.

### DIFF
--- a/src/lib/Hydra/Schema/AggregateConstituents.pm
+++ b/src/lib/Hydra/Schema/AggregateConstituents.pm
@@ -27,11 +27,11 @@ use base 'DBIx::Class::Core';
 
 __PACKAGE__->load_components("+Hydra::Component::ToJSON");
 
-=head1 TABLE: C<AggregateConstituents>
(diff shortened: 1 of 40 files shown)


grahamc:
I'm not sure what testing is appropriate for this change, but for what it is worth, these all built fine:

```
nix-build ./release.nix -A build -A manual -A tests.install -A tests.api -A tests.notifications
```

edolstra:
BTW there is already some code for running Postgres in tests/set-up.pl and tests/tear-down.pl, might be nice to unify those.

PR #712, opened 2020-02-06
Add github notifications for evals

pikajude:
Eval notifications are nice. This makes status not get stuck if an evaluation fails.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -61,6 +61,35 @@ rec {
 
       nix = pkgs.nixUnstable or pkgs.nix;
 
+      # Need for proper interpolation
+      StringInterpolate = perlPackages.buildPerlPackage {
+        pname = "String-Interpolate";
+        version = "0.32";
+        src = fetchurl {
+          url = "mirror://cpan/authors/id/N/NE/NEILB/String-Interpolate-0.32.tar.gz";
+          sha256 = "c729d89848f55a357bcf97b79b78b6b92b263e4c3c92747e7ded3639fe5ddc95";
+        };
+        propagatedBuildInputs = with perlPackages; [ PadWalker SafeHole ];
+        meta = {
+          homepage = https://github.com/neilbowers/String-Interpolate;
+          description = "Wrapper for builtin the Perl interpolation engine";
+          license = with stdenv.lib.licenses; [ artistic1 gpl1Plus ];
+        };
+      };
+      SafeHole = perlPackages.buildPerlModule {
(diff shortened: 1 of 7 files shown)


knl:
Maybe it would be good to have unified eval context (instead of one being named after the PR), so one could turn on branch protection on the context being successful.

Implemented in: https://github.com/pikajude/hydra/pull/1

src/script/hydra-notify
```diff
@@ -114,10 +157,34 @@ while (1) {
...
+                    $inputObj->{name} = $k;
+                    push @inputsWithName, $inputObj;
+                }
+                evalFailed($project, $jobset, \@inputsWithName);
             }
         };
         if ($@) {
             print STDERR "error processing message '$payload' on channel '$channelName': $@\n";
         }
+
```
knl, line 185: I don't think this is needed. My understanding is that it's used for the test, and I wrote in the test how one can do it differently, alleviating the need for `once`.

src/script/hydra-notify
```diff
@@ -88,6 +127,10 @@ for my $build ($db->resultset('Builds')->search(
     my $buildId = $build->id;
     print STDERR "sending notifications for build ${\$buildId}...\n";
     buildFinished($build);
+
```
knl, line 130: I don't think this is needed. My understanding is that it's used for the test, and I wrote in the test how one can do it differently, alleviating the need for once.

release.nix
```diff
@@ -293,6 +323,53 @@ rec {
...
+        # running it manually to check stderr output
+        $machine->stopJob("hydra-notify");
+
+        # Setup the project and jobset
+        $machine->succeed(
+          "su - hydra -c 'perl -I ${build.${system}.perlDeps}/lib/perl5/site_perl ${./tests/github/setup.pl}' >&2");
+
+        my $output = $machine->mustSucceed("hydra-notify --once 2>&1");
+        $output =~ m!GithubStatus_Debug POSTing to.*repos/run/jobset/statuses!
+          or die "GithubStatus plugin did not activate";
```
knl, line 369:
```suggestion
        $machine->waitUntilSucceeds("journalctl -u hydra-notify.service | grep -G 'GithubStatus_Debug POSTing to.*repos/run/jobset/statuses'");
```

pingiun:
I have a modified version of this running on my hydra server, and it seems to work great. If you apply the suggestions and rebase this from master I think it would be a good addition

PR #717, opened 2020-02-18
Add the project name to declarative inputs

dasJ:
This allows for more generic declarative configurations which can be
shared between projects.

### DIFF
--- a/src/script/hydra-eval-jobset
+++ b/src/script/hydra-eval-jobset
@@ -565,7 +565,7 @@ sub checkJobsetWrapped {
     if ($jobsetsJobset) {
         my @declInputs = fetchInput($plugins, $db, $project, $jobset, "decl", $project->decltype, $project->declvalue, 0);
         my $declInput = @declInputs[0] or die "cannot find the input containing the declarative project specification\n";
-        die "multiple alternatives for the input containing the declarative project specificaiton are not supported\n"
+        die "multiple alternatives for the input containing the declarative project specification are not supported\n"
             if scalar @declInputs != 1;
         my $declFile = $declInput->{storePath} . "/" . $project->declfile;
         my $declText = read_file($declFile)
@@ -599,6 +599,8 @@ sub checkJobsetWrapped {
                 updateDeclarativeJobset($db, $project, ".jobsets", $declSpec);
                 $jobset->discard_changes;
                 $inputInfo->{"declInput"} = [ $declInput ];
+                $inputInfo ->{"projectName"} = [ { "value" => $project->name, } ];
+
             }
         } else {
             die "Declarative specification file $declFile is not a dictionary"
--- a/doc/manual/src/projects.md
+++ b/doc/manual/src/projects.md
@@ -435,7 +435,8 @@ evaluated will go through the steps above in reverse order:
 2.  Hydra will use the configuration given in the declarative spec file
     as the jobset configuration for this evaluation. In addition to any
     inputs specified in the spec file, hydra will also pass the
-    `declInput` argument corresponding to the input fetched in step 1.
+    `declInput` argument corresponding to the input fetched in step 1 and
+    the `projectName` argument containing the project\'s name.
 
 3.  As normal, hydra will build the jobs specified in the jobset
     repository, which in this case is the single `jobsets` job. When


grahamc:
I don't really understand what this is changing. Can you give a concrete explanation (docs, maybe, even) of what this enables, which couldn't be done before, and how to use it?

dasJ:
@grahamc This makes https://github.com/dasJ/hydra-configs2 work.

Currently, we need one spec JSON for every Hydra project we want to configure from the same repository. This is because the Nix expression that is called in .jobsets cannot differentiate between projects. This is what I did [here](https://github.com/dasJ/hydra-configs/tree/master/specs) which inputs the project name into the expression.

However, this PR does so automatically by adding a second input to the .jobsets evaluation: `projectName`. With this, the .jobsets Nix expression can differentiate between projects without the user having to change the spec JSON, so multiple projects can share one JSON file and the decision on what jobsets are generated and what inputs are used is soley done in nix.
[hydra-configs2](https://github.com/dasJ/hydra-configs2/blob/master/spec.json) allows to do exactly that and can include [whatever additional nix code is needed](https://github.com/dasJ/hydra-configs2/blob/master/default.nix#L10).

grahamc:
Got it. Cool! Can you add some information to that effect around here? https://github.com/NixOS/hydra/blob/master/doc/manual/declarative-projects.xml#L181

ajs124:
@grahamc done

src/script/hydra-eval-jobset
```diff
@@ -587,6 +587,7 @@ sub checkJobsetWrapped {
                 updateDeclarativeJobset($db, $project, ".jobsets", $declSpec);
                 $jobset->discard_changes;
                 $inputInfo->{"declInput"} = [ $declInput ];
+                $inputInfo->{"projectName"} = [ fetchInput($plugins, $db, $project, $jobset, "", "string", $project->name, 0) ];
```
grahamc, line 590: It really threw me for a loop why this code is so complicated, but this is just making:

```
$inputInfo ->{"projectName"} = [ { "value" => $project->name,  } ]
```

which is used later on by inputsToArgs:

```
    my $alt = $inputInfo->{$input}->[0];

        if ($alt->{type} eq "string") {
            push @res, "--argstr", $input, $alt->{value};
        }
```

which is so complicated because of the support for "alternative" inputs, a largely and perhaps totally dead feature.

grahamc:
Can you rebase on master then ping me on IRC when the tests pass?

```patch
From e6ba22b96b9bc7b5646fdf3cd454dd8cb54b460d Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Janne=20He=C3=9F?= <janne@hess.ooo>
Date: Tue, 18 Feb 2020 02:16:18 +0100
Subject: [PATCH] Add the project name to declarative inputs

This allows for more generic declarative configurations which can be
shared between projects.
---
 doc/manual/declarative-projects.xml | 3 ++-
 src/script/hydra-eval-jobset        | 3 ++-
 2 files changed, 4 insertions(+), 2 deletions(-)

diff --git a/doc/manual/declarative-projects.xml b/doc/manual/declarative-projects.xml
index df909a89..8287e011 100644
--- a/doc/manual/declarative-projects.xml
+++ b/doc/manual/declarative-projects.xml
@@ -179,7 +179,8 @@
         file as the jobset configuration for this evaluation. In addition
         to any inputs specified in the spec file, hydra will also pass the
         <literal>declInput</literal> argument corresponding to the input
-        fetched in step 1.
+        fetched in step 1 and the <literal>projectName</literal> argument
+        containing the project's name.
       </para>
     </listitem>
     <listitem>
diff --git a/src/script/hydra-eval-jobset b/src/script/hydra-eval-jobset
index ff76f4ef..3bd53faa 100755
--- a/src/script/hydra-eval-jobset
+++ b/src/script/hydra-eval-jobset
@@ -560,7 +560,7 @@ sub checkJobsetWrapped {
     if ($jobsetsJobset) {
         my @declInputs = fetchInput($plugins, $db, $project, $jobset, "decl", $project->decltype, $project->declvalue, 0);
         my $declInput = @declInputs[0] or die "cannot find the input containing the declarative project specification\n";
-        die "multiple alternatives for the input containing the declarative project specificaiton are not supported\n"
+        die "multiple alternatives for the input containing the declarative project specification are not supported\n"
             if scalar @declInputs != 1;
         my $declFile = $declInput->{storePath} . "/" . $project->declfile;
         my $declText = read_file($declFile)
@@ -587,6 +587,7 @@ sub checkJobsetWrapped {
                 updateDeclarativeJobset($db, $project, ".jobsets", $declSpec);
                 $jobset->discard_changes;
                 $inputInfo->{"declInput"} = [ $declInput ];
+                $inputInfo->{"projectName"} = [ fetchInput($plugins, $db, $project, $jobset, "", "string", $project->name, 0) ];
             }
         } else {
             die "Declarative specification file $declFile is not a dictionary"
-- 
2.28.0
```

grahamc:
Thanks for rebasing!

PR #726, opened 2020-03-26
Document SlackNotifications plugin

knl:
Apart from documenting the plugin and its options, add some print debug statements, to track how well the plugin is doing.


edolstra:
Thanks!

PR #727, opened 2020-03-26
 Abort unsupported build steps

edolstra:
If we don't see machine that supports a build step for `max_unsupported_time` seconds, the step is aborted. The default is 0, which is appropriate for Hydra installations that don't provision missing machines dynamically.

### DIFF
--- a/src/hydra-queue-runner/builder.cc
+++ b/src/hydra-queue-runner/builder.cc
@@ -376,97 +376,8 @@ State::StepResult State::doBuildStep(nix::ref<Store> destStore,
             }
         }
 
-    } else {
-
-        /* Register failure in the database for all Build objects that
-           directly or indirectly depend on this step. */
-
-        std::vector<BuildID> dependentIDs;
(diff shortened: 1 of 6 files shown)


grahamc:
Seems like a good idea. I wonder how we can setup alerting on hydra.n.o to alert us before they get aborted. We might have all the metrics we need already.

edolstra:
I've added a metric to keep track of the current number of unsupported runnable steps.

grahamc:
Looking ...

grahamc:
I wonder what it would take to export this information at the prometheus endpoint as well https://github.com/NixOS/hydra/blob/037f6488f633dbf15ca2d93a0c117f6738c707fe/src/lib/Hydra/Controller/Root.pm#L205 though a method which got all of the gauge information by default would be nicest.

grahamc:
(btw I don't consider this blocking, we can re-export it elsewhere like most of these metrics)

PR #731, opened 2020-04-06
Fix build against nix master

Ma27:
The current Hydra master was broken when building against
`pkgs.nixUnstable` from NixOS 20.03 (5e7ccdc9).

Please note that the nix-shell breaks due to the switch to 20.03 with
the following error:

    bash: src: unbound variable
    build time elapsed:  0m2.172s 0m0.261s 0m0.089s 0m0.026s
    bash: succeedOnFailure: unbound variable

---

`nix-build release.nix -A build` is building.
cc @edolstra @kquick

PR #735, opened 2020-04-09
add github ci tests

Mic92:
To avoid breakages as in: https://github.com/NixOS/hydra/pull/731

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,14 @@
+name: "Test"
+on:
+  pull_request:
+  push:
+jobs:
+  tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,17 @@
+name: "Test"
+on:
+  pull_request:
+  push:
+jobs:
+  tests:
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macos]
```
edolstra, line 9: Can you remove `macos`? It's not a target we support.

Mic92: Fixed!

PR #749, opened 2020-05-03
Don't distribute localhost builds to other builders

lopsided98:
Since the recent Hydra upgrade in nixpkgs, I have noticed that builds that are supposed to be running on localhost are actually getting distributed to other machines configured in `/etc/nix/machines`. If the build fails on the remote machine, it is marked as aborted.

I suspect this is caused by a change in Nix, rather than in Hydra, but I couldn't identify a commit that is to blame. It does not seem to happen for builds that Hydra intentionally runs on remote machines, even if those remote machines have builders configured in `/etc/nix/machines`. This may be due to implementation differences between the `cmdBuildPaths` `nix-store --serve` command (used for localhost builds) and the `cmdBuildDerivation` command (used for remote builds).

Note that I am still running `hydra-migration` because I was unable to get the migration to complete successfully.

This PR passes `--builders ''` to `nix-store` to prevent it from using Nix's remote building system.

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -50,7 +50,7 @@ static void openConnection(Machine::ptr machine, Path tmpDir, int stderrFD, Chil
         Strings argv;
         if (machine->isLocalhost()) {
             pgmName = "nix-store";
-            argv = {"nix-store", "--serve", "--write"};
+            argv = {"nix-store", "--builders", "", "--serve", "--write"};
         }
         else {
             pgmName = "ssh";


basvandijk:
I also noticed our localhost builder building on remote builders. After this commit those remote builds issued by the localhost builder are gone. So far so good...

However, I now see another issue: At the moment our Hydra is building the following `preferLocalBuild` derivation:

```
[root@hydra:/]# nix show-derivation /nix/store/fn433i69vkgx9rilkj5iwc9znsv6ddqy-unit-minica.service.drv | grep preferLocalBuild
      "preferLocalBuild": "1",
```

The problem is it is not just building this derivation it's also building all of its dependencies!

Why are these dependencies not first scheduled to my remote builders before only building this final `preferLocalBuild` derivation on my localhost builder?

AmineChikhaoui:
@basvandijk that was an old bug, I had to revert https://github.com/NixOS/hydra/commit/d4b4255dd2aef51db916dfafd78891178534fd56#diff-09d35108797d5e4cc3e05fb040546314 to have it work correctly.

basvandijk:
My suspicion is that even though I have configured hydra with 
```
services.hydra.useSubstitutes = true;
```
steps build by the localhost builder are not utilizing my configured binary caches and thus have to build all already built dependencies locally. 

I'm not sure yet how to confirm my hypothesis... @edolstra any ideas?

basvandijk:
@AmineChikhaoui thanks! I'm going to see if reverting d4b4255 fixes it...

basvandijk:
I can confirm that reverting d4b4255 fixes it. Now only `preferLocalBuild` derivations are build on my localhost builder.

PR #750, opened 2020-05-04
Add hydra OpenAPI description

gilligan:
**Overview**
This is a WIP attempt to define and publish an [OpenApi 3.0](https://github.com/OAI/OpenAPI-Specification) specification for hydra "REST" interface. The specification is written in yaml as `hydra.yaml` and would be added to the repository along with a browsable/explorable view which can also be used to trigger requests (currently hosted [here](http://www.tpflug.me/api.html)

**Motivation**
- hydra generally lacks documentation so this should be a good start
- there is also tooling to generate server/client code based on the specification. I tried this and i got a working python client that could list projects at first try
- down the road we could also nicely integrate this for testing: generate a client based on the specification and use the client to make requests against the API
- currently there isn't much in the way of versioning hydra so a description like be helpful with that

**TODOs**
- [x] Document all methods for all paths
  - [x] create project
  - [x] create jobset
  - [x] evaluate jobset
  - [x] search
- [x] Add missing descriptions
  - [x] JobsetEvalInput.revision
  - [x] JobsetEvalInput.value
  - [x] JobsetEvalInput.dependency
  - [x] BuildProduct.defaultpath
  - [x] Build.releasename
  - [x] JobsetInput.jobsetinputalts
- [ ] ~Hosting `api.html` somewhere (http://www.tpflug.me/api.html)~

Note that the above probably isn't a comprehensive list of things that would mean that **everything** is covered but if possible i'd like to get a first sensible version merged and then iterate and improve on it.

----
**Help needed**:
I am going to need some help getting this over the finishing line. 

- I already received some help from @grahamc with some descriptions but the items listed above are still clear. The easiest way to help out is to look at http://www.tpflug.me/api.html or `hydra.yaml` directly and help me out with all occurrences of descriptions set to `???` :)
- I generally need people to double check if i have written any kind of nonsense and point out missing pieces in the specification. 
- If you want, you can copy paste hydra.yaml into http://editor.swagger.io/# and play with it


edolstra:
Looks good to me.

*  `JobsetInput.jobsetinputalts` should be removed since we got rid of the concept of input alternatives a long time ago (#279). 
* Same applies to `Build.releasename` (in fact we should get rid of the concept of "releases" in Hydra, which nobody uses).
*  `BuildProduct.defaultpath` is the default subpath to be used for a build product (e.g. `/index.html` if the build product is a directory).

hydra.yaml
```diff
@@ -0,0 +1,576 @@
+openapi: "3.0.0"
```
edolstra, line 1: Maybe this file should be called `hydra-api.yaml` or something? `hydra.yaml` is a bit non-descriptive.

hydra.yaml
```diff
@@ -0,0 +1,576 @@
...
+          description: ???
+          type: string
+        value:
+          nullable: true
+          description: ???
+          type: boolean # ?? seems to be either 'false' or null ?
+        dependency:
+          nullable: true
+          description: ???
+          type: string # ?? i have no idea this is always null
```
edolstra, line 295: I think this is the build ID of another build. Probably another misfeature that we shouldn't document (since using builds as inputs is not very reproducible).

hydra.yaml
```diff
@@ -0,0 +1,576 @@
...
+          description: The type of this input # XXX: should be an enum
+          type: string
+        revision:
+          nullable: true
+          description: ???
+          type: string
+        value:
+          nullable: true
+          description: ???
+          type: boolean # ?? seems to be either 'false' or null ?
```
edolstra, line 291: It's a string whose interpretation depends on the `type` of the corresponding input. E.g.
* For `boolean`: `true`, `false`
* For `nix`: an expression like `[ "i686-linux" "x86_64-linux" ]` (passed using `--arg`)
* For `string`: an arbitrary string (passed using `--argstr`)

hydra.yaml
```diff
@@ -0,0 +1,576 @@
...
+        uri:
+          nullable: true
+          description: URI of this input (which differs depending on the type of the input)
+          type: string
+        'type':
+          description: The type of this input # XXX: should be an enum
+          type: string
+        revision:
+          nullable: true
+          description: ???
```
edolstra, line 286: This is a Git/Mercurial commit hash or a Subversion revision number.

hydra.yaml
```diff
@@ -0,0 +1,576 @@
...
+          type: boolean
+        enabled:
+          description: when set to true the project gets scheduled for evaluation
+          type: boolean
+        jobsets:
+          description: list of jobsets belonging to this project
+          type: array
+          items:
+            type: string
+        releases:
```
edolstra, line 230: I've removed this (#752).

hydra.yaml
gilligan: i prefixed the description with a big **DEPRECATED** and I will remove it as soon as this isn't in the code anymore. 

I don't want to leave it out completely: I don't want `hydra-api.yaml` to have the same fate as `doc/manual/api.xml` meaning i want to verify that the specification is correct in CI. Not as a first step, but eventually. Hence my desire to stick to what is actually there (the same also applies for `jobsetinputalts` which are also still there.

hydra.yaml
gilligan: I dropped all `releases` stuff :+1:

grahamc:
We can skip hosting api.html by pointing to this:

https://petstore.swagger.io/?url=https://raw.githubusercontent.com/gilligan/hydra/swagger-config/hydra-api.yaml

but changing to the master branch of the nixos/hydra repo.

grahamc:
I think it'd be useful to have this hydra yaml file checked in CI for validity, and maybe even `codespell`:

```
$ nix-shell -p codespell --run "codespell hydra-api.yaml
hydra-api.yaml:169: hiddden  ==> hidden
```

(this is a demo error from a while back, it isn't an error now)

gilligan:
@grahamc I don't mind adding codespell but would like to do so in a follow-up PR if that is OK with you. I just tried locally and codespell is acting up for me in funny ways:

``` 
[nix-shell:~/dev/NixOS/hydra]$ codespell hydra-api.yaml
Traceback (most recent call last):
  File "/nix/store/1qmcsm3pqsghccgx7b0s6hlvzppy4v6r-codespell-1.16.0/bin/.codespell-wrapped", line 4, in <module>
    import re
  File "/nix/store/55b9ip7xkpimaccw9pa0vacy5q94f5xa-python3-3.7.6/lib/python3.7/re.py", line 143, in <module>
    class RegexFlag(enum.IntFlag):
AttributeError: module 'enum' has no attribute 'IntFlag'
```

I *did* add validation via `openapi-generator-cli` though. So we **do** validate the yaml file at least. i can open an issue to add `codespell` validation and assign it to myself once this is merged.

PR #752, opened 2020-05-06
Remove the "releases" feature

### LINKED ISSUE #282: Cannot add individual builds to releases
When clicking on a build (e.g. disnix:disnixos-trunk, http://hydra.nixos.org/build/33130082), unfolding the 'Actions' bar and selecting 'Add to release', I get the following error:

500 Internal Server Error 

This build is no longer available.

Could it be the case that builds can't be added, since their results are not stored locally anymore (since the S3 binary cache has been introduced)?

edolstra:
We haven't used this in many years (it was really only used for nix and patchelf releases).

Closes #282.


edolstra:
Note: this doesn't add a SQL migration script to delete the `Releases` and `ReleaseMembers` tables, in case somebody wants to roll back.

PR #753, opened 2020-05-07
Add stale GitHub Action

knl:
This GitHub Action will mark and close all issues and PRs that are older
than 180 days.

The list of open PRs and issues is ever growing and the
old issues will probably not ever be looked again. To make it easier for
the maintainers and also to submitters, mark all stale issues/PRs as
such and delete them after 210 days of inactivity.

### DIFF
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -1,0 +1,28 @@
+name: "Close stale issues and pull requests"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: >
+          This issue has been open 180 days with no activity. Consequently, it
+          is being marked with the "stale" label. This means is that the issue will be
+          automatically closed in 30 days unless more comments are added or the "stale"
+          label is removed. Comments that provide new information on the issue are
+          especially welcome: is it still reproducible? did it appear in other contexts?
+          how critical is it?
+        stale-pr-message: >
+          This PR has been open 180 days with no activity. Consequently, it
+          is being marked with the "stale" label. This means is that the PR will be
+          automatically closed in 30 days unless more comments are added, code is updated,
+          or the "stale" label is removed.
+        days-before-stale: 180
+        days-before-close: 30
+        stale-issue-label: 'stale'
+        stale-pr-label: 'stale'


basvandijk:
@edolstra WDYT? I think it would be good to only have a list of issues and PRs that folks still care about because that makes it easier for us to decide what to work on.

gilligan:
I really think this isn't the right approach. 

There are 245 issues and 45 PRs. What makes you think that issues that are older than 180 days are less relevant than an issue that is 3 days old? It seems to me like hydra doesn't currently have enough active maintainers. 

I think this should be addressed by encouraging people to participate and contribute and by making it easier to do both. Deleting contributions (issues and PRs are both contributions from my point of view) that never got addressed or resolved achieves the opposite.

*In summary:*
Let's try to encourage people and triage issues and find maintainers and contributors instead of just throwing away plenty of potentially important issues.

gilligan:
PS: i'm happy to help out categorizing/labeling/triaging issues - i'll just need the appropriate permissions for that.

edolstra:
@gilligan Well, this wouldn't *delete* contributions, it would only close them. They can always be reopened if there is interest (including by triagers).

gilligan:
@edolstra I understand that they are technically still there but i'm sure we can all agree that closed issues are (by definition) less visible.

Now, if there are issues for which active maintainers have had no capacity, then *closing* those issues is certainly not going to improve the chances of those tickets being addressed?

Sorry if i'm not doing a great job bringing my points across. I am just really perplexed by this approach.

> The list of open PRs and issues is ever growing and the old issues will probably not ever be looked again

The open PRs and issues are **not the problem**. The problem is the lack of active maintainers taking care of them. By deploying any kind of automatic closing of issues and PRs we aren't **solving** anything. We just give in to the neglect.

Let's please try to address the actual problem(s) instead.

grahamc:
How about we give @gilligan "triage" permissions?

grahamc:
I went ahead and gave @gilligan triage permissions to help with this. I have to say I've only ever been annoyed and/or upset by a bot auto-closing my issue just because the project maintainers never did anything about it.

knl:
> @edolstra I understand that they are technically still there but i'm sure we can all agree that closed issues are (by definition) less visible.

This proposal doesn't close issues/PRs immediately. It comments on each issue/PR with a statement that it will be closed in 30 days. This gives the chance (or even a reminder) to the contributor to revive the discussion.

andir:
In my experience those just do not help at all. Just because there isn't anything being done doesn't mean nothing could be done. Those issues might be more important then others but nobody ever looked at them. If someone looked at them and deemed them not a bug / not important / user issue / … that should be reflected by labels and/or the issue being closed by the human making the decision.

The only setting where I can see this working is where you have full time staff (or a mob of people..) that continuously monitors every single issue and is able to take action. In those cases, where you have a SLA of sorts (since you hired someone to do that repository sitting) closing issues where the initial author wasn't able to provide further input (after being asked for that) *MIGHT* make sense.

I've had a few long standing issues (in other projects) that had been closed after n days of inactivity. While the issue was still there it just let to me abandoning that project and using something else. It really discourages to contributing issues at all. IIRC as a normal user you aren't even able to re-open an issue if you were late on "bumping" it...

PR #574, opened 2018-07-15
Improve dev env

Ma27:
This PR contains two commits with the attempt to make bootstrapping and hacking on Hydra easier. Until now it missed to mention hydra-init and setting the env vars accordingly.

Please note that this PR is highly opinionated and it probably shouldn't merged as-is to master. However it may help other folks to get started with contributing to Hydra even faster.

-----
**Update** from *2019-11-02* (after adding postgresql support):

This is basically stuff I found out during several times I contributed
minor things to Hydra. This patch adds several files to `.gitignore` to
ensure that one doesn't add random stuff accidentally to their commit.

Also improved the shellHook to set all required environment variables
and to check if there's a postgresql available (optionally one can start
a simple instance in a nixos-container using `test-database.nix`).

PR #759, opened 2020-05-14
Devshell

gilligan:
### overview

This PR adds a [foreman](https://github.com/ddollar/foreman) based setup to easily start hydra from the working copy. 

### motivation

The goal of this PR is to make it dead simple to run Hydra from the working copy. This makes for a much better development and testing experience. It doesn't require any additional permissions, virtualization or networking.

- `foreman start`: this can be executed in the nix-shell when you don't want to make us of incremental builds via `make`. It uses all hydra binaries from the working copy.

### details

- hydra is stared on port 63333 to avoid collisions
- postgres is started on port 64444 to avoid collisions
- the nix-shell environment now has a `foreman` which can be used to start all required hydra services and postgresql
- foreman is configured in `Procfile`


edolstra:
Not in favor of this. It brings in another (non-systemd) service management framework which seems brittle and painful to maintain.

grahamc:
I dunno, is there an easy way to run the various hydra components + postgres under systemd for a dev-only environment, without reconfiguring the whole system? I don't know of one. I'm pretty excited about this PR -- being able to enter the shell and run `foreman` is a really nice way to develop the software.

gilligan:
@edolstra I'll agree that it isn't perfect that this adds a dependency on an (already packaged) Ruby program. However:

- the "framework" is really primitive/small surface (see Procfile)
- i would argue it provides a _drastic_ improvement of developer experience
- there have been other attempts at providing something similar which were either less convenient/and or much more involved or required additional setup/permissions: https://github.com/NixOS/hydra/pull/574 and https://github.com/NixOS/hydra/pull/353. I would consider this PR a better approach.

I generally go by "_perfection is the enemy of good_". This isn't perfect but I think it is as good as it gets in terms of providing a really easy and convenient way of working with hydra. This also seems like a very relevant factor in attracting contributors and encouragement to work on Hydra.

I am more than happy to demo it you if you want. it's quite nice, i promise ;)

gilligan:
FWIW: I am happy to remove `runHydra` and limit this to calling `foreman` in the normal shell environment. What i want to emphasize here is:

This isn't about introducing some novel way of running Hydra. This is strictly about improving the development workflow experience by providing an **easy** way to run hydra from your workspace without requiring **any** additional setup.

I explored different ways (as did others, see the PRs i mentioned above) but this seems like the best solution to me.

scripts/start-hydra.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# wait for postgresql to be up
+while ! pg_ctl -D $(pwd)/.hydra-data/postgres status; do sleep 1; done
```
grahamc, line 4:
```suggestion
# wait for postgresql to listen
while ! pg_isready -h localhost -p 64444; do sleep 1; done
```

scripts/start-hydra.sh
```diff
@@ -0,0 +1,12 @@
...
+
+createdb -h $(pwd)/.hydra-data/postgres -p 64444 hydra
+
+hydra-init
+hydra-create-user alice --password foobar --role admin
+
+touch .hydra-data/hydra.conf
+HYDRA_CONFIG=$(pwd)/.hydra-data/hydra.conf exec hydra-server --port 63333
```
grahamc, line 12:
```suggestion
HYDRA_CONFIG=$(pwd)/.hydra-data/hydra.conf exec hydra-server --port 63333
```

scripts/start-queue-runner.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# wait until hydra is listening on port 63333
+while ! nc -z localhost 63333; do sleep 1; done
+
+touch .hydra-data/hydra.conf
+HYDRA_CONFIG=$(pwd)/.hydra-data/hydra.conf exec hydra-queue-runner
```
grahamc, line 7:
```suggestion
if [ ! -f .hydra-data/machines ]; then
    echo 'localhost x86_64-linux - 1 1 big-parallel' > .hydra-data/machines
fi

export NIX_REMOTE_SYSTEMS=$(pwd)/.hydra-data/machines
export HYDRA_CONFIG=$(pwd)/.hydra-data/hydra.conf
exec hydra-queue-runner
```

grahamc:
I've checked this PR out and it looks pretty good. I find Hydra pretty annoying to run, and "okay time to look at a bug in Hydra" always starts with a sigh and "okay ... where did I put my notes about how to dev Hydra?" 

If we create a `default.nix` in `tests/`, then this jobset configuration works and creates a build:

image

samueldr:
I'm curious if there was inspiration from https://github.com/samueldr/hydra-in-a-bag or if it was only concurrent evolution of the same idea. In any case, it's great to see this in the upstream project, rather than a side-project.

gilligan:
@samueldr hah, cool. Now that you mention it, i do remember this project! I was chatting with @andir and we explored some options that didn't lead anywhere. However graham had mentioned in passing that he once tried to use `foreman` for this. That's how we ended up with that.

I *do* think i once tried your `hydra-in-a-bag` solution but i had totally forgotten about it  really ;) Thanks for reminding me!

gilligan:
FWIW I wouldn't have a problem nixifying our solution in the same way that @samueldr chose. I think we have some improvements in the way how we wait for the database and that we don't have a separate DB initialization step. But all that could be adjusted.

could be done. could also be done in a follow-up PR. But in any case only once if we can reach an agreement that this is desirable.

samueldr:
Now that I have had some time to read through the discussion more in details, here's my thoughts.

This is **not** a way to run a production or production-like Hydra instance. This is a way to *compartmentalize* a development instance of Hydra in a trivial manner, where its state is all self-contained.

(In this approach the state is hidden in an hidden directory, which I think is probably a minor issue. Showing the state directory front and center would probably be better in exposing it to the user.)

Now, this kind of self-contained and self-starting development environment is the basic courtesy we should give to anyone wanting to contribute to the Hydra project. Anything more complex to setup, requiring either a different host (server, vm) or a reconfiguration of the host system is a non-starter. It is too much friction to *just fix a thing*. Even more, this is a good way to get a test instance running to quickly *just try something* in Hydra.

Assuming that this ends up being undesirable inside the Hydra project, I guess an *-in-a-bag* kind of setup at nix-community could do, but this reduces its discoverability.

shell.nix
```diff
@@ -3,4 +3,4 @@
 
 (import (fetchTarball https://github.com/edolstra/flake-compat/archive/master.tar.gz) {
   src = builtins.fetchGit ./.;
-}).shellNix
+}).defaultNix.devShell
```
edolstra, line 6: This seems strange. How can `defaultNix` have a `devShell` attribute? Also `shellNix` is defined as `devShell.${system}` in flake-compat.

flake.nix
```diff
@@ -290,6 +293,17 @@
         nixpkgs.overlays = [ self.overlay nix.overlay ];
       };
 
+      runHydra = pkgs.callPackage ./run-hydra.nix {};
+      devShell = pkgs.hydra.overrideAttrs (old: {
```
edolstra, line 297: I think this is not a correct `devShell` and `nix flake check` will probably complain, since `devShell` needs to be a set of dev shells per platform (i.e. it should be `devShell.x86_64-linux`).

edolstra: And `devShell` shouldn't be necessary since `nix dev-shell` uses `defaultPackage` by default.

Procfile
```diff
@@ -0,0 +1,4 @@
+hydra-server: ./scripts/start-hydra.sh
+hydra-queue-runner: ./scripts/start-queue-runner.sh
+hydra-evaluator: ./scripts/start-evaluator.sh
+postgres: ./scripts/start-postgres.sh
```
edolstra, line 4: Could these be moved to another directory (e.g. `./foreman`)? Currently the naming suggests that these are *the* scripts for running Hydra. And it creates some ambiguity about where to place scripts given that `src/script` already exists.

edolstra:
> I am happy to remove runHydra and limit this to calling foreman in the normal shell environment.

Yes please. For the use case of incremental development `runHydra` isn't necessary (or even a good idea) and it might tempt people to start using it in production.

Mic92:
> Not in favor of this. It brings in another (non-systemd) service management framework which seems brittle and painful to maintain.

In theory it would be possible to use `systemd --user` or `systemd-run` however they don't work on other operating system, which makes porting painful.

shell.nix
gilligan: i frankly don't know how the flake setup really works. The way this came to be is that we added an additional shell instead of using the default one.

Looking at the current status I think I could just add `foreman` as buildInput to the default shell and then i can just drop this and use `shellNix` again.

flake.nix
gilligan: As noted above i'll go ahead and drop the `devShell` and adjust the default shell expression accordingly.

Procfile
gilligan: Okay no problem.

flake.nix
```diff
@@ -104,13 +104,16 @@
...
               perlDeps perl final.nix
-              postgresql95 # for running the tests
               boost
               (if lib.versionAtLeast lib.version "20.03pre"
                then nlohmann_json
                else nlohmann_json.override { multipleHeaders = true; })
             ];
 
+          checkInputs = [
+            postgresql95
```
edolstra, line 114: `postgres95` should probably remain a build input since hydra-queue-runner links against it. The comment about it being for tests is no longer up to date.

Procfile
gilligan: done.

flake.nix
gilligan: fixed.

gilligan:
@edolstra alright..
- runHydra is gone
- devShell is gone
- shell.nix refers to the `shellNix` attribute from `flake-compat`
- the scripts are now in `./foreman`
- i updated the README

I could imagine to encapsulate everything with Nix in the same way that @samueldr does it with hydra-in-a-bag but I would prefer to try that separately and if it turns out well offer a follow-up PR for you/everyone to look at.

edolstra:
Thanks @gilligan!

knl:
> > Not in favor of this. It brings in another (non-systemd) service management framework which seems brittle and painful to maintain.
> 
> In theory it would be possible to use `systemd --user` or `systemd-run` however they don't work on other operating system, which makes porting painful.

Maybe I missed that announcement, but does Hydra work on anything that is not Linux + systemd? If yes, could you please add some docs on how to do it?

edolstra:
It might work but it's not supported.

PR #760, opened 2020-05-14
Correct the link to hydra-api.yml file

knl:
The correct filename is `hydra-api.yml` not `hydra.yml`.


gilligan:
ooops, thanks for catching that @knl !

PR #767, opened 2020-05-25
Refactor/Split up flake.nix

gilligan:
**Motivation**
The goal is to make the `flake.nix` easier to grasp and parse. The returned attribute set should be small and easy to recognize with few big bindings.

**Details**
This is first and foremost about splitting up `flake.nix` into:
- nix/hydra-jobs.nix
- nix/hydra.nix

I also added a couple of comments and `build.x86_64-linux` in `hydra-jobs` now refers to `pkgs.hydra` instead of `packages.x86_64-linux` which (to my understanding) is identical.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,7 +16,7 @@
       # - `hydra` via overalay
       pkgs = import nixpkgs {
         system = "x86_64-linux";
-        overlays = [ self.hydraOverlay nix.overlay ];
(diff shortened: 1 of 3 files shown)


edolstra:
Not in favor. I intentionally combined everything (e.g. release.nix) into flake.nix so we don't have to deal with a bunch of different files.

PR #769, opened 2020-05-26
Make --no-allow-import-from-derivation configurable in hydra-eval-jobset

knl:
When deploying Hydra different than hydra.nixos.org one may encounter a problem
as building any job that uses IFD fails with:

```
May 22 19:41:07 hydra hydra-evaluator[6960]: error: "attempted to realize '/nix/store/1jm02mfiv58rpy8zrx95cpqxzsp64ssh-source.drv' during evaluation but 'allow-import-from-derivation' is false"
May 22 19:41:07 hydra hydra-evaluator[6960]: error: "attempted to realize '/nix/store/av3jr8ix4qcadq2wm3y3hplvxwzlhl4y-source.drv' during evaluation but 'allow-import-from-derivation' is false"
May 22 19:41:07 hydra hydra-evaluator[6960]: error: "attempted to realize
'/nix/store/2jm02mfiv58rpy8zrx95cpqxzsp64ssh-source.drv' during evaluation but
'allow-import-from-derivation' is false"
```

The recent change enforced passing `--no-allow-import-from-derivation`
to `hydra-eval-job` unconditionally. This change makes it configurable and
defaults to **NOT PASSING IT** -- most of the deployments allow IFDs.

The configuration option is called `allow_import_from_derivation` and
defaults to `true`. It is interpreted as a boolean, with only true option being
`true`.

### DIFF
--- a/src/script/hydra-eval-jobset
+++ b/src/script/hydra-eval-jobset
@@ -348,7 +348,7 @@ sub evalJobs {
                 inputsToArgs($inputInfo));
     }
 
-    push @cmd, "--no-allow-import-from-derivation";
+    push @cmd, "--no-allow-import-from-derivation" if $config->{allow_import_from_derivation} // "true" ne "true";
 
     if (defined $ENV{'HYDRA_DEBUG'}) {
         sub escape {


src/script/hydra-eval-jobset
```diff
@@ -348,7 +348,7 @@ sub evalJobs {
                 inputsToArgs($inputInfo));
     }
 
-    push @cmd, "--no-allow-import-from-derivation";
+    push @cmd, "--no-allow-import-from-derivation" if $config->{disallow_import_from_derivation} // "false" ne "true";
```
edolstra, line 351: Maybe invert this to `allow_import_from_derivation` to make it the same as the nix.conf option?

knl: Good point @edolstra. I updated the PR, please take a look.

PR #772, opened 2020-06-02
Support multiple jobsets per Flake

dasJ:
This allows usage of `#`, just as `nix run` does. This way, multiple
jobsets can be configured per flake.nix. For the path given after `#`,
an attribute is looked up and either `hydraJobs` or `checks` is searched
there.

For example `nixpkgs#example` will try to evaluate `example.hydraJobs`
in nixpkgs.

### DIFF
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -105,7 +105,17 @@ static void worker(
     if (myArgs.flake) {
         using namespace flake;
 
-        auto flakeRef = parseFlakeRef(myArgs.releaseExpr);
+        auto attrPath = std::string("");
+        auto releaseExpr = myArgs.releaseExpr;
+
+        // Handle attribute selectors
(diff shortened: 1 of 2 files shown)


edolstra:
The problem is that `nix flake check` doesn't know about these attributes. Maybe we can have an attribute `hydraJobsets`, and require that you can only specify `#hydraJobsets.<jobsetname>` or `#checks`.

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -105,7 +105,17 @@ static void worker(
     if (myArgs.flake) {
         using namespace flake;
 
-        auto flakeRef = parseFlakeRef(myArgs.releaseExpr);
+        auto attrPath = std::string("");
+        auto releaseExpr = myArgs.releaseExpr;
+
+        // Handle attribute selectors
+        auto hash = releaseExpr.find('#');
```
edolstra, line 112: This can be done using `parseFlakeRefWithFragment`.

PR #774, opened 2020-06-03
Add notification of steps that cannot be run on any current worker.

kquick:
Currently there is no notification of steps that cannot be run because there is no compatible worker configured for those steps, which makes this hard to diagnose.  This PR adds a log entry in the `hydra-queue-runner` job to provide that information.

### DIFF
--- a/src/hydra-queue-runner/dispatcher.cc
+++ b/src/hydra-queue-runner/dispatcher.cc
@@ -282,6 +282,19 @@ system_time State::doDispatch()
             if (keepGoing) break;
         }
 
+        for (auto & stepInfo : runnableSorted) {
+            auto & step(stepInfo.step);
+            bool couldRunStep = false;
+            for (auto & mi : machinesSorted)
+                if (mi.machine->supportsStep(step)) {
+                    couldRunStep = true;
+                    break;
+                }
+            if (couldRunStep) continue;
+            printMsg(lvlError, format("No machine available to run step '%1%' (needs system type '%2%')") %
+                     localStore->printStorePath(step->drvPath) % step->systemType);
+        }
+
         /* Update the stats for the auto-scaler. */
         {
             auto machineTypes_(machineTypes.lock());


src/hydra-queue-runner/dispatcher.cc
```diff
@@ -282,6 +282,19 @@ system_time State::doDispatch()
...
 
+        for (auto & stepInfo : runnableSorted) {
+            auto & step(stepInfo.step);
+            bool couldRunStep = false;
+            for (auto & mi : machinesSorted)
+                if (mi.machine->supportsStep(step)) {
+                    couldRunStep = true;
+                    break;
+                }
+            if (couldRunStep) continue;
```
knl, line 293: Wouldn’t it make it more readable by putting the print inside `if` (Line 290). That way, you can get rid of the variable all together (because that `if` does a break).

kquick: In this case I am printing on the empty set, not on membership, so the variable is needed to confirm that condition before printing.  The first member only confirms it is non-empty and shortcuts additional membership tests.

edolstra:
We already have a mechanism for this (4417f9f260b17cf2d4098df2a6a4b858856b76ea). I think it would be suboptimal to have two mechanism for detecting unsupported build steps.

edolstra:
Also, won't this end up spamming the log? Since it will print a warning for every unsupported step every time `doDispatch()` runs.

kquick:
The original PR was more for defensive analysis and predated 4417f9f; the change in the latter is preferable for overall handling of the situation.

Regarding log spamming: perhaps, although in practice, not entirely due to the periodicity of the `doDispatch()` invocations.  Also worth noting: the log spamming would make visible the internal churning over these steps, but again, the other patch which cancels them with the "aborting unsupported build step" message is a better longer term approach, so I'm happy to withdraw this PR.

Issue #706, opened 2020-01-24
No step created for build


basvandijk:
Yesterday many builds on our Hydra got queued but didn't get build. The builds from later evaluations of the same jobsets did get build successfully.

I'm now trying to understand what went wrong in Hydra to cause this to happen. I haven't restarted the `hydra-queue-runner` yet to preserve as much state as possible.

Let's look at one of the builds that are still queued: build 558937. Note this is a top-level build; no other builds depend on this one.

Let's first look at the log:

```
...
Jan 21 22:39:34 hydra hydra-evaluator[9874]: added build 558937 (...)
...
Jan 21 22:39:35 hydra hydra-queue-runner[9894]: got notification: new builds added to the queue
Jan 21 22:39:35 hydra hydra-queue-runner[9894]: checking the queue for builds > 558932...
...
Jan 21 22:39:35 hydra hydra-queue-runner[9894]: loading build 558933 (...)
...
Jan 21 22:39:35 hydra hydra-queue-runner[9894]: loading build 558934 (...)
...
Jan 21 22:39:35 hydra hydra-queue-runner[9894]: loading build 559010 (...)
...
Jan 21 22:39:37 hydra hydra-queue-runner[9894]: got notification: new builds added to the queue
Jan 21 22:39:37 hydra hydra-queue-runner[9894]: checking the queue for builds > 559010...
...
```

So the evaluator correctly adds build 558937. Indeed it's also present in the DB:

```
hydra=> SELECT * FROM builds WHERE id = 558937;
-[ RECORD 1 ]------------+---------------------------------------------------------
id                       | 558937
finished                 | 0
timestamp                | 1579646374
project                  | ...
jobset                   | ...
job                      | ...
nixname                  | ...
description              |
drvpath                  | /nix/store/gvbqrw9jygizc7ixpc02fk8gc0zi30qj-....drv
system                   | x86_64-linux
license                  |
homepage                 |
maintainers              |
maxsilent                | 7200
timeout                  | 36000
ischannel                | 0
iscurrent                | 0
nixexprinput             | src
nixexprpath              | ci/ci.nix
priority                 | 100
globalpriority           | 0
starttime                |
stoptime                 |
iscachedbuild            |
buildstatus              |
size                     |
closuresize              |
releasename              |
keep                     | 0
notificationpendingsince |
```

[...]

PR #776, opened 2020-06-05
Fix issue #706: do not use locally cached last build ID.

kquick:
Relies on database optimizations instead of local lookup
optimizations, but resolves the issue where the local ID was being
updated prematurely and leaving builds permanently (until the next
hydra restart) in the unstarted condition.

Previous to this change, builds were frequently left in a Pending status and required manual starting or a restart of the hydra-queue-runner.  After implementing this change, the local hydra instance has run for several months without any builds left Pending.

### DIFF
--- a/src/hydra-queue-runner/queue-monitor.cc
+++ b/src/hydra-queue-runner/queue-monitor.cc
@@ -33,12 +33,10 @@ void State::queueMonitorLoop()
 
     auto destStore = getDestStore();
 
-    unsigned int lastBuildId = 0;
-
     while (true) {
         localStore->clearPathInfoCache();
 
-        bool done = getQueuedBuilds(*conn, destStore, lastBuildId);
+        bool done = getQueuedBuilds(*conn, destStore);
 
         /* Sleep until we get notification from the database about an
            event. */
@@ -49,12 +47,10 @@ void State::queueMonitorLoop()
             conn->get_notifs();
 
         if (auto lowestId = buildsAdded.get()) {
-            lastBuildId = std::min(lastBuildId, static_cast<unsigned>(std::stoul(*lowestId) - 1));
             printMsg(lvlTalkative, "got notification: new builds added to the queue");
         }
         if (buildsRestarted.get()) {
             printMsg(lvlTalkative, "got notification: builds restarted");
-            lastBuildId = 0; // check all builds
         }
         if (buildsCancelled.get() || buildsDeleted.get() || buildsBumped.get()) {
             printMsg(lvlTalkative, "got notification: builds cancelled or bumped");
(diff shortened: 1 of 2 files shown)


src/hydra-queue-runner/queue-monitor.cc
```diff
@@ -74,32 +70,27 @@ struct PreviousFailure : public std::exception {
...
-    unsigned int newLastBuildId = lastBuildId;
-
     {
         pqxx::work txn(conn);
 
         auto res = txn.exec_params
             ("select id, project, jobset, job, drvPath, maxsilent, timeout, timestamp, globalPriority, priority from Builds "
-             "where id > $1 and finished = 0 order by globalPriority desc, id",
-            lastBuildId);
+             "where finished = 0 order by globalPriority desc, id");
```
edolstra, line 88: Doesn't this cause it to select the same builds over and over (at least until they're finished)?

kquick: Thanks for reviewing this, @edolstra.

As I understood the code, this particular section is building the `newIDs` list that is then processed (by the last `for` loop in this same function) to call `createBuild` which will then try `createStep` for the build's `drvPath` but the latter will find the step already existing in the `buildsteps` table and simply return that without adding to `newSteps`.  The result of that will be a call to `propagatePriorities` and a message, which I observe periodically in the logs:

    Jun 05 00:01:40 ___ hydra-queue-runner[14191]: added build 3836338 (top-level step /nix/store/___.drv, 0 new steps)

I suppose the `propagatePriorities` call and the log message could be skipped if `newSteps.empty()`.  I can make this change if you agree with this tactic.

Ultimately yes, this causes the queue runner to do some extra work re-checking builds that are already queued until they finish, but I think that cost is relatively low compared to leaving stranded pending builds that never get processed.  More efficiency could probably be gained by adding another state to to the Builds table (perhaps to `buildStatus`), but that would be a considerably more impactful change which I tried to avoid in this PR.

edolstra: On hydra.nixos.org, we can have ~100000 unfinished builds in the database, so that means we would be constantly fetching the same 100000 builds from the database (and processing them). That's a lot of database traffic...

edolstra:
The real bug here may be that
```
if (id > newLastBuildId) newLastBuildId = id;
```
interacts poorly with `order by globalPriority desc, id`. I.e. the original assumption was that builds get processed in strictly increasing ID order, but then the priority was added which messes that up.

kquick:
I think there would be a bad interaction between `globalPriority`-based sorting and a local `lastBuildId`, but I don't think that's the culprit for my situation: there are only a handful of records in my 4 million row Builds table that don't have a globalPriority of 0.

I added an update patch (6c10510) that sets the `startTime` of the Build record when it is initially processed. These are all initially created as zero, and there is no other query that references `startTime` that does not also use `stopTime` (and usually `finished`), so this should be a viable indicator that the queue-monitor has processed the Build record that doesn't create problems elsewhere.  This approach should cause each Build record to be processed only once, and it should also work just fine relative to the `globalPriority` ordering.

Please let me know what you think of this approach, @edolstra.  Thanks!

kquick:
@edolstra this latest change is running locally with good performance and not "lost" builds, so I think its ready for you to take another look and merge if it looks reasonable to you.  Thanks!

PR #789, opened 2020-06-28
reproduce: Only download the needed commit

c00w:
On something like github this is super efficient due to
https://github.com/git/git/commit/68ee628932c2196742b77d2961c5e16360734a62

But in general it should be more efficient since the entire nixpkg
history doesn't need to be downloaded.

I've manually tested this by fetching a reproduce script from hydra + modifying it to have these changes.


c00w:
@Mic92 Any idea who I should reach out to for hydra changes?

Mic92:
It's mainly maintained by @edolstra but @basvandijk and @grahamc also can merge changes.

Mic92:
I think @knl or @Ma27 could review the change to give the merger less work to do.

edolstra:
Doesn't this break the next line:
```
     revCount="$(cd "$inputDirTmp" && (git rev-list '[% input.revision %]' | wc -l))"
```

Ma27:
It says `1` then since only one ref is fetched, but at least I don't consider that an issue.

edolstra:
It means that the evaluation result may no longer be the same, which is a bad thing for a script called "reproduce".

c00w:
Sounds like I need to modify hydra to pass the revCount in as a parameter to the template.

c00w:
I'm 30% positive what I've done will work. The revCount is produced by the Git and Hg Plugins, but I totally can't figure out how the inputs to reproduce.tt get passed in (I see the controller file, but I don't see where it's actually rendered, or what type anything is).

If someone can point me in the right direction, that would be awesome.

PR #800, opened 2020-08-16
Replace bazaar input with breezy

ehmry:
Bazaar has been removed for Nixpkgs in favor of Breezy, update
the hydra buildInputs and Nixpkgs revision accordingly.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -154,6 +154,8 @@
               ];
           };
 
+          breezyOrBazaar = final.breezy or final.bazaar;
+
         in stdenv.mkDerivation {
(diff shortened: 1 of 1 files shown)


edolstra:
I think breezy doesn't exist yet in Nixpkgs 20.03, which is what Hydra uses (see flake.nix). So we should wait until we switch to 20.09, or write something like `(pkgs.breezy or pkgs.bazaar)`.

ehmry:
Fixed with an `pkgs.breezy or pkgs.bazaar`.

PR #802, opened 2020-09-01
Development-time improvements

grahamc:


examples/hello.nix
```diff
@@ -8,6 +8,11 @@ let
   # <nixpkgs> is set to the value designated by the nixpkgs input of the
   # jobset configuration.
   pkgs = (import <nixpkgs> {});
-in {
-  hello = pkgs.hello;
+in
+{
+  hello = pkgs.hello.overrideAttrs (
+    { ... }: {
+      never_hit_cache = toString builtins.currentTime;
```
edolstra, line 15: Hm, we probably shouldn't use undocumented impure functions in examples...

grahamc: Good point. I removed that commit.

edolstra:
Thanks!

PR #684, opened 2019-10-28
Add LDAP support

ajs124:
This closes #53.

The two main things that need fixing before this is ready are:

- The config parsing
  - ~Assuming a separate config file is fine, I could not manage to figure out how to load and apply it conditionally.~
  - Assuming these options should be put into `HYDRA_CONFIG`, I failed at figuring that out as well.
- The LDAP to hydra group/role mapping
  - Right now, it literally checks if they start with hydra: [1](https://github.com/NixOS/hydra/compare/master...helsinki-systems:ldap?expand=1#diff-65565fb99ce7a7d5d4887473a43888b5R57)
  - and then strips the first 6 characters from the roles: [2](https://github.com/NixOS/hydra/compare/master...helsinki-systems:ldap?expand=1#diff-65565fb99ce7a7d5d4887473a43888b5R78).

Furthermore, this introduces a dependency on [Catalyst::Authentication::Store::LDAP](https://metacpan.org/release/Catalyst-Authentication-Store-LDAP), which is not part of nixpkgs right now.

cc @dasJ @edef1c

PR #805, opened 2020-09-10
LDAP support continued (with missing packages & tests)

andir:
This is picking up the work from @ajs124 in #684.

The main difference to the previous PR is that I started writing tests for it (while I was trying to figure out if it is working). That work required adding the missing perl packages which I'm also adding to NixPkgs in https://github.com/NixOS/nixpkgs/pull/97673. For the meantime they should probably be kept in here until NixOS 21.03 is released.

As written in the original PR this still has the hard coded `hydra_` prefix for all the roles that are currently supported by hydra. IMO extending that is out of scope for an initial implementation. If someone really needs different names and/or has multiple hydra setups with different permissions for the same person they can still have a different OU's in their LDAP and filter accordingly.

### DIFF
--- a/doc/manual/installation.xml
+++ b/doc/manual/installation.xml
@@ -272,6 +272,62 @@ server {
 
     </para>
   </section>
+  <section>
+    <title>Using LDAP as authentication backend (optional)</title>
+    <para>
+      Instead of using Hydra's built-in user management you can optionally use LDAP to manage roles and users.
+    </para>
+
+    <para>
+      The <command>hydra-server</command> accepts the environment
+      variable <emphasis>HYDRA_LDAP_CONFIG</emphasis>. The value of
+      the variable should point to a valid YAML file containing the
+      Catalyst LDAP configuration. The format of the configuration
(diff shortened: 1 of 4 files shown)


andir:
I just deployed this to my private hydra instance and apparently this PR broke non-LDAP logins. I'll be looking into that today or latest on monday.

ajs124:
Oops. I could have sworn I tested that… That's what [this](https://github.com/andir/hydra/blob/b5d7ed2e99c6a32ed3d49fda9dfc7a4e7f13dc8f/src/lib/Hydra/Controller/User.pm#L34) was supposed to do…

andir:
On 07:58 12.09.20, ajs124 wrote:
> Oops. I could have sworn I tested that… That's what [this](https://github.com/andir/hydra/blob/b5d7ed2e99c6a32ed3d49fda9dfc7a4e7f13dc8f/src/lib/Hydra/Controller/User.pm#L34) was supposed to do…

Yeah, I get an error that the LDAP realm isn't available right now.

PR #806, opened 2020-09-11
fix build after nix commit 82b77a77262c414044fffc7ad8b955ad91827995

ajs124:


Ericson2314 (approved):

edolstra:
This should also update `flake.nix` unless the build already fails with the version of `nix` pinned there.

ajs124:
I assume you meant `flake.lock`? It builds with `nix build` now.

PR #811, opened 2020-09-13
Stop violating not null constraint

dasJ:
Fixes this error:

ERROR: failed to process declarative jobset test:inputs,
DBIx::Class::Storage::DBI::_dbh_execute(): DBI Exception: DBD::Pg::st
execute failed: ERROR:  null value in column "emailoverride" violates
not-null constraint

cc @basvandijk

### DIFF
--- a/src/lib/Hydra/Helper/AddBuilds.pm
+++ b/src/lib/Hydra/Helper/AddBuilds.pm
@@ -57,7 +57,7 @@ sub updateDeclarativeJobset {
                 name => $name,
                 type => $data->{type}
             };
-            $row->{emailresponsible} = $data->{emailresponsible} if defined $data->{emailresponsible};
+            $row->{emailresponsible} = $data->{emailresponsible} // 0;
             my $input = $jobset->jobsetinputs->create($row);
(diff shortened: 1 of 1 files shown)


src/lib/Hydra/Helper/AddBuilds.pm
```diff
@@ -57,7 +57,7 @@ sub updateDeclarativeJobset {
                 name => $name,
                 type => $data->{type}
             };
-            $row->{emailresponsible} = $data->{emailresponsible} if defined $data->{emailresponsible};
+            $row->{emailresponsible} = $data->{emailresponsible} or 0;
```
edolstra, line 60: Shouldn't this be `// 0`? I think `or` binds very weakly.

dasJ: But that doesn't matter, does it?

edolstra: Doesn't it parse as `($row->{emailresponsible} = $data->{emailresponsible}) or 0;`? I could be wrong.

dasJ:
@edolstra Fixed and rebased

PR #818, opened 2020-09-26
Fix build against latest Nix master

Ma27:
This PR contains an automatic `flake.lock` update (2394140) and updates the code in `hydra-queue-runner` to compile again (
9cc76f6). The details can be found in the commit messages.

cc @edolstra @grahamc @ericson2314 @lheckemann

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -263,9 +263,9 @@ void State::buildRemote(ref<Store> destStore,
             auto drv2 = localStore->readDerivation(input.first);
             for (auto & name : input.second) {
                 if (auto i = get(drv2.outputs, name)) {
-                    auto outPath = i->path(*localStore, drv2.name);
-                    inputs.insert(outPath);
-                    basicDrv.inputSrcs.insert(outPath);
+                    auto outPath = i->path(*localStore, drv2.name, name);
+                    inputs.insert(*outPath);
+                    basicDrv.inputSrcs.insert(*outPath);
                 }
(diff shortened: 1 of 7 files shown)


Ma27:
ping @edolstra anything missing to proceed here? :)

src/hydra-queue-runner/build-remote.cc
```diff
@@ -434,7 +434,11 @@ void State::buildRemote(ref<Store> destStore,
 
             auto now1 = std::chrono::steady_clock::now();
 
-            auto outputs = step->drv->outputPaths(*localStore);
+            StorePathSet outputs;
+            for (auto & i : step->drv->outputsAndOptPaths(*localStore)) {
```
edolstra, line 438: So I was going to suggest using `drv->outputsAndPaths()` here, but it seems to have disappeared mysteriously. (In a major git fail, `git log -p src/libstore/derivations.cc` shows it being added but not being removed...) @Ericson2314 Do you know why it was removed?

Ericson2314: It might have disappeared in a merge commit; I've noticed `git log -G` skipping that

Ma27: Does it mean that the patch is fine overall?

Ericson2314: I think this might have been intentionally removed because we didn't want people to unintentionally not support floating CA derivations.

The build-remote protocol needs to be changed for those anyways, so I wouldn't worry for now.

Issue #679, opened 2019-09-18
suggested replacement for binary_cache_secret_key_file is dysfunctional 


sbourdeauducq:
I was using ``binary_cache_secret_key_file=some_file`` and, after upgrading Hydra, this stopped working with the message:
```
hydra.conf: binary_cache_secret_key_file is deprecated and ignored. use store_uri=...?secret-key= instead
```

I have replaced it with ``store_uri=auto?secret-key=some_file`` and it is still unhappy:
```
hydra-queue-runner: warning: unknown setting 'secret-key'
```

What is the proper replacement for  ``binary_cache_secret_key_file=some_file``?
Why does hydra print confusing messages here?

edolstra:
`secret-key` is an option of `BinaryCacheStore`, so it doesn't work with the local store (i.e. `auto`). For the local store, you have to set `secret-key-files` in `nix.conf` (not `hydra.conf`).

sbourdeauducq:
Okay, then what is the recommended setting for serving signed binaries from the local store via Hydra?

twhitehead:
Sorry for the noise on this ticket. I accidentally typed 679 instead of 697 in my commit message.

kquick:
See also https://github.com/NixOS/hydra/issues/548

PR #820, opened 2020-09-30
Remove deprecation notice for binary_cache_secret_key_file

Taneb:
It's still used for signing `narinfo` files if using Hydra as a binary cache server. See https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/View/NARInfo.pm#L37


ivan:
Please merge this; the deprecation warning was very confusing to me as someone using a local store with `store_uri = auto`:

```
hydra-send-stats[3536553]: hydra.conf: binary_cache_secret_key_file is deprecated and ignored. use store_uri=...?secret-key= instead
```

1) It is **not ignored** with `store_uri = auto` and AFAIK `binary_cache_secret_key_file` is the only way have your packages be signed when they're downloaded over `ssh://` (unless you choose to run `nix store sign --all -k /var/secrets/nix/secret_key` all the time).
2) `store_uri=...?secret-key=` does not work for `auto` as mentioned in https://github.com/NixOS/hydra/issues/679#issuecomment-532607341

PR #825, opened 2020-10-25
Fix unhelpful error messages in aggregate jobs.

samueldr:
From the commit messages:

* * *

**hydra-eval-jobs: Identify unexpected errors in handling aggregate jobs**

The vague "[json.exception.type_error.302] type must be string, but is null" is **absolutely** unhelpful in the way Hydra currently handles it on evaluation.

This is handling *unexpected* errors only; the following commit will handle the specific instance of the previously mentioned error.

* * *

**hydra-eval-jobs: Transmit original Nix error when handling aggregate jobs**

It might happen that a job from the aggregate returned an error!

This is what the vague "[json.exception.type_error.302] type must be string, but is null" was all about in this instance; there was no `drvPath` to stringify!

So we now actively watch for errors and copy them to the aggregate job.

* * *

**Note**: This wasn't used inside Hydra. It was only ran this way:

```
env -i HYDRA_CONFIG=/Users/samuel/tmp/nixpkgs/hydra.conf /Users/samuel/Projects/nixos/hydra/./src/hydra-eval-jobs/hydra-eval-jobs -I /nix/store -I ./hydra-fail/ ./hydra-fail/nixos/release-small.nix
```

This was verified on top of https://github.com/NixOS/nixpkgs/commit/f262382e01636317bd5caf09c17e3883155545dd

So **I do not know if adding an `error` attribute on the aggregate job will work as expected**.

I'm also a bit exhausted from figuring out the error. I'll let someone that knows more about Hydra guide or fix the PR so the aggregate job properly reports that there was an error. Ideally, the aggregate job should be **failed**, otherwise we would have a missing job!

cc @andir


src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -442,45 +442,66 @@ int main(int argc, char * * argv)
         for (auto i = state->jobs.begin(); i != state->jobs.end(); ++i) {
             auto jobName = i.key();
             auto & job = i.value();
+            // For the error message
+            std::string lastTriedJobName = i.key();
```
samueldr, line 446: I guess the initialization here is pretty useless...

I think when I wrote that up originally I was thinking about using `auto` to intrinsically get the type. In that case `auto lastTriedJobName = jobName`  is probably better.

Otherwise, initializing with `""`.

What would be the better option?

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -442,45 +442,66 @@ int main(int argc, char * * argv)
...
+                    auto drvPath = store->parseStorePath((std::string) job["drvPath"]);
+                    auto drv = store->readDerivation(drvPath);
+
+                    for (std::string jobName2 : *named) {
+                        lastTriedJobName = jobName2;
+                        auto job2 = state->jobs.find(jobName2);
+                        if (job2 == state->jobs.end())
+                            throw Error("aggregate job '%s' references non-existent job '%s'", jobName, jobName2);
+
+                        if ((*job2).find("error") != (*job2).end()) {
```
samueldr, line 471: Now that tripped me up. I get that this comes from `auto job2` being the result of the search. Is there a pattern we could be using so that we don't have to dereference all the time?

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -442,45 +442,66 @@ int main(int argc, char * * argv)
...
+                            if (job.find("error") == job.end()) {
+                                job["error"] = fmt("Errors aggregating aggregate job '%1%'.\n", jobName);
+                            }
```
samueldr, line 474: Is there a better pattern to initialize the error message on the JSON object?

src/hydra-eval-jobs/hydra-eval-jobs.cc
andir: IIRC the nlohman value `key()` function returns reference. `auto` should then deduct the type `T&` which is probably not what we want here. The assignments in line 454 and 466 would then override the value in the JSON document. That being said my type deduction knowledge might be a bit rusty... I would stick with the string there.

andir (commented):
Just did a quick pass.

src/hydra-eval-jobs/hydra-eval-jobs.cc
andir: `auto& something = *job2` should do the trick here.

ju1m:
I don't know if this PR is ready, but just to let you know that merging it would unblock https://github.com/NixOS/nixpkgs/pull/101071

samueldr:
I've stated in the past on the IRC channels (a couple of times) that it needs to be reviewed by "code owners" of that section for the open questions.

Other than that, in my opinion it is ready, if not the exact code as shown, the same but with the corrections and nits fixed.

I know that, during my testing, it was working as expected.

lukegb:
This would have massively helped debugging https://github.com/NixOS/nixpkgs/issues/132328.

src/hydra-eval-jobs/hydra-eval-jobs.cc
lukegb: Honestly, this is _probably_ fine? I'd probably replace the `(*job2).` with just `job2->` but other than that this seems... fine.

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -442,45 +442,65 @@ int main(int argc, char * * argv)
...
+                            job["error"] = fmt("While handling '%1%': %2%\n", jobName2, (std::string) (*job2)["error"]);
```
ncfavier, line 475: Doesn't this always overwrite the assignment on line 473?

samueldr:
Just saying, today and possibly since two days ago `nixos:trunk-combined` is having a bad time again:

```
error: [json.exception.type_error.302] type must be string, but is null
```

And we can't know why.

grahamc:
This doesn't seem to do what it should.

PR #827, opened 2020-11-03
add test for nix-support/hydra-build-products edge case

ajs124:
This adds a tests for an edge case which I am yet to understand.
On current master, this test **fails**. You might be asking yourself "why add a failing test?". The answer is obvious, this test shouldn't fail. In fact, it doesn't (meaning it succeeds) when built against commits like b9ec1193d855c31a16afeda3aa73beb166ec0b24.

The reason why this feature is relevant, is for the `nixos.netboot` or jobs like it. As you can see, e.g. [here](https://hydra.nixos.org/build/129492856), it currently does not show any build results. In fact, that store path has an empty `nix-support` folder. Building the same thing with nix instead of hydra results in the correct `hydra-build-products`.

I'll try to bisect this, in the next few days, if I can find the time.

cc @grahamc @lassulus


ajs124:
Bisecting tells me that it broke in 5b4df3ad5a2b318a29b6bebc3e6be353b8c61d65 @edolstra 

To verify this, check out d4e4be4fd18d86826501e85e5f0195e49e2e3d1e and cherry pick this on top and observe that the test succeeds. Do the same with 5b4df3ad5a2b318a29b6bebc3e6be353b8c61d65 and it fails.

Lassulus:
so, what does happen now? would be nice to have those build products back :)

ajs124:
Now someone needs to figure out how/why 5b4df3ad5a2b318a29b6bebc3e6be353b8c61d65 broke this. I tried reverting it, but that (obviously) didn't work.

I can give it a try, but it would probably take me a full working day to try to figure out, assuming I can figure it out, because I really don't know enough about nix and hydra internals.

Kloenk:
Will try to find time to look into it.

Also CC @Ericson2314 maybe you have some ideas what are causing this bugs.

ajs124:
@edolstra

ajs124:
@edolstra

ctheune:
I'm really bad with C/C++, but here goes a shot: in my builds of netboot I can see that the nix-support/hydra-build-products file exists but isn't picked up by hydra.

The one small change I imagine could be relevant here is that the build products in this case references files from another place in the store that doesn't belong to the same NAR, so the change in https://github.com/NixOS/hydra/commit/5b4df3ad5a2b318a29b6bebc3e6be353b8c61d65#diff-6ae5dd4d7cad7e6c0000230c3c51eb5b9bdc92595ccf90d81e6e72544eefe18cR114 previously just grabbed the full store path from the filesystem but now fails to pick it up from the NAR because it's not part of the same archive ...

Issue #796, opened 2020-07-28
setting store_uri to file:// makes Hydra fail to serve content from system Nix store


sbourdeauducq:
When ``store_uri`` is set to a ``file://`` URL, then Hydra only serves content from this location. This causes problems in practice:
* If previous Hydra builds were done with ``store_uri=auto``, then they are no longer served after using ``store_uri=file://...`` and there is no simple way to copy them to ``store_uri`` (AFAIK).
* If something is built on the Hydra machine using manual ``nix-build`` and then added to Hydra (a possible scenario when testing changes), then Hydra will never serve those packages, since it sees they are already available in the system Nix store and it does not build and copy them to ``store_uri``.

I propose solving this by making Hydra first search the system Nix store when serving binaries, and then ``store_uri`` if it was not found in the system Nix store.

Doing it in this order potentially improves performance since the system Nix store is probably on faster storage than a large ``store_uri`` "archival" filesystem.

sbourdeauducq:
Some observations:
* Hydra invokes nix ``cat-store``/``ls-store`` to serve store contents.
* Nix itself does not support fetching from either location.
* Simplest way is likely to patch ``serveFile`` in ``src/lib/Hydra/Controller/Build.pm`` to try both Nix commands for the possible store locations.

sbourdeauducq:
>     * there is no simple way to copy them to `store_uri`

Using ``nix copy --to file://... /nix/store/missing_stuff`` is a bit unwieldy but works.

sbourdeauducq:
> is a bit unwieldy

... especially when you have huge files in the Nix store and it loads them entirely into RAM... sigh!

sbourdeauducq:
Actually this doesn't work, ``nix copy`` appears to copy the closures but then hydra won't serve them for some reason. Very frustrating...

sbourdeauducq:
It seems that using ``nix copy`` into the binary storage directory actually corrupts it somehow and then hydra won't serve *anything* from it. WTF...

edolstra:
Can you reproduce that? Sounds like a pretty serious bug...

sbourdeauducq:
May not have been caused by ``nix copy`` but I am now in a situation where Hydra won't serve anything unless I set ``store_uri`` back to ``auto`` and I do not know how to create or fix that situation. I tried wiping the store and starting over again - it still won't work and I do not know why it worked before, at least for some packages.
[...]

PR #829, opened 2020-11-05
Support for serving binary cache from local binary cache store

astro:
Related to #796 

This adds the capability of running with a store_uri of `file://...` that points to a local binary cache store, signing .narinfo on the fly.

I'm not really fluent in Perl, hence I appreciate a throrough review.


astro:
CI failure unrelated

sbourdeauducq:
There is at least one problem with this patch: when it is applied, accessing invalid NAR URLs (e.g. ``http://hydra/nar/bogusfile``) causes an empty file to be served instead of a 404 error.

PR #836, opened 2020-11-21
Support restarting components as admin

dasJ:
It's tedious to always log in when one of the components hangs. Instead,
expose buttons for admins to kill each process, letting systemd restart
them (they all have Restart=always).


dasJ:
Not that I expect anyone to care about this, but it looks like this:

image

cole-h:
I don't think the buttons should say `Restart` on them, when all it's doing is killing the component and letting systemd restart it. If, for whatever reason, somebody changes the unit to not be `Restart=always`, they might wonder why the button `Restart Evaluator` didn't actually restart the evaluator.

Really just splitting hairs, but I think it's better to name it what it does, not what it should do.

dasJ:
@cole-h Like this?

PR #841, opened 2020-12-26
Implement GitHub logins

pingiun:
Requires the following configuration options
`github_client_id`
`github_client_secret` or `github_client_secret_file`

### DIFF
--- a/src/lib/Hydra/Controller/User.pm
+++ b/src/lib/Hydra/Controller/User.pm
@@ -181,11 +181,23 @@ sub github_login :Path('/github-login') Args(0) {
     my $data = decode_json($response->decoded_content) or die;
     my $access_token = $data->{access_token} // die "No access_token in response from GitHub.";
 
-    $response = $ua->get('https://api.github.com/user', Authorization => "token $access_token");
-    error($c, "Did not get a response from GitHub for user info.") unless $response->is_success;
+    $response = $ua->get('https://api.github.com/user/emails', Accept => 'application/vnd.github.v3+json', Authorization => "token $access_token");
+    error($c, "Did not get a response from GitHub for email info.") unless $response->is_success;
 
     $data = decode_json($response->decoded_content) or die;
-    doEmailLogin($self, $c, "github", $data->{email}, $data->{name} // undef);
+    my $email;
+
+    foreach my $eml (@{$data}) {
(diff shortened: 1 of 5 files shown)


src/lib/Hydra/Controller/User.pm
```diff
@@ -154,6 +155,69 @@ sub google_login :Path('/google-login') Args(0) {
...
+
+    $response = $ua->get('https://api.github.com/user/emails', Accept => 'application/vnd.github.v3+json', Authorization => "token $access_token");
+    error($c, "Did not get a response from GitHub for email info.") unless $response->is_success;
+
+    $data = decode_json($response->decoded_content) or die;
+    my $email;
+
+    foreach my $eml (@{$data}) {
+        $email = $eml->{email} if $eml->{verified} && $eml->{primary};
+        print STDERR "$eml->{email}\n";
```
edolstra, line 192:
```suggestion
```

src/lib/Hydra/Controller/User.pm
```diff
@@ -154,6 +155,69 @@ sub google_login :Path('/google-login') Args(0) {
...
+    }
+
+    print STDERR "$email\n";
```
edolstra, line 195:
```suggestion
```

src/lib/Hydra/Controller/User.pm
```diff
@@ -154,6 +155,69 @@ sub google_login :Path('/google-login') Args(0) {
...
+    $response = $ua->get('https://api.github.com/user', Authorization => "token $access_token");
+    error($c, "Did not get a response from GitHub for user info.") unless $response->is_success;
+    $data = decode_json($response->decoded_content) or die;
+
+    doEmailLogin($self, $c, "github", $email, $data->{name} // undef);
```
edolstra, line 200: Can `$email` be uninitialized here?

pingiun: I don't think github users can delete their primary email address, so this could only happen when github returns faulty data. I will add a `die` here to catch this though, as it seems like something useful to be caught

edolstra:
Thanks!

edolstra:
I've enabled this on hydra.nixos.org, so GitHub logins are now possible! The only caveat is that you can't login using GitHub if the GitHub email address is equal to an already existing Google login.

PR #847, opened 2021-01-21
JobsetEvals: record evaluation errors

grahamc:
Today, Hydra records evaluation errors in the Jobsets table only. When a new evaluation happens, the old log is lost. This can be pretty annoying if you're trying to debug an issue and a new evaluation finishes while you're hunting.

This PR adds evaluation errors to the jobsetevals table and writes the evaluation log to it in addition to storing the latest evaluation error to jobsets.

The included migration copies the jobset's evaluation error output to the latest jobsetevals record for that jobset. It shouldn't take long to execute this migration: it takes a minute or so to run against hydra's production dataset, on a significantly underpowered server.

It looks like this:

### DIFF
--- a/src/root/common.tt
+++ b/src/root/common.tt
@@ -476,6 +476,9 @@ BLOCK renderEvals %]
             ELSE %]
              -
             [% END %]
+            [% IF eval.errormsg %]
+              <span class="label label-warning">Eval Errors</span>
+            [% END %]
           </td>
           <td align='right' class="nowrap">
             <span class="label label-success">[% e.nrSucceeded %]</span>
--- a/tests/set-up.pl
+++ b/tests/set-up.pl
@@ -1,5 +1,5 @@
 use strict;
-system("initdb -D postgres") == 0 or die;
-system("pg_ctl -D postgres -o \"-F -p 6433 -h '' -k /tmp \" -w start") == 0 or die;
-system("createdb -p 6433 hydra-test-suite") == 0 or die;
+system("initdb -D postgres --locale C.UTF-8 ") == 0 or die;
+system("pg_ctl -D postgres -o \"-F -p 6433  -h '' -k /tmp \" -w start") == 0 or die;
+system("createdb -l C.UTF-8 -p 6433 hydra-test-suite") == 0 or die;
 system("hydra-init") == 0 or die;
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,7 @@ hydra-config.h.in
 result
 tests/jobs/config.nix
 outputs
+config
+stamp-h1
+src/hydra-evaluator/hydra-evaluator
+src/hydra-queue-runner/hydra-queue-runner
--- a/src/lib/Hydra/Schema/JobsetEvals.pm
+++ b/src/lib/Hydra/Schema/JobsetEvals.pm
@@ -54,6 +54,16 @@ __PACKAGE__->table("jobsetevals");
   is_foreign_key: 1
   is_nullable: 0
 
+=head2 errormsg
+
+  data_type: 'text'
+  is_nullable: 1
+
+=head2 errortime
+
+  data_type: 'integer'
+  is_nullable: 1
+
 =head2 timestamp
 
   data_type: 'integer'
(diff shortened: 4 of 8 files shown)


andir (approved):

edolstra:
Not sure about this. The reason we don't keep old evaluation logs is that they're huge (especially for the Nixpkgs/NixOS jobsets) so this would clutter the database a lot.

grahamc:
What if we had a process by which only the most recent ~5 per jobset kept logs?

samueldr:
@edolstra The alternative (current situation) is that we cannot trace when an evaluation error first started happening. It would help reduce the bounds to bisect from.

Assuming this failed for four evals: 

 - last good eval (`aaaaaaaa`...`bbbbbbbb`)
 - first failure (`bbbbbbbb`...`ccccccccc`)
 - (repeat) (`cccccccc`...`dddddddd`)
 - (repeat) (`dddddddd`...`eeeeeeee`)
 - (repeat) (`eeeeeeee`...`ffffffff`)

With the current situation, we have to bisect from `bbbbbbbb` to `ffffffff`.

With these changes (AFAIUI) we can assume *at least one failure* is found between `bbbbbbbb` and `cccccccc`.

* * *

And then, this is only for when evaluation is assumed to be reproducibly healthy.

Imagine we face another issue like the one from before the new evaluator, where sometimes eval failed spuriously. Currently we lose the data about failed evals, which can make understanding, or even proving the issue harder.

* * *

Though, keeping *old historical* data I understand is of much less value. Would it be less of an issue if they weren't kept in the database, but as files?

AmineChikhaoui:
What about moving away from storing eval logs in the database and instead do something like build logs either in the filesystem (/var/lib/hydra/eval-logs/) or S3 ?

src/root/common.tt
```diff
@@ -476,6 +476,9 @@ BLOCK renderEvals %]
             ELSE %]
              -
             [% END %]
+            [% IF eval.errormsg %]
+              <span class="label label-warning">Eval Errors</span></td>
```
samueldr, line 480: I need to spin up an hydra instance, cause eval errors, but that `</td>` at first glance looks oddly unmatched.

edolstra:
> Would it be less of an issue if they weren't kept in the database, but as files?

No, since it would be on the same disk. Uploading to S3 would probably be a bit too much effort.

> What if we had a process by which only the most recent ~5 per jobset kept logs?

Yeah that would be good. No need to add it to this PR, we can just run a systemd timer on hydra.nixos.org to delete old logs for the nixpkgs/nixos jobsets.

samueldr (changes requested):
I'm not comfortable with the SQL parts and the migration. So don't assume I approve of it outright, but it does look okay.

* * *

There is that one `</td>` that I think is wrong. Though I'm spinning up a dev instance with this PR and causing errors to see.

tests/set-up.pl
```diff
@@ -1,5 +1,5 @@
 use strict;
-system("initdb -D postgres") == 0 or die;
-system("pg_ctl -D postgres -o \"-F -p 6433 -h '' -k /tmp \" -w start") == 0 or die;
-system("createdb -p 6433 hydra-test-suite") == 0 or die;
+system("initdb -D postgres --locale C.UTF-8 ") == 0 or die;
+system("pg_ctl -D postgres -o \"-F -p 6433  -h '' -k /tmp \" -w start") == 0 or die;
+system("createdb -l C.UTF-8 -p 6433 hydra-test-suite") == 0 or die;
```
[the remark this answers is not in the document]
grahamc: h.n.o is actually using en_US.UTF-8 but that felt a bit too us-centric. The major issue this is addressing is we're writing wide characters  to the DB, so any UTF-8 should do.

src/root/common.tt
grahamc: Whoops!

grahamc: Fixed up!

samueldr (approved):
FWIW, looks fine here.

grahamc:
This created a fairly significant amount of extra data transfer, as DBIx automatically fetches the column's value. It appears that disabling this column by default is not trivial, and that a simpler option would be having a separate table for evaluation errors. I'm going to work on this and try to get it in for review. In the meantime I've deleted all the evaluation messages.

PR #848, opened 2021-01-22
Normalize nixexpr{input,path} from builds to jobsetevals.

grahamc:
I meant to open this as a draft.

 Duplicating this data on every record of the builds table cost
approximately 4G of duplication.

Note that the database migration included took about 4h45m on an
untuned server which uses very slow rotational disks in a RAID5 setup,
with not a lot of RAM. I imagine in production it might take an hour
or two, but not 4. If this should become a chunked migration, I can do
that.

Note: Because of the question about chunked migrations, I have NOT
YET tested this migration thoroughly enough for merge.

### DIFF
--- a/src/lib/Hydra/Schema/Builds.pm
+++ b/src/lib/Hydra/Schema/Builds.pm
@@ -134,16 +134,6 @@ __PACKAGE__->table("builds");
   default_value: 0
   is_nullable: 1
 
-=head2 nixexprinput
-
-  data_type: 'text'
-  is_nullable: 1
-
-=head2 nixexprpath
-
-  data_type: 'text'
-  is_nullable: 1
-
 =head2 priority
 
   data_type: 'integer'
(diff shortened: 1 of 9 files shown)


grahamc:
> Duplicating this data on every record of the builds table cost
> approximately 4G of duplication.
> 
> Note that the database migration included took about 4h45m on an
> untuned server which uses very slow rotational disks in a RAID5 setup,
> with not a lot of RAM. I imagine in production it might take an hour
> or two, but not 4. If this should become a chunked migration, I can do
> that.
> 
> Note: Because of the question about chunked migrations, I have NOT
> YET tested this migration thoroughly enough for merge.

Ericson2314 (approved):
Seeing Hydra's poor track record on having the appropriate indices (per your earlier PR) I whole-heatedly endorse this. Hydra needs to prove it can follow the rules (well-planned joins) before it earns the right to break them (denormalization).

grahamc:
For what it is worth I find Hydra's schema to be pretty good :). It has scaled to over a hundred million builds, and its largest table has 1.5 billion (1,500,000,000) rows. There are a few cleanups to be done, but all in it does a really great job for the amount of maintenance it requires

edolstra:
> denormalization

FWIW, this is because the JobsetEvals table is a later addition. There is some more cruft in the DB from the time that we didn't have JobsetEvals, like BuildInputs (which still exists but isn't maintained anymore).

PR #850, opened 2021-01-26
Jobset -> JobsetEvals by JobsetEvals.jobset_id

grahamc:
Instead of referring to the jobset by its project's name and the jobset's name, refer to it by a jobset_id.

* I added a little in-line notes about how to write database migrations.
* I fixed a problem where the `make` for updating the schema libraries would succeed even if the SQL was invalid.
* I fixed a minor problem where hydra-dev-server was buffering output to stderr/stdout
* The C++ in hydra-evaluator feels a bit heavy on boilerplate, but it seemed like a good way to go. I'm open to doing something different. I wrote this in a few separate commits to make it easier for me to grok.
* The perl patches were surprisingly short and easy. I clicked around, ran the tests, ran some evaluations and builds

### DIFF
--- a/src/hydra-evaluator/hydra-evaluator.cc
+++ b/src/hydra-evaluator/hydra-evaluator.cc
@@ -68,7 +68,9 @@ struct Evaluator
         pqxx::work txn(*conn);
 
         auto res = txn.exec
-            ("select project, j.name, lastCheckedTime, triggerTime, checkInterval, j.enabled as jobset_enabled from Jobsets j join Projects p on j.project = p.name "
+            ("select project, j.name, lastCheckedTime, triggerTime, checkInterval, j.enabled as jobset_enabled "
+             "from Jobsets j "
+             "join Projects p on j.project = p.name "
              "where j.enabled != 0 and p.enabled != 0");
 
 
--- a/src/hydra-evaluator/hydra-evaluator.cc
+++ b/src/hydra-evaluator/hydra-evaluator.cc
@@ -13,7 +13,7 @@
 
 using namespace nix;
 
-typedef std::pair<std::string, std::string> JobsetName;
+typedef std::pair<std::string, std::string> JobsetIdentity;
 
 enum class EvaluationStyle
 {
@@ -30,16 +30,16 @@ struct Evaluator
 
     struct Jobset
     {
-        JobsetName name;
+        JobsetIdentity name;
         std::optional<EvaluationStyle> evaluation_style;
         time_t lastCheckedTime, triggerTime;
         int checkInterval;
         Pid pid;
     };
 
-    typedef std::map<JobsetName, Jobset> Jobsets;
+    typedef std::map<JobsetIdentity, Jobset> Jobsets;
 
-    std::optional<JobsetName> evalOne;
+    std::optional<JobsetIdentity> evalOne;
 
     const size_t maxEvals;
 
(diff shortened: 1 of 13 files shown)


src/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -13,7 +13,57 @@
 
 using namespace nix;
 
-typedef std::pair<std::string, std::string> JobsetName;
+typedef std::pair<std::string, std::string> JobsetPartialIdentity;
```
grahamc, line 16: edef suggests rename to `JobsetSymbolicIdentity`

src/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -13,7 +13,57 @@
 
 using namespace nix;
 
-typedef std::pair<std::string, std::string> JobsetName;
+typedef std::pair<std::string, std::string> JobsetSymbolicIdentity;
+
+class JobsetIdentity {
```
edolstra, line 18: Maybe just call this `JobsetId` to be less verbose?

src/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -13,7 +13,57 @@
 
 using namespace nix;
 
-typedef std::pair<std::string, std::string> JobsetName;
+typedef std::pair<std::string, std::string> JobsetSymbolicIdentity;
```
edolstra, line 16: I would just keep this as `JobsetName` since "symbolic identity" sounds like a fancy way of saying "name" :-)

src/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -13,7 +13,57 @@
...
+
+class JobsetIdentity {
+    public:
+
+    std::string project;
+    std::string jobset;
+    int id;
+
+
+    JobsetIdentity(const std::string& project, const std::string& jobset, const int id)
```
edolstra, line 26:
```suggestion
    JobsetIdentity(const std::string & project, const std::string & jobset, int id)
```

src/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -13,7 +13,57 @@
...
+        : project{ project }, jobset{ jobset }, id{ id }
+    {
+    }
+
+    friend bool operator== (const JobsetIdentity &lhs, const JobsetIdentity &rhs);
```
edolstra, line 31:
```suggestion
    friend bool operator== (const JobsetIdentity & lhs, const JobsetIdentity & rhs);
```

The style in Nix and Hydra is to have a space on both sides of `&` and `*`.

edolstra (approved):

src/hydra-evaluator/hydra-evaluator.cc
grahamc: Good idea. I went verbose to make the refactor easier, cleaning up makes sense. :+1:

src/hydra-evaluator/hydra-evaluator.cc
grahamc: :+1:

src/hydra-evaluator/hydra-evaluator.cc
grahamc: :+1:

grahamc:
Thanks for the review, @edolstra! I'll merge once it goes green.

PR #852, opened 2021-01-29
Add github notifications for evals

pingiun:
This replaces #712

I have used this functionality on my own server for a while and it seems to work fine. I adapted the PR created by @pikajude so it's based on the latest master, it uses python instead of perl tests, and it includes the suggestions from @knl 

Also in response this point of @knl:
> Maybe it would be good to have unified eval context (instead of one being named after the PR), so one could turn on branch protection on the context being successful.

I believe this is possible because you have quite a lot of freedom in the context name via StringInterpolate

If this PR is merged, #712 can be closed


pingiun:
I choose to have @pikajude as the co-author instead of committing as them, because I think the commit wouldn't be verified in that case? I'm not sure, let me know if I should change this

pingiun:
Changing this to draft because I am still trying to get the new test case to succeed

pingiun:
The new check is now succeeding on my hydra instance: https://hydra.technicie.nl/build/499

grahamc:
I wonder if it would be possible to test this feature similarly to https://github.com/NixOS/hydra/pull/878, without a separate, and relatively quite heavy VM test?

pingiun:
> I wonder if it would be possible to test this feature similarly to #878, without a separate, and relatively quite heavy VM test?

Good idea, I'll try it out

PR #854, opened 2021-01-30
search: limit queries to 20s

grahamc:
Even 20s is really long, but it cuts off queries which are today
running for 500+s.

### DIFF
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -442,50 +442,53 @@ sub search :Local Args(0) {
     my $limit = int(trim ($c->request->params->{"limit"} || "10"));
     $c->stash->{limit} = min(50, max(1, $limit));
 
-    $c->stash->{projects} = [ $c->model('DB::Projects')->search(
-        { -and =>
-            [ { -or => [ name => { ilike => "%$query%" }, displayName => { ilike => "%$query%" }, description => { ilike => "%$query%" } ] }
-            , { hidden => 0 }
-            ]
-        },
-        { order_by => ["name"] } ) ];
+    $c->model('DB')->schema->txn_do(sub {
+        $c->model('DB')->schema->storage->dbh->do("SET LOCAL statement_timeout = 20000");
+        $c->stash->{projects} = [ $c->model('DB::Projects')->search(
+            { -and =>
+                [ { -or => [ name => { ilike => "%$query%" }, displayName => { ilike => "%$query%" }, description => { ilike => "%$query%" } ] }
+                , { hidden => 0 }
+                ]
+            },
+            { order_by => ["name"] } ) ];
 
-    $c->stash->{jobsets} = [ $c->model('DB::Jobsets')->search(
-        { -and =>
-            [ { -or => [ "me.name" => { ilike => "%$query%" }, "me.description" => { ilike => "%$query%" } ] }
-            , { "project.hidden" => 0, "me.hidden" => 0 }
-            ]
-        },
-        { order_by => ["project", "name"], join => ["project"] } ) ];
+        $c->stash->{jobsets} = [ $c->model('DB::Jobsets')->search(
+            { -and =>
+                [ { -or => [ "me.name" => { ilike => "%$query%" }, "me.description" => { ilike => "%$query%" } ] }
+                , { "project.hidden" => 0, "me.hidden" => 0 }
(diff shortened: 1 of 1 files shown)


grahamc:
For example, https://hydra.nixos.org/search?query=chromium takes about 16s in total. Changing to pkgs.chromium takes 500+s.

src/lib/Hydra/Controller/Root.pm
```diff
@@ -442,50 +442,53 @@ sub search :Local Args(0) {
...
-        { order_by => ["id desc"]
-        , rows => $c->stash->{limit}
-        } ) ];
-
-    $c->stash->{resource} = { projects => $c->stash->{projects},
-                              jobsets  => $c->stash->{jobsets},
-                              builds  => $c->stash->{builds},
-                              buildsdrv  => $c->stash->{buildsdrv} };
+    $c->model('DB')->schema->txn_do(sub {
+        $c->model('DB')->schema->storage->dbh->prepare("SET statement_timeout = 20000;")->execute();
```
andir, line 446: What is the difference compared to `SET LOCAL statement_timeout …`? Does this have the possibility to escape the transaction?

andir: According to https://www.postgresql.org/docs/10/sql-set.html this would set the timeout for all other queries, even those outside of the transaction. We probably want `SET LOCAL` instead:

```
The effects of SET LOCAL last only till the end of the current transaction, whether committed or not. A special case is SET followed by SET LOCAL within a single transaction: the SET LOCAL value will be seen until the end of the transaction, but afterwards (if the transaction is committed) the SET value will take effect.
```

grahamc: Good catch.

> If SET (or equivalently SET SESSION) is issued within a transaction that is later aborted, the effects of the SET command disappear when the transaction is rolled back. Once the surrounding transaction is committed, the effects will persist until the end of the session, unless overridden by another SET.
>
>The effects of SET LOCAL last only till the end of the current transaction, whether committed or not. A special case is SET followed by SET LOCAL within a single transaction: the SET LOCAL value will be seen until the end of the transaction, but afterwards (if the transaction is committed) the SET value will take effect.

Changing to SET LOCAL.

grahamc: Fixed, tested locally that it still works. Thanks!

src/lib/Hydra/Controller/Root.pm
```diff
@@ -442,50 +442,53 @@ sub search :Local Args(0) {
...
-        { order_by => ["id desc"]
-        , rows => $c->stash->{limit}
-        } ) ];
-
-    $c->stash->{resource} = { projects => $c->stash->{projects},
-                              jobsets  => $c->stash->{jobsets},
-                              builds  => $c->stash->{builds},
-                              buildsdrv  => $c->stash->{buildsdrv} };
+    $c->model('DB')->schema->txn_do(sub {
+        $c->model('DB')->schema->storage->dbh->prepare("SET LOCAL statement_timeout = 20000;")->execute();
```
lukegb, line 446: Why prepare->execute rather than just `do`?

grahamc: No reason than muddling through the DBIx docs made me just find other examples in Hydra :sweat_smile:

andir (approved):

src/lib/Hydra/Controller/Root.pm
grahamc: Fixup'd.

andir (approved):

edolstra:
Maybe we should just remove the search feature if it's that expensive. Or limit it to things that can be queried efficiently.

grahamc:
I have some in progress work which improves search from ~15s to ~1s, so

not sure we should ditch it already. 

On January 31, 2021, GitHub <notifications@github.com> wrote:
> Maybe we should just remove the search feature if it's that expensive.

> Or limit it to things that can be queried efficiently.
>

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/NixOS/hydra/pull/854#issuecomment-770424198>, or
unsubscribe <https://github.com/notifications/unsubscribe-
auth/AAASXLHJMNC7UYQGDH7PKU3S4WMHLANCNFSM4W2OHJCA>.

PR #855, opened 2021-02-01
JobsetEvals: fixup permission references

grahamc:
Going from an eval to a project now requires hopping through the jobset

Reported by @andir.


andir (approved):

andir:
Deployed to my hydra and the issue is gone.

PR #856, opened 2021-02-01
Fix check in jobsets

immae:
The current check happening in jobsets is incorrect.
The wanted constraint is stated as follow :
- If type is 0 (legacy), then the flake field should be null, and both nixExprInput and nixExprPath should be non-null
- If type is 1 (flake), then the flake field should be non-null, and both nixExprInput and nixExprPath should be null

The current version will not catch (i.e. it will accept) situations
where you have for instance :
```
type = 1, nixExprPath null, nixExprInput non-null, flake non-null
```

This commit fixes that.

I split(ted) that into two constraints, to make it more readable and
easier to extend if a new type appears in the future.

The complete query could be instead :
```
( type = 0
  AND nixExprInput IS NOT NULL AND nixExprPath IS NOT NULL AND flake IS NULL )
OR ( type = 1
  AND nixExprInput IS NULL AND nixExprPath IS NULL AND flake IS NOT NULL )
```

(but an "OR" cannot be split, hence the other formulation)

At the meantime I statically named the checks, because otherwise it’s impossible to guess their name (in case we need to update them later)

cc @grahamc

### DIFF
--- a/src/sql/hydra.sql
+++ b/src/sql/hydra.sql
@@ -84,9 +84,14 @@ create table Jobsets (
     startTime     integer, -- if jobset is currently running
     type          integer not null default 0, -- 0 == legacy, 1 == flake
     flake         text,
-    check (schedulingShares > 0),
-    check ((type = 0) = (nixExprInput is not null and nixExprPath is not null)),
-    check ((type = 1) = (flake is not null)),
+    constraint jobsets_schedulingshares_nonzero_check check (schedulingShares > 0),
+    constraint jobsets_type_known_check   check (type = 0 or type = 1),
+    -- If the type is 0, then nixExprInput and nixExprPath should be non-null and other type-specific fields should be null
+    -- Otherwise the check passes
+    constraint jobsets_legacy_paths_check check ((type = 0) = (nixExprInput is not null and nixExprPath is not null and flake is     null)),
+    -- If the type is 1, then flake should be non-null and other type-specific fields should be null
+    -- Otherwise the check passes
+    constraint jobsets_flake_paths_check  check ((type = 1) = (nixExprInput is     null and nixExprPath is     null and flake is not null)),
     primary key   (project, name),
     foreign key   (project) references Projects(name) on delete cascade on update cascade,
     constraint    Jobsets_id_unique UNIQUE(id)
--- a/src/sql/upgrade-74.sql
+++ b/src/sql/upgrade-74.sql
@@ -1,0 +1,9 @@
+alter table Jobsets add constraint jobsets_type_known_check   check (type = 0 or type = 1);
+alter table Jobsets add constraint jobsets_legacy_paths_check check ((type = 0) = (nixExprInput is not null and nixExprPath is not null and flake is     null));
+alter table Jobsets add constraint jobsets_flake_paths_check  check ((type = 1) = (nixExprInput is     null and nixExprPath is     null and flake is not null));
+alter table Jobsets add constraint jobsets_schedulingshares_nonzero_check check (schedulingShares > 0);
+
+alter table Jobsets drop constraint if exists jobsets_schedulingshares_check;
+alter table Jobsets drop constraint if exists jobsets_check;
+alter table Jobsets drop constraint if exists jobsets_check1;
+alter table Jobsets drop constraint if exists jobsets_check2;


andir (commented):
The second bulletpoint in the commit message isn't clear to me:
```
... and both nixExprInput and nixExprPath
```
It reads as if there is something missing at the end

src/sql/hydra.sql
```diff
@@ -85,8 +85,8 @@ create table Jobsets (
     type          integer not null default 0, -- 0 == legacy, 1 == flake
     flake         text,
     check (schedulingShares > 0),
-    check ((type = 0) = (nixExprInput is not null and nixExprPath is not null)),
-    check ((type = 1) = (flake is not null)),
+    constraint jobsets_check_legacy check ((type != 0) OR (nixExprInput is not null and nixExprPath is not null and flake is     null)),
```
andir, line 88: Now that we have identified that this is already too complex for the "average developer" to maintain we might want to add a comment here to inform the next person coming along.

The commit message is doing a pretty good job on explaining how it is supposed to be. Mind adding two lines here that do the same?

immae:
> The second bulletpoint in the commit message isn't clear to me:
> 
> ```
> ... and both nixExprInput and nixExprPath
> ```
> 
> It reads as if there is something missing at the end

"the flake field should be null, and both nixExprInput and nixExprPath should be non-null", it was a (possibly too short) shortcut for "the flake field should be null, and nixExprInput should be non-null, and nixExprPath should be non-null", I’ll rephrase that :)

immae:
Thanks for the review @andir . I took your comments into account and fixed the text of the commit.

However, I thought that travis would make a live check, but it seems like it doesn’t. Please give me some time to correctly check it before merging.

immae:
@andir :
I made those tests locally:
- "nix flake check" on the present commit
- from a brand new psql server, I imported the old hydra.sql and the update-72.sql

If there is anything more you think that needs test, please tell me

andir:
> @andir :
> I made those tests locally:
> - "nix flake check" on the present commit
> - from a brand new psql server, I imported the old hydra.sql and the update-72.sql
> 
> If there is anything more you think that needs test, please tell me

Thanks but I can't merge a thing here. I just like to review properly written commit messages :)

grahamc:
As I mentioned in IRC, I'm a bit concerned about what names the constrains will get, since right now they're unnamed.

Here is wha hydra.nixos.org has for its jobsets table:

```
hydra=> \d+ jobsets;
                                                     Table "public.jobsets"
      Column      |  Type   | Collation | Nullable |               Default               | Storage  | Stats target | Description 
------------------+---------+-----------+----------+-------------------------------------+----------+--------------+-------------
 name             | text    |           | not null |                                     | extended |              | 
 project          | text    |           | not null |                                     | extended |              | 
 description      | text    |           |          |                                     | extended |              | 
 nixexprinput     | text    |           |          |                                     | extended |              | 
 nixexprpath      | text    |           |          |                                     | extended |              | 
 errormsg         | text    |           |          |                                     | extended |              | 
 errortime        | integer |           |          |                                     | plain    |              | 
 lastcheckedtime  | integer |           |          |                                     | plain    |              | 
 enabled          | integer |           | not null | 1                                   | plain    |              | 
 enableemail      | integer |           |          | 1                                   | plain    |              | 
 emailoverride    | text    |           |          | ''::text                            | extended |              | 
 hidden           | integer |           | not null | 0                                   | plain    |              | 
 keepnr           | integer |           |          | 3                                   | plain    |              | 
 triggertime      | integer |           |          |                                     | plain    |              | 
 checkinterval    | integer |           | not null | 300                                 | plain    |              | 
 schedulingshares | integer |           | not null | 100                                 | plain    |              | 
 fetcherrormsg    | text    |           |          |                                     | extended |              | 
 forceeval        | boolean |           |          |                                     | plain    |              | 
 starttime        | integer |           |          |                                     | plain    |              | 
 type             | integer |           | not null | 0                                   | plain    |              | 
 flake            | text    |           |          |                                     | extended |              | 
 id               | integer |           | not null | nextval('jobsets_id_seq'::regclass) | plain    |              | 
Indexes:
    "jobsets_pkey" PRIMARY KEY, btree (project, name)
    "jobsets_id_unique" UNIQUE CONSTRAINT, btree (id)
Check constraints:
    "jobsets_check" CHECK (schedulingshares > 0)
    "jobsets_check1" CHECK ((type = 0) = (nixexprinput IS NOT NULL AND nixexprpath IS NOT NULL))
    "jobsets_check2" CHECK ((type = 1) = (flake IS NOT NULL))
Foreign-key constraints:
    "fk1" FOREIGN KEY (project) REFERENCES projects(name) ON UPDATE CASCADE ON DELETE CASCADE
Referenced by:
    TABLE "buildmetrics" CONSTRAINT "buildmetrics_project_fkey1" FOREIGN KEY (project, jobset) REFERENCES jobsets(project, name) ON UPDATE CASCADE
    TABLE "builds" CONSTRAINT "builds_jobset_id_fkey" FOREIGN KEY (jobset_id) REFERENCES jobsets(id) ON DELETE CASCADE
    TABLE "builds" CONSTRAINT "builds_jobset_id_fkey1" FOREIGN KEY (jobset_id) REFERENCES jobsets(id) ON DELETE CASCADE
    TABLE "jobsetinputs" CONSTRAINT "fk1" FOREIGN KEY (project, jobset) REFERENCES jobsets(project, name) ON UPDATE CASCADE ON DELETE CASCADE
    TABLE "builds" CONSTRAINT "fk2" FOREIGN KEY (project, jobset) REFERENCES jobsets(project, name) ON UPDATE CASCADE
    TABLE "jobsetevals" CONSTRAINT "jobsetevals_jobset_id_fkey" FOREIGN KEY (jobset_id) REFERENCES jobsets(id) ON DELETE CASCADE
    TABLE "jobsetrenames" CONSTRAINT "jobsetrenames_project_fkey1" FOREIGN KEY (project, to_) REFERENCES jobsets(project, name) ON UPDATE CASCADE ON DELETE CASCADE
    TABLE "starredjobs" CONSTRAINT "starredjobs_project_fkey1" FOREIGN KEY (project, jobset) REFERENCES jobsets(project, name) ON UPDATE CASCADE ON DELETE CASCADE
Triggers:
    jobsetsadded AFTER INSERT ON jobsets FOR EACH STATEMENT EXECUTE PROCEDURE notifyjobsetsadded()
    jobsetschedulingchanged AFTER UPDATE ON jobsets FOR EACH ROW WHEN (old.triggertime IS DISTINCT FROM new.triggertime AND new.triggertime IS NOT NULL OR old.checkinterval <> new.checkinterval OR old.enabled <> new.enabled) EXECUTE PROCEDURE notifyjobsetschedulingchanged()
    jobsetsdeleted AFTER DELETE ON jobsets FOR EACH STATEMENT EXECUTE PROCEDURE notifyjobsetsdeleted()
    jobsetshareschanged AFTER UPDATE ON jobsets FOR EACH ROW WHEN (old.schedulingshares <> new.schedulingshares) EXECUTE PROCEDURE notifyjobsetshareschanged()
```

We should compare this to what a brand new, just-created hydra database looks like... presumably it would only be one of the two: a brand new version, and the oldest hydra... and hopefully they're the same.

immae:
> Check constraints:
>     "jobsets_check" CHECK (schedulingshares > 0)
>     "jobsets_check1" CHECK ((type = 0) = (nixexprinput IS NOT NULL AND nixexprpath IS NOT NULL))
>     "jobsets_check2" CHECK ((type = 1) = (flake IS NOT NULL))

Well, in my case (a brand new hydra database) I had jobsets_check and jobsets_check1 for the last two checks, so this is indeed worrying. What should the official migration constraint contain then?

grahamc:
I see that difference, too. Ouch.

Maybe we should:

1. create ` "jobsets_schedulingshares_check" CHECK (schedulingshares > 0)` if it doesn't exist yet
2. create your two new constraints with good names
3. drop jobsets_check, jobsets_check1, and drop jobsets_check2 if it exists

Then we don't at any point lack the constraint. Does this seem reasonable?

immae:
As said in IRC:
The plan of action looks all fine to me, but it assumes that all the hydra will have more or less the same checks names (jobsets_check, jobsets_check1, and drop jobsets_check2), which we know is false (in my local brand new one it was jobsets_schedulingshares_check, jobsets_check, jobsets_check1).

Good point: it doesn’t matter to psql if two checks are identical, as long as they have a different name. So it doesn’t really matter if there are lingering checks at the end (the two checks I’m adding in this PR are strictly more constrained than the existing ones)

Bad point: lingering check is not nice.

There is a plan of action that would avoid that:
- unconditionnally delete all constraints on the table.
- recreate the constraints.

This would leave us with an unconstrained table for a few milliseconds between the two queries (seems like it’s not a real problem, the checks are just "safety checks").

But the drawback is that the "unconditionnally delete all constraints on the table" is a complicated one:
```
DO $body$
DECLARE r record;
BEGIN
    FOR r IN SELECT c.conname AS constraint
             FROM pg_catalog.pg_constraint c
             WHERE c.conrelid = 'jobsets'::regclass::oid AND c.contype = 'c'
    LOOP     
       EXECUTE 'ALTER TABLE jobsets DROP CONSTRAINT '|| quote_ident(r.constraint) || ';';
    END LOOP;
END 
$body$;
```
Maybe not worth the trouble?

grahamc:
I checked with #postgresql on Freenode about how the rules are named, and I think this is tractable and easily solvable. They did tell me that we should really have named them in the first place, and only told me the rule after I explained my regret and am trying to fix the problem. Anywho:

* if the check constraint refers to exactly one column, the name is tablename_columnname_checkN  where N is either nothing or a number to make the result unique; the part before the N is trimmed down to size if need be
* if it refers to more than one column, then the _columnname part is omitted

This is from RhodiumToad.

This explains the two variants:

* Users who ran the migration got an explicitly named rule: `jobsets_check`.
* Users who did not run the migration got a generated rule name: `jobsets_schedulingshares_check`.


I think that makes this plan reasonable in all cases:

>    create "jobsets_schedulingshares_check" CHECK (schedulingshares > 0) if it doesn't exist yet
>    create your two new constraints with good names
>    drop jobsets_check, jobsets_check1, and drop jobsets_check2 if it exists


... and we should also explicitly name the checks in the hydra.sql...

it is a bit late here... does that check out?

immae:
Thanks @grahamc for your investigation! The rule is clear and simple, so I’m now confident that your plan is valid :)
I’ll work on that!

immae:
Ok so partial failure so far:
There is an `drop constraint if exists` but there is no `add constraint if not exist`.

I’ll need to write a more complicated query to implement that, but that will be for later today

src/sql/hydra.sql
```diff
@@ -84,9 +84,13 @@ create table Jobsets (
     startTime     integer, -- if jobset is currently running
     type          integer not null default 0, -- 0 == legacy, 1 == flake
     flake         text,
-    check (schedulingShares > 0),
-    check ((type = 0) = (nixExprInput is not null and nixExprPath is not null)),
-    check ((type = 1) = (flake is not null)),
+    constraint jobsets_schedulingshares_nonzero_check check (schedulingShares > 1),
```
cole-h, line 87: Should this be `schedulingShares > 0`?

immae: I made a copy-paste :thinking: good catch!

immae: Fixed thanks!

grahamc:
I was working through the constraints you propose here, and they were really sending my brain for a bit of a loop. To let me play with it and get a feel for it, I wrote a little Python program to explore the entire space. It suggests we could use these constraints:

```sql
            constraint jobsets_type_known_check   check (type = 0 or type = 1),
            constraint jobsets_legacy_paths_check check ((type = 0) = (nixExprInput is not null and nixExprPath is not null and flake is     null)),
            constraint jobsets_flake_paths_check  check ((type = 1) = (nixExprInput is     null and nixExprPath is     null and flake is not null))
```

(the `known` check was just to play further ...) Given these constraints, only these inputs were acceptable:

```
{'flake': None,
 'jobset_type': 0,
 'nixexprinput': 'Something',
 'nixexprpath': 'Something'}
{'flake': 'Something',
 'jobset_type': 1,
 'nixexprinput': None,
 'nixexprpath': None}
```

you can try it yourself:

```python
#!/usr/bin/env nix-shell
#!nix-shell -i python3 -p python3 python3.pkgs.psycopg2

from pprint import pprint
from typing import TypedDict, List, Optional
import psycopg2

class Case(TypedDict):
    jobset_type: int
    nixexprinput: Optional[str]
    nixexprpath: Optional[str]
    flake: Optional[str]

jobset_types = [0, 1, 2]
nixexprinputs = [ None, "Something" ]
nixexprpaths = [ None, "Something" ]
flakes = [ None, "Something" ]


cases: List[Case] = []
for jobset_type in jobset_types:
    for nixexprinput in nixexprinputs:
        for nixexprpath in nixexprpaths:
            for flake in flakes:
                cases.append({
                    'jobset_type': jobset_type,
                    'nixexprinput': nixexprinput,
                    'nixexprpath': nixexprpath,
                    'flake': flake,
                })

print(f"Exploring {len(cases)} cases... valid cases:")

conn = psycopg2.connect("host=... dbname=... user=...")

with conn:
    with conn.cursor() as cur:
        cur.execute("DROP TABLE IF EXISTS constraint_check")

        cur.execute("""
        create table constraint_check (
            type          integer not null default 0, -- 0 == legacy, 1 == flake
            nixExprInput  text, -- name of the jobsetInput containing the Nix or Guix expression
            nixExprPath   text, -- relative path of the Nix or Guix expression
            flake         text,

            constraint jobsets_type_known_check   check (type = 0 or type = 1),
            constraint jobsets_legacy_paths_check check ((type = 0) = (nixExprInput is not null and nixExprPath is not null and flake is     null)),
            constraint jobsets_flake_paths_check  check ((type = 1) = (nixExprInput is     null and nixExprPath is     null and flake is not null))
        )
        """)

    conn.commit()



i = 0
for case in cases:
    try:
        with conn:
            with conn.cursor() as ins:
                ins.execute(
                    """
                        INSERT INTO constraint_check (type, nixexprinput, nixexprpath, flake)
                        VALUES (%(jobset_type)s, %(nixexprinput)s, %(nixexprpath)s, %(flake)s)
                    """,
                    case
                )
        pprint(case)
    except psycopg2.errors.CheckViolation:
        pass
```

What do you think about this?

grahamc (approved):
Okay, good news, against the production dataset this migration takes essentially no time:

```
2021-02-03 21:19:40.198 UTC [[unknown]/25483] LOG:  statement: begin
2021-02-03 21:19:40.202 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets add constraint jobsets_type_known_check   check (type = 0 or type = 1)
2021-02-03 21:19:40.212 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets add constraint jobsets_legacy_paths_check check ((type = 0) = (nixExprInput is not null and nixExprPath is not null and flake is     null))
2021-02-03 21:19:40.216 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets add constraint jobsets_flake_paths_check  check ((type = 1) = (nixExprInput is     null and nixExprPath is     null and flake is not null))
2021-02-03 21:19:40.377 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets add constraint jobsets_schedulingshares_nonzero_check check (schedulingShares > 0)
2021-02-03 21:19:40.386 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets drop constraint if exists jobsets_schedulingshares_check
2021-02-03 21:19:40.390 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets drop constraint if exists jobsets_check
2021-02-03 21:19:40.397 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets drop constraint if exists jobsets_check1
2021-02-03 21:19:40.401 UTC [[unknown]/25483] LOG:  statement: alter table Jobsets drop constraint if exists jobsets_check2
2021-02-03 21:19:40.407 UTC [[unknown]/25483] LOG:  execute <unnamed>: UPDATE schemaversion SET version = $1
2021-02-03 21:19:40.407 UTC [[unknown]/25483] DETAIL:  parameters: $1 = '74'
2021-02-03 21:19:40.410 UTC [[unknown]/25483] LOG:  statement: commit
```

and the resulting table looks right:

```
Check constraints:
    "jobsets_flake_paths_check" CHECK ((type = 1) = (nixexprinput IS NULL AND nixexprpath IS NULL AND flake IS NOT NULL))
    "jobsets_legacy_paths_check" CHECK ((type = 0) = (nixexprinput IS NOT NULL AND nixexprpath IS NOT NULL AND flake IS NULL))
    "jobsets_schedulingshares_nonzero_check" CHECK (schedulingshares > 0)
    "jobsets_type_known_check" CHECK (type = 0 OR type = 1)
```

I reviewed the patch once more to compare the hydra.sql patch to the migrations to double-check they're equivalent. Thank you so much for sticking through the (multiple) rounds of review. Also, your commit message looks good too.

Thank you!

PR #859, opened 2021-02-01
Make github auth via files possible

pingiun:
Note: this is untested


grahamc (commented):
It looks pretty good. It'd be good to abstract out the common behavior in to a function, write a test for it, and convert the docs to markdown.

PR #860, opened 2021-02-02
Move evaluation errors from evaluations to EvaluationErrors, a new table

grahamc:
BIx likes to eagerly select all columns without a way to really tell
it so. Therefore, this splits this one large column in to its own
table.

I'd also like to make "jobsets" use this table too, but that is on hold
to stop the bleeding caused by the extreme amount of traffic this is
causing.

I checked this by running an evaluation locally and ensuring the evaluation's errors were recorded. I also checked by fetching a recent copy of prod's database and running the migration and verifying the evaluation data was correctly migrated.

### DIFF
--- a/src/lib/Hydra/Helper/Nix.pm
+++ b/src/lib/Hydra/Helper/Nix.pm
@@ -211,8 +211,8 @@ sub getEvals {
 
     my @evals = $evals->search(
         { hasnewbuilds => 1 },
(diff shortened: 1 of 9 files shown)


grahamc:
I've heard that @edolstra is super busy right now. Since this is a fairly low impact PR even if we have to migrate away from it, I'm going to go ahead and merge / deploy it within the hour.

PR #866, opened 2021-02-05
Fix Github status plugin for flakes

Infinisil:
Without this, only non-flake jobsets are sent GitHub status updates.

This is implemented by checking whether the flake is of the form `github:$owner/$repo/$rev`, and sending the status update to that repository. Note that the `$rev` is automatically determined, so jobsets with just the flake `github:$owner/$repo` work too.

The `githubstatus->inputs` configuration isn't used in that case.

I have tested this with both a flake jobset and a legacy jobset, both of which successfully update GitHub statuses now.


Infinisil:
Unfortunately the github status plugin (and I imagine others too) are still rather poor. In particular, they don't show a started or failed evaluation, and they also don't show successful jobs if they have been built before already.

PR #867, opened 2021-02-07
Fix login if Hydra runs behind HTTP proxy with sub-path location

ck3d:
The used HTTP proxy is nginx.

### DIFF
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -76,8 +76,8 @@ sub begin :Private {
 
     # XSRF protection: require POST requests to have the same origin.
     if ($c->req->method eq "POST" && $c->req->path ne "api/push-github") {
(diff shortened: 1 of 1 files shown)


grahamc:
Can you explain a little bit about why the old behavior was buggy, and why the new behavior is still correct in the cases where the old one was also correct?

ck3d:
When Hydra location is not at root URL path, a successful login is redirected wrongly to /. We applied this patch to fix this issue.

ck3d:
HTTP referer contains the URL path and oridin does not. Therefor referer is the better default. By the way, the variable is named referer.
see also:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer

markuskowa:
Any objection to merging. I am having the same problem. This patch would fix it.

PR #868, opened 2021-02-07
GitInput: Avoid fetch if rev is available

ck3d:
Similar to libfetcher in Nix, avoid fetching if rev is already
available.


markuskowa (approved):
This is a great feature. I am using this patch for the last 2 years. Works well.

PR #869, opened 2021-02-09
Add Queue Runner Status to the topbar

andir:
I've been searching for this waaay too often in the past and I simply do not see a reason not to include it in the topbar by default.

NOTE: I've not yet tested this on an actual install.


edolstra:
/queue-runner-status is a quick hack, just a dump of a raw JSON structure. It's not really something intended to be exposed in a "proper" web interface.

grahamc:
It is sort of the best we've got right now :sweat_smile:

andir:
What would a proper UI for that information look like? A table per system type, machine, ... displaying the information from the JSON?

There have been many occasions throughout the recent years where someone in #nixos-dev was looking for that particular URL to see what is going on. It always comes down to someone remembering that it exists. Having it in the navigation at least makes it discoverable even if it isn't pretty.

dasJ:
Well we could throw https://highlightjs.org/ at it. That would make it less of "just a dump" and more of a real UI experience

grahamc:
:shrug: https://codepen.io/gchristensen/pen/LYxdEew

Issue #865, opened 2021-02-05
Explicit build products not picked up if out of NAR


ctheune:
I've been debugging an issue with the netboot job (both on hydra.nixos.org and hydra.flyingcircus.io) where the build products are declared in `nix-support/hydra-build-products` but do not show up in hydra.

I stumbled over PR #827 but @ajs124 says his problem seems to be that the hydra-build-products file is empty. In my case the hydra-build-products file correctly there (I also don't have remote builders involved) and filled.

The issue specifically arises from https://github.com/NixOS/hydra/commit/5b4df3ad5a2b318a29b6bebc3e6be353b8c61d65#diff-6ae5dd4d7cad7e6c0000230c3c51eb5b9bdc92595ccf90d81e6e72544eefe18cR114 where before the change any file available from the store (and AFAICT references to a remote store would have been fetched at that point already).

While I was trying to create a patch that fixes this (by rolling back some of the changes) I stumbled over this comment:

```
            /* Ensure that the path exists and points into the Nix
               store. */
            // FIXME: should we disallow products referring to other
            // store paths, or that are outside the input closure?
```

My perspective: yes, we want to be able to reference store paths from the input closure (but I'd also agree that we don't want to refer to things outside the input closure) because otherwise synthetic builds that simply are there to reference other outputs will need to copy things and those could be quite large (netboot in my case requires around 550MiB) and I'd prefer if we didn't have to explicitly copy things around.

So, my patch (https://github.com/NixOS/hydra/compare/master...flyingcircusio:build-products-missing-external-references?expand=1
) seems to likely be a step in the wrong (long term) direction as @edolstra wanted to get rid of those fs-accessor-based calls ... so either this means we have to retrieve NARs that we refer to in the build products through the same API or extend the store API so we can get access to the hashes (I guess the NARs would need to be transferred in any case so that the generated URLs in hydra to those products will be reachable).

Anyway, the patch could be a stopgap until we resolve the discussion from the comment (and the NAR change seems to have accidentally preempted that discussion) ... 

The specific patch doesn't build cleanly yet, but I wanted to open up the discussion here as I have to leave for the week and can continue working on this next week if I get a bit of direction from @edolstra or maybe @grahamc ...

ctheune:
Gnah. I screwed up my patch. Will provide a fixed version in a minute.

ctheune:
Ok, the patch now builds cleanly. I can try working with it next week on our instance and maybe get some feedback here until then ...

[...]

PR #870, opened 2021-02-09
WIP: Fix showing build products for out-of-nar references

ctheune:
This is a patch for issue #865 and PR https://github.com/NixOS/hydra/pull/827 (we might want to include the test case from there but it might not fit exactly).

I'm not sure whether this change fits with @edolstra's intention about streaming the NAR vs. accessing files on the (local?) store ... 

Nevertheless, it works for our hydra and passes the existing tests.

### DIFF
--- a/src/hydra-queue-runner/build-result.cc
+++ b/src/hydra-queue-runner/build-result.cc
@@ -69,6 +69,9 @@ BuildOutput getBuildOutput(
 
         explicitProducts = true;
 
+
+        auto accessor = store->getFSAccessor();
+
         for (auto & line : tokenizeString<Strings>(productsFile->second.contents.value(), "\n")) {
             BuildProduct product;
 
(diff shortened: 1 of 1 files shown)


edolstra:
Unfortunately I think this breaks the streaming behaviour introduced in 5b4df3ad5a2b318a29b6bebc3e6be353b8c61d65, by requiring the entire NAR to be read back from the destination store.

ctheune:
I kind of expected that I'd be screwing something up (looks like we don't need the streaming behaviour on our instance) because I was just hunting down where it broke but don't have enough knowledge around the rest of the infrastructure.

@edolstra whats a recommended path here? Something that I could also live with would be along the lines of:

a) you need to do X to get proper build products instead of what worked before, and 
b) make hydra error out on the build if the build products do not work properly

I experienced it multiple times that the build products (which external processes rely on being able to acquire) were not properly integrated in hydra without any indication of an error. Yesterday I spent 3 hours debugging a partially working build product because trailing whitespace in build product entries are treated as part of the filename ... -_-

So if you give me some guidance on a) and b) I'd be more than willing to a) change the netboot code to work in the new situation and b) implement better error handling in hydra so it gets more reliable on this end.

PR #872, opened 2021-02-19
Use `yath` for testing Hydra

grahamc:
The second commit can be reverted once https://github.com/NixOS/nixpkgs/pull/113702 is released in a channel.

The goal with this is to have parallel test runs and bring some more modern test tooling to the table.


grahamc:
To do:

- [x] Run each test with its own DB -- https://github.com/grahamc/hydra/pull/2
- [x] Run each test with its own Nix store -- https://github.com/grahamc/hydra/pull/2
- [ ] Run each test with its own work tree so they don't clobber each other -- right now I have to run `(cd tests; rm bzr-checkout-repo svn-checkout-repo; rm -rf bzr-repo/ darcs-repo/ hg-repo/ svn-checkout svn-repo/)` fairly often.
- [ ] Break each test in to its own individual test so they can parallelize
- [ ] Support running tests with just `yath test` (or something similarly easy) -- right now you have to use the makefile or `HYDRA_HOME=../src PERL5LIB=../:../src/lib:../tests/lib:$PERL5LIB yath test -v --default-search ./`.

pingiun (commented):
I think these tests run on nix-shell? That seemed to work well. I'm not familiar with perl testing, but saw your tweet and wanted to help. I think it might be nice to add information on how to run the tests manually

grahamc:
Right now, with this PR, the only way to really run the tests easily is with:

./bootstrap
configurePhase
make check
```

soon, you'll be able to run `yath test` or `yath test ./path/to/specific/test.t`

pingiun:
I think it's nice to add information on how to run tests to the README, but if the command is going to change soon, it can be added then

grahamc:
I've updated the docs to include how to run tests with make check. Might as well include it now :).

pingiun (approved):

PR #873, opened 2021-02-23
Fix build with the latest Nix master

regnat:
Fix the build to adjust to some breaking changes in Nix master:

- string2Int now has a slightly different signature
- The printHelp function doesn't exist anymore.

Note that this removes the --help of hydra-eval-jobs.
That doesn't seem too bad given that most hydra commands don't have such a flag, but ideally we should probably migrate it to use the same autogenerated manpages as Nix now uses.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,32 +1,15 @@
 {
   "nodes": {
-    "lowdown-src": {
-      "flake": false,
-      "locked": {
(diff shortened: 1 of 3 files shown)


regnat:
Looks like this actually just got fixed by a direct push yesterday

PR #875, opened 2021-02-23
Allow building content-addressed derivations with hydra

### LINKED ISSUE #838: Add support for CA derivations
Not clear yet how much of an impact this will have on the Hydra code base. The main consequence of CA derivations is that we don't know output paths in advance, but maybe that doesn't matter much for Hydra.

thufschmitt:
Fix #838

There should be a couple of UI artifacts (because the output paths were showing up in a couple places but we don't know them anymore), but everything seems to be working fine otherwise (I managed to build a couple generations of the `nixpkgs-minimal` jobset with this).

~~Depends on #874 (could probably be made to work without it, but I built it on top of it and the effectively touch the same code)~~

~~Also depends on NixOS/nix#4477~~

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -507,12 +507,14 @@ void State::buildRemote(ref<Store> destStore,
           for (auto & [outputId, realisation] : builtOutputs) {
               // Register the resolved drv output
               localStore->registerDrvOutput(realisation);
+              destStore->registerDrvOutput(realisation);
 
               // Also register the unresolved one
               auto unresolvedRealisation = realisation;
               unresolvedRealisation.signatures.clear();
               unresolvedRealisation.id.drvHash = outputHashes.at(outputId.outputName);
               localStore->registerDrvOutput(unresolvedRealisation);
+              destStore->registerDrvOutput(unresolvedRealisation);
           }
         }
 
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -243,12 +243,20 @@ void State::buildRemote(ref<Store> destStore,
            outputs of the input derivations. */
         updateStep(ssSendingInputs);
 
-        StorePathSet inputs;
         BasicDerivation basicDrv;
         auto outputHashes = staticOutputHashes(*localStore, *step->drv);
-        if (auto maybeBasicDrv = step->drv->tryResolve(*localStore))
+        if ((derivationIsCA(step->drv->type())
+             || step->drv->type() == DerivationType::DeferredInputAddressed)) {
+            auto maybeBasicDrv = step->drv->tryResolve(*localStore);
+            if (!maybeBasicDrv)
+                throw Error(
+                    "the derivation '%s' can’t be resolved. It’s probably "
+                    "missing some outputs",
+                    localStore->printStorePath(step->drvPath));
             basicDrv = *maybeBasicDrv;
-        else {
+        } else {
+            // If the derivation is a real `InputAddressed` derivation, we must
+            // resolve it manually to keep the original output paths
             basicDrv = BasicDerivation(*step->drv);
             for (auto & input : step->drv->inputDrvs) {
               auto drv2 = localStore->readDerivation(input.first);
@@ -257,15 +265,11 @@ void State::buildRemote(ref<Store> destStore,
                 if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
                   auto inputRealisation = localStore->queryRealisation(DrvOutput{hashes.at(name), name});
                   assert(inputRealisation);
-                  inputs.insert(inputRealisation->outPath);
                   basicDrv.inputSrcs.insert(inputRealisation->outPath);
                 }
               }
             }
         }
-
-        for (auto & p : step->drv->inputSrcs)
-            inputs.insert(p);
 
         /* Ensure that the inputs exist in the destination store. This is
            a no-op for regular stores, but for the binary cache store,
@@ -287,7 +291,7 @@ void State::buildRemote(ref<Store> destStore,
 
             auto now1 = std::chrono::steady_clock::now();
 
-            copyClosureTo(machine->state->sendLock, destStore, from, to, inputs, true);
+            copyClosureTo(machine->state->sendLock, destStore, from, to, basicDrv.inputSrcs, true);
 
             auto now2 = std::chrono::steady_clock::now();
 
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -259,13 +259,10 @@ void State::buildRemote(ref<Store> destStore,
             basicDrv = BasicDerivation(*step->drv);
             for (auto & input : step->drv->inputDrvs) {
               auto drv2 = localStore->readDerivation(input.first);
-              auto hashes = staticOutputHashes(*localStore, drv2);
+              auto drv2Outputs = drv2.outputsAndOptPaths(*localStore);
               for (auto & name : input.second) {
-                if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
-                  auto inputRealisation = localStore->queryRealisation(DrvOutput{hashes.at(name), name});
-                  assert(inputRealisation);
-                  basicDrv.inputSrcs.insert(inputRealisation->outPath);
-                }
+                auto inputPath = drv2Outputs.at(name);
+                basicDrv.inputSrcs.insert(*inputPath.second);
               }
             }
         }
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -245,6 +245,7 @@ void State::buildRemote(ref<Store> destStore,
 
         StorePathSet inputs;
         BasicDerivation basicDrv;
+        auto outputHashes = staticOutputHashes(*localStore, *step->drv);
         if (auto maybeBasicDrv = step->drv->tryResolve(*destStore))
             basicDrv = *maybeBasicDrv;
         else {
@@ -413,7 +414,6 @@ void State::buildRemote(ref<Store> destStore,
                 step->drv->type() != DerivationType::DeferredInputAddressed
             );
             auto outputMap = localStore->queryPartialDerivationOutputMap(step->drvPath);
-            auto outputHashes = staticOutputHashes(*localStore, *step->drv);
             for (auto & [outputName, outputPath] : outputMap)
               if (outputPath) {
                 auto outputHash = outputHashes.at(outputName);
@@ -504,8 +504,15 @@ void State::buildRemote(ref<Store> destStore,
 
         /* Register the outputs of the newly built drv */
         if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
-          for (auto & [_, realisation] : builtOutputs) {
+          for (auto & [outputId, realisation] : builtOutputs) {
+              // Register the resolved drv output
               localStore->registerDrvOutput(realisation);
+
+              // Also register the unresolved one
+              auto unresolvedRealisation = realisation;
+              unresolvedRealisation.signatures.clear();
+              unresolvedRealisation.id.drvHash = outputHashes.at(outputId.outputName);
+              localStore->registerDrvOutput(unresolvedRealisation);
           }
         }
 
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -213,7 +213,7 @@ void State::buildRemote(ref<Store> destStore,
         unsigned int remoteVersion;
 
         try {
-            to << SERVE_MAGIC_1 << 0x204;
+            to << SERVE_MAGIC_1 << 0x205;
             to.flush();
 
             unsigned int magic = readInt(from);
@@ -244,21 +244,27 @@ void State::buildRemote(ref<Store> destStore,
         updateStep(ssSendingInputs);
 
         StorePathSet inputs;
-        BasicDerivation basicDrv(*step->drv);
+        BasicDerivation basicDrv;
+        if (auto maybeBasicDrv = step->drv->tryResolve(*destStore))
+            basicDrv = *maybeBasicDrv;
+        else {
+            basicDrv = BasicDerivation(*step->drv);
+            for (auto & input : step->drv->inputDrvs) {
+              auto drv2 = localStore->readDerivation(input.first);
+              auto hashes = staticOutputHashes(*localStore, drv2);
+              for (auto & name : input.second) {
+                if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+                  auto inputRealisation = destStore->queryRealisation(DrvOutput{hashes.at(name), name});
+                  assert(inputRealisation);
+                  inputs.insert(inputRealisation->outPath);
+                  basicDrv.inputSrcs.insert(inputRealisation->outPath);
+                }
+              }
+            }
+        }
 
         for (auto & p : step->drv->inputSrcs)
             inputs.insert(p);
-
-        for (auto & input : step->drv->inputDrvs) {
-            auto drv2 = localStore->readDerivation(input.first);
-            for (auto & name : input.second) {
-                if (auto i = get(drv2.outputs, name)) {
-                    auto outPath = i->path(*localStore, drv2.name, name);
-                    inputs.insert(*outPath);
-                    basicDrv.inputSrcs.insert(*outPath);
-                }
-            }
-        }
 
         /* Ensure that the inputs exist in the destination store. This is
            a no-op for regular stores, but for the binary cache store,
@@ -398,6 +404,27 @@ void State::buildRemote(ref<Store> destStore,
             result.logFile = "";
         }
 
+        DrvOutputs builtOutputs;
+        if (GET_PROTOCOL_MINOR(remoteVersion) >= 6) {
+            builtOutputs = worker_proto::read(*localStore, from, Phantom<DrvOutputs>{});
+        } else {
+            assert(
+                step->drv->type() != DerivationType::CAFloating &&
+                step->drv->type() != DerivationType::DeferredInputAddressed
+            );
+            auto outputMap = localStore->queryPartialDerivationOutputMap(step->drvPath);
+            auto outputHashes = staticOutputHashes(*localStore, *step->drv);
+            for (auto & [outputName, outputPath] : outputMap)
+              if (outputPath) {
+                auto outputHash = outputHashes.at(outputName);
+                auto drvOutput = DrvOutput{outputHash, outputName};
+                builtOutputs.insert({drvOutput, Realisation{drvOutput, *outputPath}});
+              }
+        }
+        StorePathSet outputs;
+        for (auto & [_, realisation] : builtOutputs)
+            outputs.insert(realisation.outPath);
+
         /* Copy the output paths. */
         if (!machine->isLocalhost() || localStore != std::shared_ptr<Store>(destStore)) {
             updateStep(ssReceivingOutputs);
@@ -405,12 +432,6 @@ void State::buildRemote(ref<Store> destStore,
             MaintainCount<counter> mc(nrStepsCopyingFrom);
 
             auto now1 = std::chrono::steady_clock::now();
-
-            StorePathSet outputs;
-            for (auto & i : step->drv->outputsAndOptPaths(*localStore)) {
-                if (i.second.second)
-                   outputs.insert(*i.second.second);
-            }
 
             /* Get info about each output path. */
             std::map<StorePath, ValidPathInfo> infos;
@@ -479,6 +500,13 @@ void State::buildRemote(ref<Store> destStore,
             auto now2 = std::chrono::steady_clock::now();
 
             result.overhead += std::chrono::duration_cast<std::chrono::milliseconds>(now2 - now1).count();
+        }
+
+        /* Register the outputs of the newly built drv */
+        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+          for (auto & [_, realisation] : builtOutputs) {
+              localStore->registerDrvOutput(realisation);
+          }
         }
 
         /* Shut down the connection. */
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -246,7 +246,7 @@ void State::buildRemote(ref<Store> destStore,
         StorePathSet inputs;
         BasicDerivation basicDrv;
         auto outputHashes = staticOutputHashes(*localStore, *step->drv);
-        if (auto maybeBasicDrv = step->drv->tryResolve(*destStore))
+        if (auto maybeBasicDrv = step->drv->tryResolve(*localStore))
             basicDrv = *maybeBasicDrv;
         else {
             basicDrv = BasicDerivation(*step->drv);
@@ -255,7 +255,7 @@ void State::buildRemote(ref<Store> destStore,
               auto hashes = staticOutputHashes(*localStore, drv2);
               for (auto & name : input.second) {
                 if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
-                  auto inputRealisation = destStore->queryRealisation(DrvOutput{hashes.at(name), name});
+                  auto inputRealisation = localStore->queryRealisation(DrvOutput{hashes.at(name), name});
                   assert(inputRealisation);
                   inputs.insert(inputRealisation->outPath);
                   basicDrv.inputSrcs.insert(inputRealisation->outPath);
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -528,7 +528,7 @@ void State::buildRemote(ref<Store> destStore,
         }
 
         /* Register the outputs of the newly built drv */
-        if (settings.isExperimentalFeatureEnabled("ca-derivations")) {
+        if (settings.isExperimentalFeatureEnabled(Xp::CaDerivations)) {
           auto outputHashes = staticOutputHashes(*localStore, *step->drv);
           for (auto & [outputId, realisation] : builtOutputs) {
               // Register the resolved drv output
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -153,6 +153,71 @@ StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo> & paths
     return sorted;
 }
 
+/**
+ * Replace the input derivations by their output paths to send a minimal closure
+ * to the builder.
+ *
+ * If we can afford it, resolve it, so that the newly generated derivation still
+ * has some sensible output paths.
+ */
+BasicDerivation inlineInputDerivations(Store & store, Derivation & drv, const StorePath & drvPath)
+{
+    BasicDerivation ret;
+    auto outputHashes = staticOutputHashes(store, drv);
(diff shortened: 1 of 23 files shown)


src/lib/Hydra/Controller/Build.pm
```diff
@@ -66,10 +66,10 @@ sub build_GET {
 
     $c->stash->{template} = 'build.tt';
     $c->stash->{isLocalStore} = isLocalStore();
-    $c->stash->{available} =
-        $c->stash->{isLocalStore}
-        ? all { isValidPath($_->path) } $build->buildoutputs->all
-        : 1;
+    $c->stash->{available} = 1; # XXX: obviously not always true
+        # $c->stash->{isLocalStore}
```
SuperSandro2000, line 70: Is that left from debugging?

tomberek (approved):
It works!

t/content-addressed.t
```diff
@@ -0,0 +1,35 @@
...
+
+ok(evalSucceeds($jobset),                  "Evaluating jobs/content-addressed.nix should exit with return code 0");
+ok(nrQueuedBuildsForJobset($jobset) == 3 , "Evaluating jobs/content-addressed.nix should result in 3 builds");
+
+for my $build (queuedBuildsForJobset($jobset)) {
+    ok(runBuild($build), "Build '".$build->job."' from jobs/content-addressed.nix should exit with code 0");
+    my $newbuild = $db->resultset('Builds')->find($build->id);
+    my $expected = $build->job eq "fails" ? 1 : $build->job =~ /with_failed/ ? 6 : 0;
+    ok($newbuild->finished == 1 && $newbuild->buildstatus == $expected, "Build '".$build->job."' from jobs/content-addressed.nix should have buildstatus $expected");
+}
```
grahamc, line 31: Check out the other test files: there are other helpers like `is` which, when the test fails, prints the actual/expected values. Also, please don't check two facts at once (ie: don't use `&&` in a check.)

t/jobs/content-addressed.nix
```diff
@@ -0,0 +1,28 @@
+let cfg =  import ./config.nix; in
+let
+  mkDerivation = args: cfg.mkDerivation ({
+    __contentAddressed = true;
+    outputHashMode = "recursive";
+    outputHashAlgo = "sha256";
+  } // args);
```
grahamc, line 7: Feel free to extend config.nix.in (and regenerate with make) with more functions, like makeContetAddressedDerivation :)

src/lib/Hydra/Controller/Build.pm
```diff
@@ -66,9 +66,11 @@ sub build_GET {
 
     $c->stash->{template} = 'build.tt';
     $c->stash->{isLocalStore} = isLocalStore();
+    # XXX: If ca-derivations is enabled then this will always return false
+    # because `$_->path` will be empty
```
grahamc, line 70: Is this always true, or is it true if it is set AND the derivation is ca-addressed? This feels like a fairly spooky change to the data model of Hydra.

grahamc:
I should have done this as a review, sorry.

In the C++ I'm seeing a lot of in-place complication. Would it be possible to clean up the campsite a bit, and make some functions and break up the behavior a bit? I'm increasingly uncertain about understanding those pieces of the code.

grahamc:
I also wonder if there would be anything interesting in tests for:

1. requesting the Builds page. It is fairly easy to test controllers using https://github.com/NixOS/hydra/blob/master/t/Controller/Jobset/channel.t#L15-L16 :` https://github.com/NixOS/hydra/blob/master/t/Controller/Jobset/channel.t#L36
2. what happens when trying to build a content-addressed derivation but CA is not enabled?
3. what happens when building a CA derivation but its already been built and pushed to a binary cache? Here is a test that checks similar behavior: https://github.com/NixOS/hydra/blob/master/t/queue-runner/notifications.t
4. how does Hydra handle it if Hydra is ca-enabled but its builders are not?

src/lib/Hydra/Controller/Build.pm
regnat: Mh… I see what you mean. It's only true if the derivation is ca. I've updated the comment in ca86783e9d9c5e6fe744bfd33dc9846122203935 to make it clearer.

> This feels like a fairly spooky change to the data model of Hydra.

It is indeed :smiley: . But the `path` must be nullable, because of the very nature of CA derivations

t/jobs/content-addressed.nix
regnat: Right. Done in 8ef36cc7cf95cdf5b1ff7b629b9dc0f498f72548

t/content-addressed.t
regnat: Oh indeed. This test was actually a copy/paste of the “basic” test, but it didn't follow the overhaul of the test infrastructure.
Fixed in a12704f38f57a0c8a84773f917c612fc133caea7

regnat:
Thanks for the review @grahamc 

> In the C++ I'm seeing a lot of in-place complication. Would it be possible to clean up the campsite a bit, and make some functions and break up the behavior a bit? I'm increasingly uncertain about understanding those pieces of the code.

A large part of the complication is annoyingly due to the fact that we need to support both the nix+ca-derivations and nix-without-ca-derivations cases, which means two slightly different code paths. But it should be indeed possible to factor at least part of it out.

> I also wonder if there would be anything interesting in tests for:
>
>    requesting the Builds page. It is fairly easy to test controllers using https://github.com/NixOS/hydra/blob/master/t/Controller/Jobset/channel.t#L15-L16 :` https://github.com/NixOS/hydra/blob/master/t/Controller/Jobset/channel.t#L36
>    what happens when trying to build a content-addressed derivation but CA is not enabled?
>    what happens when building a CA derivation but its already been built and pushed to a binary cache? Here is a test that checks similar behavior: https://github.com/NixOS/hydra/blob/master/t/queue-runner/notifications.t
>    how does Hydra handle it if Hydra is ca-enabled but its builders are not?

I’d have to think it a bit more, but I think there's a number of things worth testing here indeed.

src/hydra-queue-runner/build-remote.cc
```diff
@@ -244,22 +244,29 @@ void State::buildRemote(ref<Store> destStore,
...
-        for (auto & input : step->drv->inputDrvs) {
-            auto drv2 = localStore->readDerivation(input.first);
-            for (auto & name : input.second) {
-                if (auto i = get(drv2.outputs, name)) {
-                    auto outPath = i->path(*localStore, drv2.name, name);
-                    inputs.insert(*outPath);
-                    basicDrv.inputSrcs.insert(*outPath);
+        BasicDerivation basicDrv;
+        auto outputHashes = staticOutputHashes(*localStore, *step->drv);
+        if (auto maybeBasicDrv = step->drv->tryResolve(*localStore))
```
regnat, line 249: Note to self: Reading through the code again, I'm actually wondering whether this isn't messing-up the build of non-ca derivations (because `tryResolve` changes the output paths, so applying it unconditionally might do some funny things). I'd expect this would cause a test failure (or a hydra crash), but it's worth double-checking first

regnat:
I've added a bunch of tests (and fixed a bunch of related issues by the way).

I didn't test

>     3. what happens when building a CA derivation but its already been built and pushed to a binary cache? Here is a test that checks similar behavior: https://github.com/NixOS/hydra/blob/master/t/queue-runner/notifications.t
> 
>     4. how does Hydra handle it if Hydra is ca-enabled but its builders are not?

because
- substituting ca derivations isn't supported yet in hydra (and doing so will probably require a bit of work)
- A “mixed” network of machines (some ca-enabled, some not) isn't supported by Nix atm (I've just opened https://github.com/NixOS/nix/issues/4755 to track it)

regnat:
The test hydra instance just showed some weird failures with fixed-output derivations (https://hydra.ngi0.nixos.org/build/171 − the cbindgen vendor derivation should build just fine but is marked as failed without anything appearing in the log). This needs a bit of investigation − Might be linked to https://github.com/NixOS/hydra/pull/875#discussion_r622156625 because maybe that bit of code also rewrites the output hashes of FO derivations.

regnat:
> The test hydra instance just showed some weird failures with fixed-output derivations (https://hydra.ngi0.nixos.org/build/171 − the cbindgen vendor derivation should build just fine but is marked as failed without anything appearing in the log). This needs a bit of investigation − Might be linked to [#875 (comment)](https://github.com/NixOS/hydra/pull/875#discussion_r622156625) because maybe that bit of code also rewrites the output hashes of FO derivations.

OK, looks like these are “legitimate” has mismatches. So the only issue is that the error isn’t properly reported − which I think is a Nix issue

grahamc:
It would be really great if whatever the underlying issue is has an
accompanying test wherever the fix belongs :) these sort of mystery
problems are exhausting 😔

On April 29, 2021, GitHub ***@***.***> wrote:
> > The test hydra instance just showed some weird failures with fixed-
> > output derivations (<https://hydra.ngi0.nixos.org/build/171> − the
> > cbindgen vendor derivation should build just fine but is marked as
> > failed without anything appearing in the log). This needs a bit of
> > investigation − Might be linked to #875 (comment)
> > <https://github.com/NixOS/hydra/pull/875#discussion_r622156625>
> > because maybe that bit of code also rewrites the output hashes of FO
> > derivations.
> >
>
> OK, looks like these are “legitimate” has mismatches. So the only
> issue is that the error isn’t properly reported − which I think is a
> Nix issue
>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/NixOS/hydra/pull/875#issuecomment-829117758>, or
unsubscribe <https://github.com/notifications/unsubscribe-
auth/AAASXLE7UTSXLSFPVILDEATTLEYFRANCNFSM4YCDMVTQ>.

tomberek:
Possible issue with remote builders and hydra. (note: there should be no CA builds occurring)
Using hydra:

```
 src = pkgs.fetchFromGitHub {
    owner = "regnat";
    repo = "hydra";
    rev = "2090d3a0899db4b63d3fbf93545be0020385a165";
    sha256 = "sha256-GrYR/xX/kWxxB0VoHKj0c3Pw7A6RLPexDkVNNVa7kn4=";
};
```
Getting errors like:
```
possibly transient failure building ‘/nix/store/XYZ.drv’ on ‘root@333.33.3.3’: error: dependency '/nix/store/3fs5wqlj9ipv7wjx4av81skmpcdjjh3d-jq-1.6-dev' of '/nix/store/XYZ.drv' does not exist, and substitution is disabled
```

Reverting to `package = hydra-unstable;` restores behavior as expected with regards to remote building. All builds have `__contentAddressed = false`. Any thoughts, leads? I attempted setting use-substitutes, useSubstitutes, and builders-use-substitutes to no effect. Only solution was to revert.

regnat:
> Getting errors like:
>
> ```
> possibly transient failure building ‘/nix/store/XYZ.drv’ on ‘root@333.33.3.3’: error: dependency '/nix/store/3fs5wqlj9ipv7wjx4av81skmpcdjjh3d-jq-1.6-dev' of '/nix/store/XYZ.drv' does not exist, and substitution is disabled
> ```

We’ve also started getting the same on https://hydra.ngi0.nixos.org/build/202 .

That might (quite ironically) be related to https://github.com/NixOS/hydra/pull/875/commits/d7a1b6b982bf64e560998ee6d041d0da66e518c8

Mindavi:
Running this branch now with ca-derivation builds, but getting errors in the UI like this: `Caught exception in Hydra::Controller::Build->build "[31;1merror:[0m path '[33;1m/1vbp723njr8wq6m404hzkfmqck0vk73gcgr6pjc78il65mh6p9mn[0m' is not in the Nix store at /nix/store/62znwr88yc5mi5v0qxmzp9rd84km68q6-hydra-2021-05-03/libexec/hydra/lib/Hydra/Controller/Build.pm line 73."`. Seems to be building anyway though, which wasn't the case with hydra-unstable.

Not sure if that's to be addressed in a future patch or if I hit some edge case.

regnat:
@tomberek can you try 854dcec ? I think it should fix your issue.

@Mindavi there were some known issues with earlier commits in this branch where the database wasn’t properly filled with all the required informations, causing this kind of issue (and which persisted after an upgrade because the incomplete data was still there). Any chance that’s the case for you?

Mindavi:
I guess I broke it with trying to build with ca-derivations first and only then finding out hydra doesn't support it and applying thid patchset.

It seems to be ok now though, with a fresh jobset, apart from the (already expected) ui artifacts (notably, empty package names).

Mindavi:
This has now been running pretty smoothly for a while for me. However, there's one thing that's now starting to happen:

```
    Aborted: error: Trying to register a realisation of 'sha256:5a7f41bac1b8abb186eb4f0ec1a76ee90809fabbdd620bde7ea0ddb66cb96e6a!out', but we already have another one locally.
      Local: /nix/store/2489p9qk8i8wi627maxix9w15li2y890-systemd-247.6
      Remote: /nix/store/dvbv3d914ii22vnr277bslzq8lnwz68x-systemd-247.6
```

I already have systemd built locally, but it appears to be trying to substitute it anyways? And the remote appears to have another hash as well. I'm not sure what's happening here, but it shouldn't be trying to substitute it, I guess.

This has happened for some different packages over the last few days, e.g. also for glibc.

I have the binary cache for the ngi0 machine that's building the ca derivations enabled, I guess those might also have something to do with this?

Versions:

- nix (Nix) 2.4pre20210707_02dd6bb
- hydra: d20639a322e340770c9672e0181e70e48214a21c (latest commit on this PR as of today)

Mindavi:
The issue where realisations cannot be registered might be related to `separateDebugInfo` (or more specifically, the `--build-id` flag), since I can imagine that that can cause ca derivations to be mismatched, while otherwise the inputs appear the same. Not 100% sure how it's generated though, so this may be a red herring.

systemd does not have separateDebugInfo enabled, but uses the flag for itself to build this file: `/lib/systemd/boot/efi/linuxx64.elf.stub`.

Since nix tries to rebuild even if I try to do a build manually, I don't think this is a hydra-only problem.

It seems that this issue occurs mostly by building staging regularly. I'll have to find 2 versions of nixpkgs which will clearly reproduce this to help debug this issue at some point.

Ericson2314:
Should we get this merged and working with 2.4?

Mindavi:
I've run this for a while and even though it's not perfect, it does seem quite ok. Getting this in the main codebase may ensure that it doesn't get too stale, since it now needs a rebase with lots of subtle changes (since I'm not familiar with the codebase it was too hard to get right for me). It was not stable enough for me to keep running, since it kept getting stuck due to the issue described above, where a realisation couldn't be registered. The only way I found to solve that was to do a garbage collection, which is not great.

I think getting this in may also lead to more people testing it out and ironing out the last issues (no idea how many people run hydra though). I wouldn't say that it's wise to run the whole of nixpkgs on this yet, but it is good enough for a first cut IMO.

More testers would be nice, but it doesn't seem like that's going to happen on the short term, and this will get more stale over time.

Depends on how much time the author has, of course :).

Ericson2314:
Pinging again :) Really hope we can see this merged, and then discuss QAing Nixpkgs!

t184256:
Needed [this](https://gist.github.com/t184256/f1532591033a522d4daba16aa66b5de0) with nix-2.6.0pre20220117_fc2443a.

Ericson2314:
@edolstra @thufschmitt @grahamc we should do...something about this!

Ericson2314:
Hydra now uses Nix 2.6 so this can shrink a little bit on rebase.

thufschmitt:
@aciceri that's very good news. I'll try to schedule a call with you to discuss it if that's OK so that I can give you an overview of the current state (as much as I remember, it's been a while 😬)

thufschmitt:
Closing in favor of #1228

Ericson2314:
Reopening because I can push to this one.

Ericson2314:
(I backed out of my last merge commit to see if that fixes CI, as it did fix the CA test locally)

t/jobs/dir-with-file-builder.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
+
+# Workaround for https://github.com/NixOS/nix/pull/6051
+echo "some output"
```
Mindavi, line 4: If I read upstream correctly, this workaround shouldn't be needed anymore.

SuperSandro2000:
Could it be that we are missing the remote builder check implemented here https://github.com/NixOS/nix/pull/4905/files#diff-e89a2224d1d4c39bf5a2d448b7b2fac1e3e75c7e08e414ee448846266c8bfeab ?

Because I am experiencing that hydra tries to use remote builders for ca-derivations that do not have that feature enabled in the daemon or the machines file.

SuperSandro2000:
I think https://github.com/SuperSandro2000/hydra/commit/27b1935f5df476845ec5b909d14e8c95cb79e49a solved that issue but I am not yet 100% sure.

Ericson2314:
(For anyone else reading this, the scheduling issues @SuperSandro2000 brings up are now fixed on `master`.)

Ericson2314:
Renamed because since I factored out #1349, what is left just in this PR is just UI changes (and the extra column that those UI changes leverage).

SuperSandro2000:
nix_2_21 is unhappy with this PR/hydra in generell:

```
hydra> g++ -DHAVE_CONFIG_H -I. -I../..    -std=c++2a -I/nix/store/90ns4amhyqqgb9lnx3kvqsl6vjigpzww-nix-2.21.0-dev/include/nix -I/nix/store/h5g13jk10ziz3xfnpdvgcsc0brzxmkkw-boehm-gc-8.2.4-dev/include -Wall -I ../libhydra -Wno-deprecated-declarations -g -O2 -include nix/config.h -c -o hydra_evaluator-hydra-evaluator.o `test -f 'hydra-evaluator.cc' || echo './'`hydra-evaluator.cc
hydra> hydra-evaluator.cc: In member function 'std::string JobsetId::display() const':
hydra> hydra-evaluator.cc:41:20: error: 'format' was not declared in this scope
hydra>    41 |         return str(format("%1%:%2% (jobset#%3%)") % project % jobset % id);
hydra>       |                    ^~~~~~
hydra> hydra-evaluator.cc:41:20: note: suggested alternatives:
hydra> In file included from /nix/store/7vc88ixc6yca2hwgszjk51dh23j3g2nr-gcc-13.2.0/include/c++/13.2.0/bits/chrono_io.h:39,
hydra>                  from /nix/store/7vc88ixc6yca2hwgszjk51dh23j3g2nr-gcc-13.2.0/include/c++/13.2.0/chrono:3330,
hydra>                  from /nix/store/n2v9lm5vvam9v847qvpdrjik0rk1qz02-libpqxx-7.7.5/include/pqxx/time.hxx:12,
hydra>                  from /nix/store/n2v9lm5vvam9v847qvpdrjik0rk1qz02-libpqxx-7.7.5/include/pqxx/pqxx:26,
hydra>                  from ../libhydra/db.hh:3,
hydra>                  from hydra-evaluator.cc:1:
hydra> /nix/store/7vc88ixc6yca2hwgszjk51dh23j3g2nr-gcc-13.2.0/include/c++/13.2.0/format:3764:5: note:   'std::format'
hydra>  3764 |     format(const locale& __loc, wformat_string<_Args...> __fmt,
hydra>       |     ^~~~~~
hydra> In file included from /nix/store/n2v9lm5vvam9v847qvpdrjik0rk1qz02-libpqxx-7.7.5/include/pqxx/result.hxx:26,
hydra>                  from /nix/store/n2v9lm5vvam9v847qvpdrjik0rk1qz02-libpqxx-7.7.5/include/pqxx/binarystring.hxx:22,
hydra>                  from /nix/store/n2v9lm5vvam9v847qvpdrjik0rk1qz02-libpqxx-7.7.5/include/pqxx/pqxx:5:
hydra> /nix/store/n2v9lm5vvam9v847qvpdrjik0rk1qz02-libpqxx-7.7.5/include/pqxx/types.hxx:80:12: note:   'pqxx::format'
hydra>    80 | enum class format : int
hydra>       |            ^~~~~~
hydra> In file included from include/boost/format.hpp:34,
hydra>                  from /nix/store/90ns4amhyqqgb9lnx3kvqsl6vjigpzww-nix-2.21.0-dev/include/nix/fmt.hh:4,
hydra>                  from /nix/store/90ns4amhyqqgb9lnx3kvqsl6vjigpzww-nix-2.21.0-dev/include/nix/error.hh:21,
hydra>                  from /nix/store/90ns4amhyqqgb9lnx3kvqsl6vjigpzww-nix-2.21.0-dev/include/nix/util.hh:5,
hydra>                  from ../libhydra/db.hh:6:
hydra> include/boost/format/format_fwd.hpp:27:37: note:   'boost::format'
hydra> hydra-evaluator.cc:41:16: error: 'str' was not declared in this scope
hydra>    41 |         return str(format("%1%:%2% (jobset#%3%)") % project % jobset % id);
hydra>       |                ^~~
hydra> hydra-evaluator.cc:41:16: note: suggested alternatives:
hydra> In file included from include/boost/format.hpp:53:
hydra> include/boost/format/free_funcs.hpp:22:38: note:   'boost::str'
hydra> include/boost/format/free_funcs.hpp:22:38: note:   'boost::str'
hydra> make[3]: *** [Makefile:443: hydra_evaluator-hydra-evaluator.o] Error 1
hydra> make[3]: Leaving directory '/build/source/src/hydra-evaluator'
hydra> make[2]: *** [Makefile:363: all-recursive] Error 1
hydra> make[2]: Leaving directory '/build/source/src'
hydra> make[1]: *** [Makefile:423: all-recursive] Error 1
hydra> make[1]: Leaving directory '/build/source'
hydra> make: *** [Makefile:351: all] Error 2
```

Mindavi:
I think currently (at least) a database migration is missing here.

SuperSandro2000:
> (For anyone else reading this, the scheduling issues @SuperSandro2000 brings up are now fixed on `master`.)

Not sure if I ever confirmed this, but lately I have noticed that sometimes builds couldn't be scheduled because the feature was missing, so from my perspective things seem to work as expected 🎉

gador:
When building packages and system configurations with this PR, it all seems to work with and without content-addressed derivations.
Only when I tried to build one of my system configurations, I get:
```
Nov 02 12:33:34 hel hydra-queue-runner[255887]: performing step ‘/nix/store/dxzgs08hhmcvm48pflwv1lfmimcls7gx-tmpfiles.d.drv’ 1 times on ‘localhost’ (needed by build 198 and 0 others)
Nov 02 12:33:34 hel hydra-queue-runner[255887]: warning: output out of input /nix/store/qp9sqj34pl45l65hlslz1hidjh8vb9l7-nixos-tmpfiles.d.drv missing, aborting the resolving
```
Note: The derivation `dxzgs08hhmcvm48pflwv1lfmimcls7gx-tmpfiles.d.drv` is present on the machine, as well as the build output. When I look at it, I see:
```
nix build --print-out-paths "/nix/store/dxzgs08hhmcvm48pflwv1lfmimcls7gx-tmpfiles.d.drv^*"
/nix/store/f3a9xyvjh2g98xdnfl1z01vp66r2hwrr-tmpfiles.d
ls -la /nix/store/f3a9xyvjh2g98xdnfl1z01vp66r2hwrr-tmpfiles.d
total 25064
dr-xr-xr-x     2 root root      16 Jan  1  1970 .
drwxrwxr-t 14919 root nixbld 92640 Nov  2 12:49 ..
lrwxrwxrwx     2 root root      89 Jan  1  1970 00-nixos.conf -> /nix/store/55bhwpm4zk1p25m73lbsv3bchxkc1czm-nixos-tmpfiles.d/lib/tmpfiles.d/00-nixos.conf
lrwxrwxrwx     5 root root      86 Jan  1  1970 home.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/home.conf
lrwxrwxrwx     5 root root      95 Jan  1  1970 journal-nocow.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/journal-nocow.conf
lrwxrwxrwx     3 root root      81 Jan  1  1970 lvm2.conf -> /nix/store/w0w4mcvrma1b11kv2fxi6x02dcnqxkyz-lvm2-2.03.26/lib/tmpfiles.d/lvm2.conf
lrwxrwxrwx     4 root root      85 Jan  1  1970 nix-daemon.conf -> /nix/store/s94nymfmjg1bknwfbbmz2r4k8qr7g964-nix-2.24.9/lib/tmpfiles.d/nix-daemon.conf
lrwxrwxrwx     5 root root      91 Jan  1  1970 portables.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/portables.conf
lrwxrwxrwx     5 root root     106 Jan  1  1970 static-nodes-permissions.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/static-nodes-permissions.conf
lrwxrwxrwx     5 root root      89 Jan  1  1970 systemd.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/systemd.conf
lrwxrwxrwx     5 root root      97 Jan  1  1970 systemd-nologin.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/systemd-nologin.conf
lrwxrwxrwx     5 root root      96 Jan  1  1970 systemd-nspawn.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/systemd-nspawn.conf
lrwxrwxrwx     5 root root      93 Jan  1  1970 systemd-tmp.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/systemd-tmp.conf
lrwxrwxrwx     5 root root      85 Jan  1  1970 tmp.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/tmp.conf
lrwxrwxrwx     5 root root      85 Jan  1  1970 var.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/var.conf
lrwxrwxrwx     5 root root      85 Jan  1  1970 x11.conf -> /nix/store/zqmkf68sb4j70v89j1hhsa44b0fb0sdv-systemd-256.6/example/tmpfiles.d/x11.conf
```
So the derivation, as well as the output, is present. Restarting the build (or wait for the queue-runner to do it again), results in the exact same error message. So does repeating or restarting the build. It it impossible to build the system config right now with hydra.

This happens on the `localhost` builder just as well as on remote builders. 
This only happens with content-addressed derivations and only when I build the system configuration. All other packages ( I maintain) build fine as CA's.

The only related issue I found was https://github.com/NixOS/nix/issues/6700 which did happen to me i the past, but got fixed later on. I also don't get the `error: unexpected end-of-file` message, but just the above.

The nix-daemon.service doesn't report any errors either, which is why I think it has something to do with how hydra build the CA derivations (ergo this and the related PR's)

Also, I already did build the system configuration with hydras own `reproduce locally` command just fine. Once on a remote builder, once on localhost. So this seems hydra-specific.

gador:
Small update. I have been running hydra with this PR for the past month. After I uploaded the (for some reason) broken `drv`s to the store with
```
nix copy --to "s3://nix-cache?profile=nixbuilder[...]" "/nix/store/avl0b0qvw6qh4m9padwkbawn0h4gp4n6-perl5.40.0-Test2-Harness-1.000155.drv^*"
```
everything worked just fine. I had to do this manually a few times in the beginning and everything has been running smoothly since then (with CA nixos configs, normal configs and package sets)

Maybe there was a hickup with the local store or something, I don't know.

Mindavi:
FYI: The main thing this PR does is give a better UI, all other ca-derivations support is already part of master.

PR #876, opened 2021-02-23
Create an ephemeral PostgreSQL database and Nix store per test, split up tests.

grahamc:
This allows us to run multiple tests simultaneously without interfering with each other.

Some notes:

1. the change to the db.hh is required because Test::PostgreSQL creates the DBI string with the alternate casing.
2. the changes from use to require is to avoid fixating the DBI string in the DB.pm configuration too early.

~I'll  follow up to this PR with a PR breaking apart the existing test file to reap the benefits of parallelized tests. You can see that WIP PR here: https://github.com/grahamc/hydra/pull/3~ I've included those changes in this PR as well because of the issues around cleaning up the test directory without that PR.

This adds a second or so overhead per test, but allows tests to be run individually and also in parallel. Here is a asciinema of the test runner:

https://asciinema.org/a/JrW7RAgra0dzHvsVvYw4uHowg


grahamc:
cc @pingiun if you would!

tests/tear-down.pl
```diff
@@ -1,12 +0,0 @@
-use strict;
-
-my $fail = 0;
-
-system("dropdb -p 6433 hydra-test-suite") == 0 or $fail = 1;
-system("pg_ctl -D postgres -w stop") == 0 or $fail = 1;
-
-system("chmod -R a+w nix") == 0 or $fail = 1;
-system("rm -rf postgres data nix git-repo hg-repo svn-repo svn-checkout svn-checkout-repo bzr-repo bzr-checkout-repo darcs-repo") == 0 or $fail = 1;
```
grahamc, line 9: Looks like this line has been lost without replacement.

grahamc: This has been resolved in what would be a follow-up UP https://github.com/grahamc/hydra/pull/3 by creating these repositories in a per-test scratch directory.

edolstra (approved):

src/libhydra/db.hh
```diff
@@ -13,10 +13,16 @@ struct Connection : pqxx::connection
...
-        std::string prefix = "dbi:Pg:";
-        if (std::string(s, 0, prefix.size()) != prefix)
-            throw Error("$HYDRA_DBI does not denote a PostgreSQL database");
-        return concatStringsSep(" ", tokenizeString<Strings>(string(s, prefix.size()), ";"));
+
+        std::string lower_prefix = "dbi:Pg:";
+        std::string upper_prefix = "DBI:Pg:";
+
+        if ((std::string(s, 0, lower_prefix.size()) == lower_prefix) ||
+            (std::string(s, 0, upper_prefix.size()) == upper_prefix)) {
```
edolstra, line 21:
```suggestion
        if (hasPrefix(s, lower_prefix) || hasPrefix(s, upper_prefix)) {
```

grahamc: I applied this locally and it works perfectly. Thanks, nice!

PR #877, opened 2021-02-23
Convert documentation to markdown / mdbook

immae:
projects.xml and declarative-projects.xml were merged with xmllint, and
then I ran that to convert files
```
for i in *.xml; do pandoc -s -f docbook -t markdown $i -o ${i/xml/md}; done
```

Some minor rewording needed to be done to handle the fact that mdbook doesn’t handle references.


immae:
I think the code is now ready for review. Note that I’m unfamiliar with the automake magics, so the Makefile.am might be more verbose than necessary.

The output of index.html is in the same place as before, that’s what I considered "successful"

immae:
(cc @grahamc )

grahamc:
Thank you!

edolstra:
Instead of pandoc, can we use mdbook to generate the HTML manual? pandoc is a very big dependency.

immae:
(Wed, Feb 24, 2021 at 05:04:13AM -0800) Eelco Dolstra :
> Instead of pandoc, can we use mdbook to generate the HTML manual? pandoc is a very big dependency.

It’s not a dependency, it’s just to do the initial conversion (docbook
-> md), that I described in the PR description. From now on it’s only
mdbook!

-- 
Ismael

PR #880, opened 2021-02-25
RunCommand: emit the `finished` field as a boolean

grahamc:
DRAFT: I'd like to get some feedback from hydra users first. It feels a bit rude to break the compatibility of this field, but I'd quite like to clean up the campsite a bit.


grahamc:
I guess I'll merge it :shrug:

PR #882, opened 2021-03-04
tests/input-types: split out scminputs into individual tests

cole-h:
This makes the test faster (by removing it and replacing it with a
`TestScmInput` module that exports the `testScmInput` subroutine). Now,
all the input tests can be run in parallel.

Some of the `tests/jobs/*-update.sh` scripts were "broken" (e.g. tests
failed for various reasons on my machine), so I fixed those up as well.


tests/jobs/svn-checkout-update.sh
```diff
@@ -11,6 +11,11 @@ fi
 
 case $state in
     (0) echo "::Create repo. -- continue -- updated::"
+    svnadmin create svn-repo
+    svn co file://$PWD/$repo svn-checkout
+    touch svn-checkout/svn-file
+    svn add svn-checkout/svn-file
+    svn commit -m "add svn file" svn-checkout/svn-file
```
grahamc, line 18: Is this needed because it is no longer sharing a state space with a previously ordered svn test?

cole-h: Correct (same with the `bzr-checkout-update.sh` script) -- or at least, because the `bzr-checkout-update` test can be run before `bzr` now (unsure if this is always the case).

grahamc (commented):
Can you add a comment to the top of each of these newly created files indicating what their test is supposed to be covering? Like, right below the `hydra_setup(..` call.

grahamc (approved):
Nice! Thanks!

PR #883, opened 2021-03-04
Trivially run tests with `yath test`

cole-h:
```
tests: move to t, allow `yath test` from root

By moving the tests subdirectory to t, we gain the ability to run `yath
test` with no arguments from inside `nix develop` in the root of the
the repo.

(`nix develop` is necessary in order to set the proper env vars for
`yath` to find our test libraries.)
```

---

```
doc: document how to run tests

Both `make check` and `make && yath test` are now documented ways to
run the test suite.
```

---

Depends on / incorporates https://github.com/NixOS/hydra/pull/884. Will be rebased after that is merged.


flake.nix
```diff
@@ -319,13 +319,21 @@
...
+          # YATH_JOB_COUNT = "\${NIX_BUILD_CORES-$(nproc)}";
+          YATH_JOB_COUNT = "4";
+
           shellHook = ''
+            pushd $(git rev-parse --show-toplevel) >/dev/null
+
             PATH=$(pwd)/src/hydra-evaluator:$(pwd)/src/script:$(pwd)/src/hydra-eval-jobs:$(pwd)/src/hydra-queue-runner:$PATH
-            PERL5LIB=$(pwd)/src/lib:$PERL5LIB
-            export HYDRA_HOME="src/"
+            PERL5LIB=$(pwd)/src/lib:$(pwd)/t/lib:$PERL5LIB
```
grahamc, line 330: I'm not sure if this is a good idea: we don't really want the test's code getting in the environment of the actual program execution ... I'll have to put up the bat-signal for more of a perlmonger for this question.

flake.nix
```diff
@@ -319,13 +319,21 @@
               gzip bzip2 lzma gnutar unzip git gitAndTools.topGit mercurial darcs gnused breezy
             ] ++ lib.optionals stdenv.isLinux [ rpm dpkg cdrkit ] );
 
+          # Can be uncommented when https://github.com/TJC/Test-postgresql/issues/40 gets fixed.
+          # YATH_JOB_COUNT = "\${NIX_BUILD_CORES-$(nproc)}";
+          YATH_JOB_COUNT = "4";
```
grahamc, line 324: Hmm... We probably don't want to force parallel job running for everybody: some users will definitely want to run not in parallel. Maybe we should have it be not-parallel by default, and document -j? Not sure exactly. Maybe more specifically we'd want it to be `YATH_JOB_COUNT=max(4, NIX_BUILD_CORES, nproc)`

grahamc (commented):
Looking pretty good, but I'm not keen on the path calculation code duplication. I think we probably want to apply a patch like this: https://gsc.io/content-addressed/a9bfb8b2ee08dfcce97e42422ee85ef5e5313d6cd39ffab8322dc1455ab67749.patch (sorry, WIP I never pushed ... :)) and then extend the returned type with a `testdir` field, and maybe a `jobsdir` field.

flake.nix
cole-h: FWIW, this is only in the `shellHook` -- it's basically the same as running with `PERL5LIB=../:../src/lib:../tests/lib:$PERL5LIB` (from within the `tests` dir), if that changes anything.

grahamc: I think we want to make a [`.yath.rc`](https://metacpan.org/pod/App::Yath#PROJECT-SPECIFIC-YATH-CONFIG) so we can add the test lib for tests, but not pollute the PERL5LIB for the actual running of Hydra's executables during development time. It'd be gnarly to create code that can only run during test time!

Check out this diff:

```diff
[nix-shell:~/projects/github.com/NixOS/hydra]$ git diff --cached
diff --git a/.yath.rc b/.yath.rc
new file mode 100644
index 00000000..f5a68a06
--- /dev/null
+++ b/.yath.rc
@@ -0,0 +1,3 @@
+[test]
+-I=rel(t/lib)
+
diff --git a/flake.nix b/flake.nix
index b74060aa..40671415 100644
--- a/flake.nix
+++ b/flake.nix
@@ -327,7 +327,7 @@
             pushd $(git rev-parse --show-toplevel) >/dev/null
 
             PATH=$(pwd)/src/hydra-evaluator:$(pwd)/src/script:$(pwd)/src/hydra-eval-jobs:$(pwd)/src/hydra-queue-runner:$PATH
-            PERL5LIB=$(pwd)/src/lib:$(pwd)/t/lib:$PERL5LIB
+            PERL5LIB=$(pwd)/src/lib:$PERL5LIB
             export HYDRA_HOME="$(pwd)/src/"
             mkdir -p .hydra-data
             export HYDRA_DATA="$(pwd)/.hydra-data"

```

cole-h: Ah, I completely missed the point of your comment 😅

That sounds like a better idea, will do.

flake.nix
cole-h: I think I'll drop this altogether, in favor of updating the hacking docs to note the ability to run tests in parallel (with a sidenote that they may fail spuriously, with a link to that issue).

cole-h:
Just as a note for posterity: the `mv tests t` was necessary because `yath test` automatically runs tests in a `t` or `t2` subdirectory; otherwise, you'd have to specify a path (e.g. `yath test tests` or `yath test tests/input-types`).

PR #884, opened 2021-03-05
tests: replace the flat list of contexts with a hash

cole-h:
This way we can return more values without breaking callers.

---

Also add `jobsdir` and `testdir` to the ctx hash.

### DIFF
--- a/tests/lib/Setup.pm
+++ b/tests/lib/Setup.pm
@@ -5,7 +5,8 @@ use Exporter;
 use Test::PostgreSQL;
 use File::Temp;
 use File::Path qw(make_path);
-use Cwd;
+use File::Basename;
+use Cwd qw(abs_path getcwd);
 
 our @ISA = qw(Exporter);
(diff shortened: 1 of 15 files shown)


cole-h:
I would do the same with `tests/s3-backup-test.pl` (replace `getcwd."/jobs"` with `$ctx{jobsdir}`), but I don't know how to run it.

tests/lib/Setup.pm
```diff
@@ -98,7 +101,7 @@ sub nrQueuedBuildsForJobset {
 }
 
 sub createBaseJobset {
-    my ($jobsetName, $nixexprpath) = @_;
+    my ($jobsetName, $nixexprpath, $jobspath) = @_;
```
grahamc, line 104: I wasn't keen on this change: it seemed better to set the jobs path in a local variable in this module and refer to it from here. However, I could imagine a future where we're making a per-test jobs directory, and so this would be interesting.

grahamc (commented):
Nice.

PR #886, opened 2021-03-08
statsd: add a chance to set hostname and port in hydra.conf

grahamc:
Co-authored-by: Graham Christensen <graham@grahamc.com>

originally PR'd in #269 and updated, with tests and documentation. cc @matejc

### DIFF
--- a/src/script/hydra-send-stats
+++ b/src/script/hydra-send-stats
@@ -9,6 +9,11 @@ use JSON;
 STDERR->autoflush(1);
 binmode STDERR, ":encoding(utf8)";
 
+my $config = getHydraConfig();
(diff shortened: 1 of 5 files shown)


dasJ (approved):
Looks good code-wise, I cannot say anything about the functionality since I'm not using statsd

src/script/hydra-send-stats
```diff
@@ -9,6 +9,11 @@ use JSON;
 STDERR->autoflush(1);
 binmode STDERR, ":encoding(utf8)";
 
+my $config = getHydraConfig();
```
Mic92, line 12: That script no longer works because it does not import the hydra configuration.

PR #888, opened 2021-03-09
Make eval restriction an option instead of forcing

dasJ:


Ma27:
Might make sense to leave a note in e.g. `doc/manual/src/installation.md`, but LGTM :+1:

dasJ:
I wanted to document it next to the other evaluator options but they are not documented either :laughing:

kquick:
It seems like this would result in un-reproducible builds and builds that can be dependent on elements not captured in the store or the derivation?

dasJ:
Yes

kquick:
Maybe you could say more about what problem this addresses, because a CI  testing system that doesn't generate builds that are reproducible or stable doesn't seem ... useful?

ajs124:
AFAIR this enables IFD.

grahamc:
I am not thrilled that such an inauspicious PR got such an auspicious number.

dasJ:
Any updates on this? I don't really see why this would not be an option. cc @grahamc

PR #890, opened 2021-03-10
Bump Nix version

shlevy:


shlevy:
@regnat FYI, this may not do the right thing with CA stuff.

Issue #894, opened 2021-03-20
hydra-send-stats uses undefined function getHydraConfig


Mic92:
```
● hydra-send-stats.service
     Loaded: loaded (/nix/store/ws5saprglyw83al17hwpzvsnllf78h39-unit-hydra-send-stats.service/hydra-send-stats.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Sat 2021-03-20 05:40:50 UTC; 1min 51s ago
    Process: 28067 ExecStart=hydra-send-stats (code=exited, status=25)
   Main PID: 28067 (code=exited, status=25)
         IP: 0B in, 0B out
        CPU: 92ms

Mar 20 05:40:50 nix-community-build01 systemd[1]: Started hydra-send-stats.service.
Mar 20 05:40:50 nix-community-build01 hydra-send-stats[28067]: Undefined subroutine &main::getHydraConfig called at /nix/store/m3mz9nv3wad8qigf9mb88k8adk6pwacm-hydra-2021-03-10/bin/.hydra-se>
Mar 20 05:40:50 nix-community-build01 systemd[1]: hydra-send-stats.service: Main process exited, code=exited, status=25/n/a
Mar 20 05:40:50 nix-community-build01 systemd[1]: hydra-send-stats.service: Failed with result 'exit-code'.
```

Introduced in https://github.com/NixOS/hydra/pull/886

PR #895, opened 2021-03-20
hydra-send-stats: fix imports

Mic92:
fixes https://github.com/NixOS/hydra/issues/894

### DIFF
--- a/src/script/hydra-send-stats
+++ b/src/script/hydra-send-stats
@@ -2,6 +2,7 @@
 
 use strict;
 use utf8;
+use Hydra::Helper::Nix;
 use Net::Statsd;
 use File::Slurp;
 use JSON;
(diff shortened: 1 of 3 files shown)


Mic92:
Tested in nix-community: https://github.com/nix-community/infra/commit/a9295a7298a46bde2d1ef89a983e1e47da21e9bf

src/script/hydra-send-stats
```diff
@@ -2,14 +2,15 @@
...
+use Hydra::Helper::Nix;
 use Net::Statsd;
 use File::Slurp;
 use JSON;
 
 STDERR->autoflush(1);
 binmode STDERR, ":encoding(utf8)";
 
-my $config = getHydraConfig();
+my $config = Hydra::Helper::Nix::getHydraConfig();
```
grahamc, line 13: `use` brings it in to scope by default, so this part isn't necessary. I'll fix this up in a push I'm about to make adding some tests. Thanks!

grahamc: (I know I made this mistake one line lower. I know more perl than I did then :'))

grahamc (approved):
I pushed to your branch adding a test, making sure it failed, and that your commit fixed it. I also ran the test without "--once" to make sure it took "forever". It'd be nice to have a more thorough test, but I'm happy enough with "does it run without exiting badly?" for now.

Mic92:
I deployed this patch to nix-community.

PR #899, opened 2021-03-25
Projects, jobsets: order jobsets with disabled, hidden rows at the end

grahamc:
The rows otherwise are often improperly zebra striped.

### DIFF
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -104,7 +104,7 @@ sub deserialize :ActionClass('Deserialize') { }
 sub index :Path :Args(0) {
     my ($self, $c) = @_;
     $c->stash->{template} = 'overview.tt';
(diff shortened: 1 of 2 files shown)


src/lib/Hydra/Controller/Root.pm
```diff
@@ -104,7 +104,7 @@ sub deserialize :ActionClass('Deserialize') { }
 sub index :Path :Args(0) {
     my ($self, $c) = @_;
     $c->stash->{template} = 'overview.tt';
-    $c->stash->{projects} = [$c->model('DB::Projects')->search({}, {order_by => 'name'})];
+    $c->stash->{projects} = [$c->model('DB::Projects')->search({}, {order_by => ['enabled DESC', 'name']})];
```
grahamc, line 107: It turned out I missed a `[...]` here, and I've pushed a fixup. This didn't cause a problem, apparently the table is naturally sorted by name anyway. However, it did make the parameters to `search` a bit strange. Also strange: perlcritic didn't complain about it.

PR #904, opened 2021-03-30
Add `GiteaStatus`-Plugin

Ma27:
This plugin is fairly heavily inspired by the implementations for GitHub and GitLab. It basically checks for the build-state of a jobset evaluation and then marks the commit as `failed`/`succeeded` using the REST API.

A simple example how to configure/use is added to the manual, also it's verified using a VM test in `flake.nix`.

cc @grahamc @cole-h

### DIFF
--- a/doc/manual/src/projects.md
+++ b/doc/manual/src/projects.md
@@ -486,9 +486,10 @@ analogous:
 * For a jobset with a `Git`-input which points to a `gitea`-instance, add the following
   additional inputs:
 
-  | Type           | Name                | Value                               |
-  | -------------- | ------------------- | ----------------------------------- |
-  | `String value` | `gitea_repo_name`   | *Name of the repository to build*   |
-  | `String value` | `gitea_repo_owner`  | *Owner of the repository*           |
-  | `String value` | `gitea_status_repo` | *Name of the `Git checkout` input*  |
+  | Type           | Name                | Value                              |
+  | -------------- | ------------------- | ---------------------------------- |
+  | `String value` | `gitea_repo_name`   | *Name of the repository to build*  |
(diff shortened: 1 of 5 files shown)


Ma27:
@grahamc as discussed in `#nixos-dev` a while ago, I added a `yath`-test.

However the package build is currently failing like this:

```
hydra> ( STDERR )  job 17    "my" variable $uri masks earlier declaration in same scope at plugins/gitea.t line 75.
hydra> ( STDERR )  job 17    No such file or directory at /nix/store/bamdbsi5dirrsrlm84rrsq6dz4hjmh1s-hydra-perl-deps/lib/perl5/site_perl/5.32.0/Test/HTTP/MockServer.pm line 20.
hydra> ( FAILED )  job 17    plugins/gitea.t
hydra> < REASON >  job 17    Test script returned error (Err: 2)
hydra> < REASON >  job 17    No plan was declared, and no assertions were made.
```

The line in question has `my $proto = getprotobyname("tcp") or die $!;`. I don't really understand why this would break here, but that's enough Perl for today :D 

Do you have an idea there?

Ma27:
Ok, it is because `getprotobyname` requires `/etc/protocols` from `iana-etc`. Let's see if I can patch the sandbox env together.

Ma27:
OK, I did a bit more digging, but didn't understand, why the CPAN modules were ignoring the `libredirect` trick I used to make `/etc/protocols` available. Maybe something fiddles around in the environment or a libc function is somewhere used that's not wrapped in libredirect (wouldn't be the first time I encountered this, but pretty painful to debug).

I hacked together a small testserver in Python now. Sorry for bringing in another language. Feel free to replace it with a bash equivalent :)

Ma27:
ping @grahamc is there anything else tbd? :)

PR #906, opened 2021-04-04
tree view of deps

tomberek:
This starts to convert the deps tree into a toggle-able tree. Is there any interest in this? It's potentially easier to navigate, but removes ability to Ctrl-F (any way around this? an "expand-all" toggle? or just start with all of them expanded?). 

20210404_02h38m54s_grim


SuperSandro2000:
Are you changing this tree https://hydra.nixos.org/build/140784550#tabs-build-deps ? If so I really appreciate it because right now it is pretty much useless.

> an "expand-all" toggle?

that would work but would be laggy. Maybe some fancy js can make them searchable?

tomberek:
> `Are you changing this tree https://hydra.nixos.org/build/140784550#tabs-build-deps ?

Yes, that's the idea, make the very large trees a bit more manageable. I'm trying to support both a "nice view" and "Ctrl-F" at the same time. Preferably without JS.

samueldr:
@tomberek Expand all should work fine I assume.

The only performance concern I have is *already present* in the current implementation, the initial rendering *could* take a while, but it would *already* take a while if it did. (I made no measurements.)

I guess it's easy enough to test `$("details").each(function () { this.setAttribute("open", "true") })` in your JS console. The real implementation should use a selector like "all details child of the element next to self", just in case we have multiple trees in a page in the future.

tomberek:
out

grahamc:
@tomberek did you intend for this to be a draft PR?

tomberek:
It’s a draft because I’m not quite sure what the right design is. So I am exploring the possibilities and looking for any feedback or suggestions.

tomberek:
Are the buttons and collapse levels too complex and we can just go with "Expand All"/"Collapse All"?

Mindavi:
Implementation-wise they don't look that complex, but expand/collapse all seems the most useful.

Could be useful to have those different levels, to indeed make it easier to navigate. Having it fully closed is not super useful (but I guess you then open by hand?), and I guess having it fully expanded was the previous behavior already.

PR #910, opened 2021-04-12
UI fixups

grahamc:
This add button did not cross all the way:

image

And, new jobsets now default to Enabled:

image


samueldr (approved):
Looks good in implementation. I have not tested it, but can't see issues in the code.

PR #911, opened 2021-04-12
meta.outputsToInstall: store and propogate

grahamc:
Allow RunCommand to access the outputs to install, and propagate the fake expressions to the generated Nix channel.


edolstra:
I'm not very enthusiastic about adding a column to the `Builds` table to support an essentially deprecated feature (Hydra channels, or channels in general). It would probably be better to move channel generation out of Hydra (e.g. into a RunCommand script, or the way the NixOS/Nixpkgs channels are updated).

grahamc:
This was originally developed for the generated Nix channel, but the user of that feature has stopped using it. Now, it is being used to provide the outputsToInstall data to the `RunCommand` hook.

grahamc:
note to self: replace this PR with storing the entire derivation with `show-derivation`.

PR #915, opened 2021-04-15
Hydra auth: support Argon2, transparently upgrade hashes

### LINKED ISSUE #63: Use salted password hashes
Because unsalted hashes are bad m'kay.

grahamc:
New users will have their passwords stored with Argon2id. Existing users that already have unsalted sha1 passwords will have their stored password transparently migrated to Argon2id.

This patch also upgrades the sha1 checking to be constant time, an improvement over Catalyst's default behavior.

Closes #63

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -229,6 +229,19 @@
               license = with final.stdenv.lib.licenses; [ artistic1 ];
             };
           };
+
+          StringCompareConstantTime = final.buildPerlPackage {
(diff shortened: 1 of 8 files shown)


grahamc:
Note Argon2 was selected since it is the current recommendation by OWASP: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

PR #916, opened 2021-04-16
Rehash existing sha1 passwords with Argon2

grahamc:
In #915 we transparently upgrade user passwords to Argon2 on login. However, as [wkral mentioned on Discourse,](https://discourse.nixos.org/t/hydra-upgrade-hashes-to-argon2/12497/4?u=grahamc), this leaves the passwords poorly hashed in the database for users who don't log in. They suggested what [OWASP suggests](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#upgrading-legacy-hashes):

> use the existing password hashes as inputs for a more secure algorithm. For example, if the application originally stored passwords as md5($password), this could be easily upgraded to bcrypt(md5($password)). Layering the hashes avoids the need to know the original password; however, it can make the hashes easier to crack. These hashes should be replaced with direct hashes of the users' passwords next time the user logs in.

I have implemented that here.

The first commit extends the password validation routine to support this layered hash. If the user's password does not validate as either Argon2 or sha1, the check_password routine hashes the user-provided password with sha1 and tries the password validation routine again. If the password is valid after hashing, the stored password hash is updated to only be Argon2 hashed.

The second commit updates hydra-create-user to write this layered hash if the `--password-hash` flag is used and the provided hash is a 40 character hexadecimal value. This commit also lays the ground work for the third commit ...

The third commit has hydra-init update all sha1 passwords in the database as part of the initial migration. It is unlikely that new sha1 passwords will be entered in to the database at this point, but if they do they will be immediately upgraded to the nested Argon2 hash.

Since it is still possible for sha1 passwords to exist in the database, this PR does not remove (nor do I think it should) the plain sha1 alternative to Argon2 in the check process.

### DIFF
--- a/src/lib/Hydra/Schema/Users.pm
+++ b/src/lib/Hydra/Schema/Users.pm
@@ -239,6 +239,12 @@ sub check_password {
         }
 
         return 1;
+    } elsif ($authenticator->verify_password(sha1_hex($password), $self->password)) {
+        # The user's database record has their old password as sha1, re-hashed as Argon2.
+        # Store their password hashed only with Argon2.
+        $self->setPassword($password);
+
+        return 1;
     } else {
         return 0;
     }
--- a/src/lib/Hydra/Schema/Users.pm
+++ b/src/lib/Hydra/Schema/Users.pm
@@ -258,4 +258,15 @@ sub setPassword {
     });
 }
 
+sub setPasswordHash {
+    my ($self, $passwordHash) = @_;;
+
+    if ($passwordHash =~ /^[a-f0-9]{40}$/) {
+        # This is (probably) a sha1 password, re-hash it and we'll check for a hashed sha1 in Users.pm
+        $self->setPassword($passwordHash);
+    } else {
+        $self->update({ password => $passwordHash });
+    }
+}
+
 1;
--- a/src/script/hydra-create-user
+++ b/src/script/hydra-create-user
@@ -109,7 +109,10 @@ $db->txn_do(sub {
         if (defined $password && !(defined $passwordHash)) {
             $user->setPassword($password);
         }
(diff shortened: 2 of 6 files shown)


SuperSandro2000:
Maybe we should notify users to change their passwords just in case :)

grahamc:
We probably should, though note this change is impacting more deployments than hydra.nixos.org. hydra.nixos.org has very few users which log in with a password. I'll contact those users.

andir (approved):

grahamc:
Deployed, and all users who log in with a password have been notified with this text:

---

Hello,

My name is Graham Christensen, an administrator of hydra.nixos.org and a member of the NixOS Infrastructure Team[1] and Security Team[2].

This week I developed[3,4] and deployed an improvement to Hydra changing the way Hydra stores passwords. The new mechanism is more robust against brute force attacks and rainbow tables, an improvement over the previously outdated process.

This request is just a precaution: we have no reason to think the database or user information has been leaked or any security breach has taken place.

Please take a moment and change your password on hydra.nixos.org:

1. Visit https://hydra.nixos.org/
2. Click "Sign in" in the top right of the page
3. Click "Sign in with a Hydra account"
4. Enter your username and password then click "Sign In"
5. Click "Preferences" in the top right
6. Enter your NEW password in the "Password" and "Confirm password" fields
7. Click "Apply changes"

If you're no longer using your Hydra account, please reply and I will disable your account instead. If you would rather use GitHub or Google to log in, let me know.

Thank you so much for your time and attention. If you have any questions, don't hesitate to ask.

Best,
Graham Christensen

1. https://nixos.org/community/teams/infrastructure.html
2. https://nixos.org/community/teams/security.html
3. https://github.com/NixOS/hydra/pull/915
4. https://github.com/NixOS/hydra/pull/916

Issue #688, opened 2019-11-09
No support for ssh-ng:// protocol


knedlsepp:
The most recent version of nix supports coloured build outputs even when doing remote builds by specifying the build machines with the `ssh-ng://` protocol in the `/etc/nix/machines` file.
This in itself works great, however I also have configured a hydra instance on that particular machine. Hydra's parser of the `machines` file extracts 'ssh-ng://hostname' as the hostname, which results in connection failure.

Ericson2314:
Hydra uses its own custom gRPC protocols, and the daemon protocol now. There is nothing more to do here.

PR #918, opened 2021-04-17
hydra-queue-runner: fix building on machines specified with the `ssh-ng` protocol

Ma27:
Most of my machines in `/etc/nix/machines` are specified as
`ssh-ng://foobar` where `foobar` is defined in e.g. `~/.ssh/config`. The
problem here is that while it's possible to use the remote Nix
installation via SSH, `hydra-queue-runner` refuses to do so since `ssh(1)`
can't resolve hostnames that begin with `ssh-ng://`.

To work around this problem, this prefix will be drop if it exists on a
machine declaration that was chosen for a build step.

cc @grahamc @edolstra

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -54,7 +54,13 @@ static void openConnection(Machine::ptr machine, Path tmpDir, int stderrFD, Chil
         }
         else {
             pgmName = "ssh";
-            argv = {"ssh", machine->sshName};
+            auto sshName = machine->sshName;
+            std::string prefix = "ssh-ng://";
+            size_t pos = sshName.find(prefix);
+            if (pos != std::string::npos) {
+                sshName.erase(pos, prefix.length());
+            }
+            argv = {"ssh", sshName};
             if (machine->sshKey != "") append(argv, {"-i", machine->sshKey});
             if (machine->sshPublicHostKey != "") {
                 Path fileName = tmpDir + "/host-key";


Ericson2314 (approved):

grahamc:
It seems like we would be better served by using Nix's own store abstractions instead of running our own `ssh` incantation. I don't suppose there is a way we could test this behavior?

Ma27:
> I don't suppose there is a way we could test this behavior?

Well, I have an entry like this in my `/etc/nix/machines`:

```
ssh-ng://builder-ng x86_64-linux - 10 4 big-parallel,kvm,nixos-test
```

With this patch on a Hydra, `hydra-queue-runner` happily picks up this server to build derivations :) 

> It seems like we would be better served by using Nix's own store abstractions instead of running our own ssh incantation

Not sure if I agree here. First of all, this `nix-store --serve` via `ssh` approach is also used by e.g. `nix.sshServe` (IIRC). Also, to me this currently sounds like yet another thing to catch up with (rather rapid) changes in `nixUnstable`.

knedlsepp:
Fixes: https://github.com/NixOS/hydra/issues/688

edolstra:
This PR just makes hydra-queue-runner treat `ssh-ng://` as `ssh://`, which IMHO is not all that useful and probably isn't what users expect.

Currently hydra-queue-runner uses the `nix-store --serve` protocol, so it can only use `ssh://`. I have a branch somewhere that makes it use the `Store` abstraction. However, we can't use it because of some build log issues (fixable) and lack of build step cancellation (not fixable without some major changes to the Store API - currently build steps are cancelled by sending a signal to the corresponding SSH process).

Ma27:
Hmm yeah, so I know the difference between `ssh-ng` and `ssh`, however this was supposed to be a cheap fix for existing `/etc/nix/machines` where I use `ssh-ng` by default for good reasons.

But if you've got already a draft for using Nix's store abstraction in Hydra (and hence allow both protocols), that's way better, so :+1:  for closing it then :)

edolstra:
To be clear, the code that I wrote isn't usable because it doesn't support cancellation.

ajs124:
The current code supports cancellation? I'm not sure that's ever worked for me.

grahamc:
It does support cancellation, and it often does work :)

dasJ:
I thought the exact log message was "maybe cancelling build" and it finishes the current step and aborts the rest?

Ericson2314:
> To be clear, the code that I wrote isn't usable because it doesn't support cancellation.

Where is that branch again? I've been thinking if some of legacy ssh store and remote store methods were split up to separate the parts hydra needs to do separately, it could work. This code split would be of no use to Nix itself, but I think the separation could be worth it.

edolstra:
Looks like I never committed anything, but I recovered this from my stash: https://pastebin.com/suP2ZX2D

PR #921, opened 2021-04-24
Make it possible to enable email notifications when creating a jobset

Ma27:
The checkbox is only enabled if `email_notification = 1` is set in
`hydra.conf`. However, when creating jobset (in contrast to the edit
form), the checkbox is always disabled because the `emailNotification`
parameter in Catalyst's stash was missing.

cc @grahamc


grahamc:
Thanks!

PR #922, opened 2021-04-24
Allow specifying jobset inputs for flake builds for Hydra plugins

Ma27:
It seems to be a common pattern to configure Hydra plugins on a
per-jobset basis with jobset inputs. Since these are not needed anymore
for flakes, it's also not possible anymore to use plugins for flake
jobsets.

This patch changes this and adds a warning to avoid confusion. In the
future, we may want to restrict jobset inputs for flakes to string
values only.

cc @grahamc

### DIFF
--- a/src/lib/Hydra/Controller/Jobset.pm
+++ b/src/lib/Hydra/Controller/Jobset.pm
@@ -285,26 +285,24 @@ sub updateJobset {
     # Set the inputs of this jobset.
     $jobset->jobsetinputs->delete;
 
-    if ($type == 0) {
-        foreach my $name (keys %{$c->stash->{params}->{inputs}}) {
-            my $inputData = $c->stash->{params}->{inputs}->{$name};
-            my $type = $inputData->{type};
-            my $value = $inputData->{value};
-            my $emailresponsible = defined $inputData->{emailresponsible} ? 1 : 0;
-            my $types = knownInputTypes($c);
+    foreach my $name (keys %{$c->stash->{params}->{inputs}}) {
+        my $inputData = $c->stash->{params}->{inputs}->{$name};
+        my $type = $inputData->{type};
+        my $value = $inputData->{value};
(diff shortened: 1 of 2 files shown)


Ma27:
also CCing @edolstra whoi originally implemented flake support for Hydra.

grahamc:
I'm not sure it is a good idea to bring back jobset inputs like this. I wonder if it would make more sense to have plugin configuration be a different thing? What plugins use jobset inputs for config?

grahamc:
Concretely, I can imagine someone being confused about why their jobset inputs aren't available to their jobset. Also, it causes perhaps internal confusion about if the inputs should be fetched, or input values tracked over time... overall I'm not keen on this. I wonder if we could create a different way to provide the needed configuration to plugins, or if the plugins should adapt in a different way.

Ma27:
> I wonder if it would make more sense to have plugin configuration be a different thing?

@grahamc plugins for gitea/gitlab/bitbucket also require a `git` input to obtain e.g. the git revision. So in the end we'd need some input fetching either way AFAIU.

> I wonder if we could create a different way to provide the needed configuration to plugins, or if the plugins should adapt in a different way.

I could think of a "plugin configuration", but I'm uncertain about the following aspects:

* should this be a purely UI-related change (with good documentation in the code, of course to avoid confusion for contributors)
* how should "Legacy" jobsets (not an expression of opinion, just using the terminology of Hydra btw :D) behave then? Should these mix up the plugin configuration with jobset inputs (that would speak for not separating the datastructure IMHO) or should we segregate these kinds of inputs there as well? (That would be a pretty big breaking change and something I would not want to backport to a stable NixOS btw).

Ma27:
@grahamc rebased onto latest master.

Ma27:
@grahamc any updates here? :)

Ma27:
@grahamc given your latest comments in the Hydra matrix channel, do you consider this mergeable now? :)

Ma27:
@grahamc @cole-h anything I can currently do to get this one merged? :)

PR #923, opened 2021-04-24
Add AuthenSASL to perl deps; improve email docs

Ma27:
The addition of AuthenSASL seems to be necessary to properly
authenticate against an SMTP server. Without this I got errors
such as

    error with Hydra::Plugin::EmailNotification=HASH(0x6ad0128)->buildFinished: SMTP auth requires MIME::Base64 and Authen::SASL

cc @grahamc

### DIFF
--- a/doc/manual/src/projects.md
+++ b/doc/manual/src/projects.md
@@ -458,6 +458,45 @@ Hydra can send email notifications when the status of a build changes.
 This provides immediate feedback to maintainers or committers when a
 change causes build failures.
 
+The feature can be turned on by adding the following line to `hydra.conf`
+
+``` conf
+email_notification = 1
+```
+
+By default, Hydra only sends email notifications if a previously successful
+build starts to fail. In order to force Hydra to send an email for each build
+(including e.g. successful or cancelled ones), the environment variable
+`HYDRA_FORCE_SEND_MAIL` can be declared:
+
+``` nix
+services.hydra-dev.extraEnv.HYDRA_FORCE_SEND_MAIL = "1";
+```
+
+SASL Authentication for the email address that's used to send notifications
+can be configured like this:
+
+``` conf
+EMAIL_SENDER_TRANSPORT_sasl_username=hydra@example.org
+EMAIL_SENDER_TRANSPORT_sasl_password=verysecret
+EMAIL_SENDER_TRANSPORT_port=587
+EMAIL_SENDER_TRANSPORT_ssl=starttls
+```
+
+Further information about these environment variables can be found at the
+[MetaCPAN documentation of `Email::Sender::Manual::QuickStart`](https://metacpan.org/pod/Email::Sender::Manual::QuickStart#specifying-transport-in-the-environment).
+
+It's recommended to not put this in `services.hydra-dev.extraEnv` as this would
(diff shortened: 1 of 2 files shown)


doc/manual/src/projects.md
```diff
@@ -458,6 +458,32 @@ Hydra can send email notifications when the status of a build changes.
...
 change causes build failures.
 
+The feature can be turned on by adding the following line to `hydra.conf`
+
+    email_notification = 1
+
+Hydra can be forced to send an email for each build using the environment
+variable `HYDRA_FORCE_SEND_MAIL`:
+
+    services.hydra-dev.extraEnv.HYDRA_FORCE_SEND_MAIL = "1";
```
grahamc, line 468: Why would someone use the environment variable instead of the configuration value?

Ma27: I'm not aware of any equivalent configuration value. Please note that this isn't a "replacement" for `email_notification = 1` in `hydra.conf` It essentially forces Hydra to send emails even for e.g. succeeded builds (or repeatedly failing builds etc). See also `src/lib/Hydra/Plugin/EmailNotification.pm` at line ~60.

grahamc (commented):
Looks good, would you mind changing thees code blocks to use fences, like:

` ```nix ` and ` ```conf `?

doc/manual/src/projects.md
Ma27: But as you seem to misunderstand it, I guess I should rephrase that :)

Ma27:
@grahamc fixed :)

doc/manual/src/projects.md
```diff
@@ -458,6 +458,42 @@ Hydra can send email notifications when the status of a build changes.
...
+The feature can be turned on by adding the following line to `hydra.conf`
+
+``` conf
+email_notification = 1
+```
+
+By default, Hydra only sends email notifications if a previously successful
+build starts to fail. In order to force Hydra to send an email for each build
+(including e.g. successful or cancelled ones), the environment variable
+`HYDRA_FORCE_SEND_MAIL` can be declared:
```
grahamc, line 470: I'm not suuuuper keen about these magic environment variables, and would rather wish they were actual configuration  options (consider: email_notifications = no|yes|always) but we've got the code we've got :).

doc/manual/src/projects.md
```diff
@@ -458,6 +458,42 @@ Hydra can send email notifications when the status of a build changes.
...
+
+``` nix
+services.hydra-dev.extraEnv.HYDRA_FORCE_SEND_MAIL = "1";
+```
+
+SASL Authentication for the email address that's used to send notifications
+can be configured like this:
```
grahamc, line 477: Is there somewhere the user can learn more about these variables, and perhaps other similar variables?

Ma27: I have to admit that I looked up the `sasl_username`/`sasl_password` things directly up from the source code of the CPAN module for the mailer :D 

But I guess it makes sense to leave a link to the CPAN page of the module at least :)

Ma27: Added a link to the metacpan page.

grahamc (approved):
Thanks, looks great!

Issue #926, opened 2021-04-26
Attempting to create a project with a nonexistent user should return 400


cole-h:
```
$ curl -v -X 'PUT' 'http://0:63333/project/nix2' \
-H 'accept: application/json' \
-H 'Content-Type: application/json' \
-d '{
  "displayname": "string",
  "enabled": true,
  "hidden": true,
  "owner": "alice2",
  "description": "test"
}'

*   Trying 0.0.0.0:63333...
* Connected to 0 (127.0.0.1) port 63333 (#0)
> PUT /project/nix2 HTTP/1.1
> Host: 0:63333
> User-Agent: curl/7.74.0
> Cookie: hydra_session=c56881b67c794797f461880c2d5c9fc6c1e4c201
> accept: application/json
> Content-Type: application/json
> Content-Length: 112
>
* upload completely sent off: 112 out of 112 bytes
* Mark bundle as not supporting multiuse
* HTTP 1.0, assume close after body
< HTTP/1.0 500 Internal Server Error
< Date: Mon, 26 Apr 2021 18:21:03 GMT
< Server: HTTP::Server::PSGI
< Vary: Accept
< Content-Type: application/json
< Set-Cookie: hydra_session=c56881b67c794797f461880c2d5c9fc6c1e4c201; path=/; expires=Mon, 03-May-20
21 18:21:03 GMT; HttpOnly; SameSite=Lax
< Content-Length: 54
<
* Closing connection 0
{"error":"The user name ‘alice2’ does not exist."}
```

grahamc:
This looks pretty easy to fix. I used search.nix.gsc.io to find the relevant bit of code:

https://search.nix.gsc.io/?q=The%20user%20name%20%E2%80%98&i=nope&files=&repos=

https://github.com/NixOS/hydra/blob/bc292db45bc553331874c00fa60bc23798823e9c/src/lib/Hydra/Controller/Project.pm#L139-L141

and it looks like we just need to add an error code of 400 to it:

```perl
error($c, "msg", 410)
```

cole-h:
Done in https://github.com/NixOS/hydra/pull/927.

PR #927, opened 2021-04-26
Return HTTP 400 when creating Project with nonexistent user

cole-h:
Closes https://github.com/NixOS/hydra/issues/926.


grahamc (approved):
Nice abstracting out the sub.

PR #934, opened 2021-04-27
hydra-api: add missing `name` property

cole-h:
The Hydra website allows you to change the name of a Project, so it only makes
sense for the API to expose this functionality as well.

### DIFF
--- a/hydra-api.yaml
+++ b/hydra-api.yaml
@@ -160,8 +160,11 @@ paths:
             schema:
               type: object
               properties:
-                displayname:
+                name:
                   description: name of the project
+                  type: string
+                displayname:
+                  description: display name of the project
                   type: string
                 description:
                   description: description of the project


grahamc (approved):

grahamc:
Confirmed it is already in the response:

```

[grahamc@hyperchicken:~/projects/github.com/NixOS/hydra]$ curl --header "Accept: application/json" http://127.0.0.1:63333/project/nixpkgs
{"jobsets":[],"enabled":true,"description":"Nix Packages collection","name":"nixpkgs","hidden":true,"displayname":"Nixpkgs","owner":"alice","homepage":"http://nixos.org/nixpkgs"}
```

grahamc:
Oh this is in the PUT. Right. And the PUT handler DOES already handle it. Cool. :+1:

PR #935, opened 2021-04-27
hydra-api: implement DELETE /project/{id}

cole-h:


hydra-api.yaml
```diff
@@ -258,6 +258,33 @@ paths:
...
+        '200':
+          description: project deleted
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  redirect:
+                    type: string
+                    description: url of the Hydra instance
```
grahamc, line 280: Like, the root URL?

```suggestion
                    description: the root of the Hydra instance
```

cole-h: Yes, thanks, I always forget this wording :D

PR #939, opened 2021-04-28
Include localhost's features in the default machine file

### LINKED ISSUE #936: The default machine file doesn't include localhost's declared features
If hydra doesn't have a specified machine file, it’ll default to an [auto-generated one](https://github.com/NixOS/hydra/blob/107d60027fa22c3b919cce3e8565affbbd8aed3b/src/hydra-queue-runner/hydra-queue-runner.cc#L156-L159) with only `localhost` as a remote builder.
This generated machine file will specify the system and max-jobs according to the Nix configuration. However it doesn’t specify any `system-features` for it, meaning that building a derivation with a non-empty `requiredSystemFeatures` requires an explicit machine file (even if the nix conf specifies the relevant system features).

regnat:
Fix #936

### DIFF
--- a/t/default-machine-file.t
+++ b/t/default-machine-file.t
@@ -1,0 +1,30 @@
+use feature 'unicode_strings';
+use strict;
+use Setup;
+
+my %ctx = test_init();
+
+require Hydra::Schema;
+require Hydra::Model::DB;
+
+use Test2::V0;
+
+my $db = Hydra::Model::DB->new;
+hydra_setup($db);
+
+my $project = $db->resultset('Projects')->create({name => "tests", displayname => "", owner => "root"});
+
+my $jobset = createBaseJobset("default-machine-file", "default-machine-file.nix", $ctx{jobsdir});
+
+ok(evalSucceeds($jobset),               "Evaluating jobs/default-machine-file.nix should exit with return code 0");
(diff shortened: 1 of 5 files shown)


t/lib/Setup.pm
```diff
@@ -48,6 +48,7 @@ sub test_init {
     my $nixconf = "$ENV{'NIX_CONF_DIR'}/nix.conf";
     open(my $fh, '>', $nixconf) or die "Could not open file '$nixconf' $!";
     print $fh "sandbox = false\n";
+    print $fh "system-features = test-system-feature\n";
```
grahamc, line 51: Actually if you could delete this line from Setup.pm and ...

t/default-machine-file.t
```diff
@@ -0,0 +1,30 @@
+use feature 'unicode_strings';
+use strict;
+use Setup;
+
+my %ctx = test_init();
```
grahamc, line 5: move it here, with

```
my %ctx = test_init(
  nix_config => q|
system-features = test-system-feature
|
);
```

(as documented in https://github.com/regnat/hydra/blob/abff212d0631726502ce8a025e5ac5e4493828aa/t/lib/Setup.pm#L22)

grahamc (commented):
Looks pretty good, but can you also move the test under `t/queue-runner`?

PR #940, opened 2021-04-28
Run the tests with a remote dest store

### LINKED ISSUE #938: The testsuite never checks what happens when `destStore != localStore`
Hydra distinguishes between the “local store” (the one on which the evaluation takes place) and the “destination store” (the one to which all the build outputs are sent). However in the test suite both these stores are the same one, which might hide some logic errors (caused me some surprise in https://github.com/NixOS/hydra/pull/875 for example).

This could be fixed for example by setting `store_uri = file:/some/where` in the hydra config used for the tests.

regnat:
Set `dest_store` in the test hydra config, so that the testsuite ensures
that the distinction between the local store and the destination store
is properly taken into account.

Fix #938

This currently causes the `channel.t` test to fail − [this call](https://github.com/NixOS/hydra/blob/019aef3d419b03266f7a137300b54bcfa167350e/t/Controller/Jobset/channel.t#L36) returns a 404.
I'm not sure whether this is an issue with the test or hydra itself (I've no idea how the channels feature works). @grahamc do you have an idea as the author of the test?

### DIFF
--- a/t/lib/Setup.pm
+++ b/t/lib/Setup.pm
@@ -20,6 +20,9 @@ our @EXPORT = qw(test_init hydra_setup nrBuildsForJobset queuedBuildsForJobset
 #
 #  * hydra_config: configuration for the Hydra processes for your test.
 #  * nix_config: text to include in the test's nix.conf
+#  * use_external_destination_store: Boolean indicating whether hydra should
+#       use a destination store different from the evaluation store.
+#       True by default.
 #
 # This clears several environment variables and sets them to ephemeral
 # values: a temporary database, temporary Nix store, temporary Hydra
--- a/t/lib/Setup.pm
+++ b/t/lib/Setup.pm
@@ -54,6 +54,7 @@ sub test_init {
     $ENV{'HYDRA_CONFIG'} = "$dir/hydra.conf";
 
     open(my $fh, '>', $ENV{'HYDRA_CONFIG'}) or die "Could not open file '" . $ENV{'HYDRA_CONFIG'}. " $!";
+    print $fh "store_uri = file:$dir/nix/dest-store\n";
     print $fh $opts{'hydra_config'} || "";
     close $fh;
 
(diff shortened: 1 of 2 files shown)


grahamc:
It isn't clear to me if all the tests should, by default, use an external store... Hydra supports both modes of operation, so maybe some tests should be written to explicitly check it both ways: once with a local store, once with a remote store, with the store_dir passed in through hydra_config.

regnat:
> It isn't clear to me if all the tests should, by default, use an external store... Hydra supports both modes of operation, so maybe some tests should be written to explicitly check it both ways: once with a local store, once with a remote store, with the store_dir passed in through hydra_config.

My reasoning for that was that using an external dest store is strictly more restrictive than not using it, so it's much more unlikely to have a bug that shows up without it but not with it than the inverse. But maybe we could have the tests use an external store by default, and have some tests also not use it

t/lib/Setup.pm
```diff
@@ -54,6 +54,7 @@ sub test_init {
     $ENV{'HYDRA_CONFIG'} = "$dir/hydra.conf";
 
     open(my $fh, '>', $ENV{'HYDRA_CONFIG'}) or die "Could not open file '" . $ENV{'HYDRA_CONFIG'}. " $!";
+    print $fh "store_uri = file:$dir/nix/dest-store\n";
```
grahamc, line 57:
```suggestion
    if ($opts{'use_external_destination_store'} || true) {
        print $fh "store_uri = file:$dir/nix/dest-store\n"
    }
```

grahamc (commented):
Okay, I'm persuaded. I think the channel feature probably doesn't work when the store is remote. How about conditioning this line on a new argument to the `test_init` function? Maybe named something like `use_external_destination_store` and defaulting to true if it isn't set:

then documenting it here: https://github.com/NixOS/hydra/blob/0d3977866e14b9bd912b738a82ed6fd5184fcd88/t/lib/Setup.pm#L21

and then editing the channel test to set it to false.

regnat:
> I think the channel feature probably doesn't work when the store is remote

Indeed it doesn’t https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Helper/CatalystUtils.pm#L416-L419

grahamc:
Thanks!

PR #942, opened 2021-04-28
Port api-test to yath

grahamc:
This will prevent us from breaking the test in the future, since it'll run on all GitHub PRs. I inadvertently broken 2 pieces in the test, one in 725c9c2f81b6c3a2082b985bf25772c60c1b4f37 (users) and the other in a rename of jobsetinputs.


cole-h (approved):
Diff LGTM!

Do we want to fix up some of the issues with the test itself (e.g. `"my" variable $result masks earlier declaration in same scope at t/api-test.t line 56.`, `"my" variable $evals masks earlier declaration in same scope at t/api-test.t line 100.`, etc.)?

PR #943, opened 2021-04-29
Cleanup the localhost-features test

regnat:
- Only add the Nix system feature to the test that requires it
- Move the default-machine-file test under `queue-runner`

follow-up for <https://github.com/NixOS/hydra/pull/939#pullrequestreview-647532083>

/cc @grahamc


grahamc:
Thanks for the cleanup!

PR #950, opened 2021-04-30
Properly report hash mismatches

### LINKED ISSUE #947: Hash-mismatch in fixed-output derivations aren’t properly reported
When hydra is asked to build a fixed-output derivation that has a wrong hash, it will report it as failed, but without indicating that the failure is related to a hash mismatch (the logs won’t show it, and the step will just report that it’s failed).

## What I’d like to see

Some clear indication that the failure is due to a hash mismatch, either in the build logs or in the build page

## A bit more context

The issue is that:

1. This doesn’t appear in the logs because these only contain the actual output of the builder (and the hash mismatch error is an exception thrown by the `nix-store` process
2. The exception message should be propagated by hydra, but for some reason [it is set to an empty string](https://github.com/NixOS/hydra/blob/master/src/hydra-queue-runner/build-remote.cc#L390). @edolstra do you remember why it was this way and whether this still makes sense?

regnat:
Fix #947

I’m not utterly sure what the implications of not setting `errorMsg` to the empty string are. Removing it seems like the right thing to do, but maybe I’m missing something

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -387,8 +387,6 @@ void State::buildRemote(ref<Store> destStore,
         }
         if (result.stepStatus != bsSuccess) return;
 
-        result.errorMsg = "";
-
         /* If the path was substituted or already valid, then we didn't
            get a build log. */
         if (result.isCached) {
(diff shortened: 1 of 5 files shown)


Ericson2314:
We should still do this, right?

PR #952, opened 2021-05-02
Project: add declfile, decltype, declvalue to API

cole-h:
This makes it possible to create declarative projects via a PUT request, and
also exposes the currently-configured values to GET requests.

---

`decl*` are all nullable -- should these be string columns (where they are `""` if null), or "normal" columns (where they are actual `null`s if null)?

### DIFF
--- a/hydra-api.yaml
+++ b/hydra-api.yaml
@@ -181,6 +181,10 @@ paths:
                 visible:
                   description: when set to true the project is displayed in the web interface
                   type: boolean
+                declarative:
+                  description: declarative input configured for this project
+                  type: object
+                  $ref: '#/components/schemas/DeclarativeInput'
       responses:
         '400':
           description: bad request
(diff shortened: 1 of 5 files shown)


grahamc:
I'm not keen on the descriptions here, maybe they could be clarified to say they're just like JobsetInputs?

src/lib/Hydra/Schema/Projects.pm
```diff
@@ -252,7 +252,10 @@ my %hint = (
         "displayname",
         "description",
         "homepage",
-        "owner"
+        "owner",
+        "declfile",
+        "decltype",
+        "declvalue"
```
grahamc, line 258: It would be a bigger lift, but it might be nicer if the declarative configuration here were a JobsetInput type, and thus a nested value here. What do you think?

cole-h: It would have to be its own type because JobsetInput has an `emailresponsible` field. I created its own type in 43cbe6693e7dc39612999f193b103d87137710dc, but reverted it in later fixups (to see how both would look) -- thoughts?

cole-h:
Tests will fail because, for whatever reason, `GET /project/{id}` returns `18` as the content...?

hydra-api.yaml
```diff
@@ -577,12 +582,30 @@ components:
...
         jobsets:
           description: list of jobsets belonging to this project
           type: array
           items:
             type: string
 
+    DeclarativeInput:
+      type: object
+      properties:
+        name:
```
cole-h, line 599: TODO for me: This should be `file`.

grahamc (approved):
I tested the web portion by hand here, created a project, edited it a few times, validated it was updated in the database and rendered correctly. Since this is strictly an expansion of the API to support more of the data model, I don't think it warrants a discourse post.

PR #954, opened 2021-05-03
Projects: use JSON in Schema

cole-h:


grahamc:
Embarrassing that this happened just the other day and I didn't catch it this time. I don't know why GHA isn't catching it, I've pinged my friendly perlmonger.

cole-h:
*We* didn't catch it :P You even showed me what went wrong (and how to fix it) last time!

PR #961, opened 2021-05-05
lazy tabs: trigger the load event

### LINKED ISSUE #959: Direct links to tabs with dynamically loaded contents stopped working
Example: https://hydra.nixos.org/jobset/disnix/disnix-trunk#tabs-jobs

grahamc:
Something in the upgrade of Bootstrap and JQuery broke lazy tab loading.
I don't understand what is providing the tab behavior, how it should
work, or what the correct fix is.

I can tell you that this patch fixes the issue: when loading a tab
with a URL fragment deep-linking to a lazily loaded tab... it now
loads.

Close #959

### DIFF
--- a/src/root/static/js/common.js
+++ b/src/root/static/js/common.js
@@ -60,6 +60,7 @@ $(document).ready(function() {
 
     /* Activates tab according to URL anchor. */
     if (window.location.hash) {
+        setTimeout(function () { $('.nav-tabs > .nav-item:not(.dropdown) a[href="' + window.location.hash + '"]').trigger('show.bs.tab'); }, 0);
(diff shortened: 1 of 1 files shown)


samueldr:
AFAIUI, though I did not dig deep, *things probably got more async*, so what once worked because of synchronuous execution ordering, now doesn't.

My guess is that for a more "proper" fix, [in `makeLazyTab`](https://github.com/NixOS/hydra/blob/886e6f85e45a1f757e9b77d2a9e4539fbde29468/src/root/static/js/common.js#L135-L150) we should look at the URL to set the initial state. But IMO that's not less ugly than delaying the load through `setTimeout`, since that really mixes concerns. `makeLazyTab` shouldn't care about the state, only initialize things.

grahamc:
I changed the commit to delay 0ms which just pushes it out to the next event loop and seems to fix it too.

Issue #960, opened 2021-05-05
Changing a declarative project to a normal project doesn't delete the `.jobsets` jobset


cole-h:
I don't see any reason this "meta" jobset should hang around after the project is no longer declarative; it should be cleaned up automatically.

grahamc:
I agree, but I wonder what the other users thinks?

PR #962, opened 2021-05-05
Project: delete the `.jobsets` jobset if project is no longer declarative

cole-h:
"No longer declarative" as defined by the "Edit project" page is an empty spec
file.

---

Fixes https://github.com/NixOS/hydra/issues/960.

~~TODO: write a test~~

### DIFF
--- a/src/lib/Hydra/Controller/Project.pm
+++ b/src/lib/Hydra/Controller/Project.pm
@@ -171,6 +171,10 @@ sub updateProject {
             });
     } else {
         $project->jobsets->search({ name => ".jobsets" })->delete;
+        $project->update(
+            { decltype => ""
+            , declvalue => ""
+            });
     }
 }
 
--- a/src/lib/Hydra/Controller/Project.pm
+++ b/src/lib/Hydra/Controller/Project.pm
@@ -169,6 +169,8 @@ sub updateProject {
             , emailoverride => ""
             , triggertime => time
             });
+    } else {
+        $project->jobsets->search({ name => ".jobsets" })->delete;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/Hydra/Controller/Project.pm
```diff
@@ -169,6 +169,8 @@ sub updateProject {
             , emailoverride => ""
             , triggertime => time
             });
+    } else {
+        $project->jobsets->search({ name => ".jobsets" })->delete;
```
cole-h, line 173: Should we maybe also clear the `decltype` and `declvalue`?

cole-h: I opted to do so, but it can be easily reverted.

cole-h:
Ready for review, but NOT ready for merge (if clearing `decltype` and `declvalue` are deemed acceptable when going back to a normal jobset, 98633e75d65aad71e8b9621b6e9cb5e7a1cffc11 will need to be reworded to remove the question mark).

src/lib/Hydra/Controller/Project.pm
grahamc: I think this is a good idea.

t/Controller/projects.t
```diff
@@ -17,27 +17,133 @@ Catalyst::Test->import('Hydra');
...
+    Content => {
+        username => 'alice',
+        password => 'foobar'
     }
-});
+);
+is($req->code, 302);
+my $cookie = $req->header("set-cookie");
+
+subtest "Read project 'tests'" => sub {
```
grahamc, line 38: Thanks for moving these in to subtests.

grahamc (approved):
Confirmed it works locally on a declarative project which has no generated projects. I then created a declarative project which did create jobsets, emptied the file field, saved, and the `.jobsets` jobset was removed and other jobsets remained.

PR #964, opened 2021-05-10
Jobsets: remove defunct Jobs relationship

cole-h:
It appears the Jobs table was removed in
8adb433e3b48638f3faaecc61ae5cb63efefff6e, but the Jobsets schema was never
updated to reflect this. This relationship was added in
efa1f1d4fbdc468c61f7403e8709d0295e052c7f, roughly 3 months prior.

Previously, one would see a message similar to the following logged when
deleting a jobset:

    17:38:23 hydra-server.1       | DBIx::Class::Relationship::CascadeActions::delete(): Skipping cascade delete on relationship 'jobs' - related resultsource 'Hydra::Schema::Jobs' is not registered with this schema at /home/vin/workspace/vcs/hydra/src/script/../lib/Hydra/Controller/Jobset.pm line 106

### DIFF
--- a/src/lib/Hydra/Schema/Jobsets.pm
+++ b/src/lib/Hydra/Schema/Jobsets.pm
@@ -392,21 +392,6 @@ __PACKAGE__->has_many(
   undef,
 );
 
-=head2 jobs
-
-Type: has_many
-
-Related object: L<Hydra::Schema::Jobs>
-
-=cut
-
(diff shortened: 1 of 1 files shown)


grahamc (approved):
lgtm, pretty sure no code is pointing at this relation since ... well, it'd have blown up pretty obviously.

PR #965, opened 2021-05-10
Fit more content on screen

ajs124:
I'm not a web designer or frontend developer, but it felt like I spent more time scrolling after the recent library upgrade, so I modified the css a bit.

Not sure if this is of interest to other users, but here we go.

### DIFF
--- a/src/root/static/css/hydra.css
+++ b/src/root/static/css/hydra.css
@@ -1,5 +1,5 @@
 div.skip-topbar {
-    padding-top: 40px;
+    padding-top: 20px;
     margin-bottom: 1.5em;
 }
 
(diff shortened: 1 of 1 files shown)


dasJ:
This is what it looks like in my Browser:
image

dasJ:
Poke @grahamc what about this? The current width is kind of annoying and it doesn't seem like there will be another response on this PR

ajs124:
I know there are other priorities, right now, but if and when anyone comes back to this, here are some more comparisons, because I'm honestly annoyed, every time I look at hydra.nixos.org:
hydra.nixos.org: image
hydra.helsinki.tools: image
hydra.nixos.org: image
hydra.helsinki.tools: image
hydra.nixos.org: image
hydra.helsinki.tools: image

PR #967, opened 2021-05-24
Set `isSystemUser = true;` for each statically declared user

Ma27:
This is needed to evaluate the module on 21.05.
cc @grahamc


Jccp1983 (approved):
302f5e0de9a53d6dc73aea90851b54dc3ff29055

PR #974, opened 2021-06-17
flake: stdenv.lib -> lib

AmandaCameron:
Nixpkgs on unstable has removed `stdenv.lib` as they've been warning for a while now. This removes the extra references to it in the flake.nix

I'm not entirely sure if this is right, but I figured it was trivial enough to give a quick try using the GH Editor while I was waiting for a job to finish


grahamc (approved):
Thanks!

PR #976, opened 2021-06-19
GitInput: Include deepClone option in the cache key

### LINKED ISSUE #510: deepClone option in gitInput troubles
Consider two jobsets with the same git input, one without the `deepClone` option while the other one have the `deepClone` enabled.
If the first jobset is evaluated before the second one, the second one will not have the `.git` directory since it reuses the store path of the first evaluation.

To reproduce, create two jobset with gitinput such as:
```
input of jobset 1: git-repo master
input of jobset 2: git-repo master deepClone
```
Next, evaluate jobset1 before jobset2 and check the content of the clone directory.

ztzg:
Without this commit, two jobsets using the same repository as input, but different `deepClone` options, end up incorrectly sharing the same "checkout" for a given (`uri`, `branch`, `revision`) tuple.  The presence or absence of `.git` is determined by the jobset execution order.

This patch adds the missing `isDeepClone` boolean to the cache key.

The database upgrade script empties the `CachedGitInputs` table, as we don't know if existing checkouts are deep clones.  Unfortunately, this generally forces rebuilds even for correct `deepClone` checkouts, as the binary contents of `.git` are not deterministic.

Fixes #510

### DIFF
--- a/src/sql/upgrade-76.sql
+++ b/src/sql/upgrade-76.sql
@@ -1,0 +1,12 @@
+-- We don't know if existing checkouts are deep clones.  This will
+-- force a new fetch (and most likely trigger a new build for deep
+-- clones, as the binary contents of '.git' are not deterministic).
+DELETE FROM CachedGitInputs;
+
+ALTER TABLE CachedGitInputs
+    ADD COLUMN isDeepClone BOOLEAN NOT NULL;
+
+ALTER TABLE CachedGitInputs DROP CONSTRAINT cachedgitinputs_pkey;
+
+ALTER TABLE CachedGitInputs ADD CONSTRAINT cachedgitinputs_pkey
+    PRIMARY KEY (uri, branch, revision, isDeepClone);
--- a/src/lib/Hydra/Plugin/GitInput.pm
+++ b/src/lib/Hydra/Plugin/GitInput.pm
@@ -182,7 +182,7 @@ sub fetchInput {
     # TODO: Fix case where the branch is reset to a previous commit.
     my $cachedInput;
     ($cachedInput) = $self->{db}->resultset('CachedGitInputs')->search(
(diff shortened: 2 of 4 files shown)


src/sql/upgrade-76.sql
```diff
@@ -0,0 +1,12 @@
+-- We don't know if existing checkouts are deep clones.  This will
+-- force a new fetch (and most likely trigger a new build for deep
+-- clones, as the binary contents of '.git' are not deterministic).
+DELETE FROM CachedGitInputs;
```
grahamc, line 4: This freaked me out a bit but we already do this often: https://github.com/NixOS/hydra/blob/04fdb8e6e66d68d2625bdefa7f1c9d99e812c15b/src/lib/Hydra/Controller/Admin.pm#L51-L59

grahamc (approved):
lgtm, but I'd like another set of eyes on it.

I applied this diff to hydra.nixos.org's database and the migration went smoothly, nice.

cole-h (approved):
Diff LGTM as well.

ztzg:
Hi Graham,

>> @@ -0,0 +1,12 @@
> +-- We don't know if existing checkouts are deep clones.  This will
> +-- force a new fetch (and most likely trigger a new build for deep
> +-- clones, as the binary contents of '.git' are not deterministic).
> +DELETE FROM CachedGitInputs;
>
> This freaked me out a bit but we already do this often:
> https://github.com/NixOS/hydra/blob/04fdb8e6e66d68d2625bdefa7f1c9d99e812c15b/src/lib/Hydra/Controller/Admin.pm#L51-L59

Right: that tidbit is what convinced me that it was okay to clear the
table; I should have referenced it from the comment.

> Merged #976 into master.

Thanks!

Cheers, -D

PR #981, opened 2021-06-28
Allow for evaluator_pure_eval to be configured

jonringer:
In a similar vein to #888 allow users to configure purity.

The main reason for this is to support legacy nix code which may call things like `builtins.currentSystem`. With large highly coupled code bases, it's often the case that there is a lot of IFD and impurity going on...


grahamc:
I'm surprised you merged this, @edolstra  -- it seems counter to your position on purity and Hydra. Have you had a change of heart on the role of Hydra?

Issue #978, opened 2021-06-24
Hydra-unstable broken on master. Need lipqxx>=7


jonringer:
https://github.com/NixOS/nixpkgs/pull/126996 bumped `libqpxx` to 7.5.2 which breaks the build.

Why is this an issue for NixOS/hydra? 

The fixes don't look terrible to fix code side. And the PR removes one of the few remaining usages of python2.
I can pin another version of libpqxx until the fix lands here, but we should probably move to libpqxx anyway.

Error:

```
g++ -DHAVE_CONFIG_H -I. -I../..    -std=c++17 -I/nix/store/by2bh6wawngyxw4rha6l8yy5slq7pplc-boehm-gc-8.0.4-dev/include -I/nix/store/28c5xg650ig6yn2gc26hvra1d76fdjc4-nix-2.4pre20210601_5985b8b-dev/include/nix -Wall -I ../libhydra -Wno-deprecated-declarations -g -O2 -std=c++17 -include nix/config.h -c -o hydra_evaluator-hydra-evaluator.o `test -f 'hydra-evaluator.cc' || echo './'`hydra-evaluator.cc
In file included from /nix/store/0p9hqklkykhrig45y8r5iw8ysw4q83vk-libpqxx-7.5.2/include/pqxx/util.hxx:31,
                 from /nix/store/0p9hqklkykhrig45y8r5iw8ysw4q83vk-libpqxx-7.5.2/include/pqxx/result.hxx:25,
                 from /nix/store/0p9hqklkykhrig45y8r5iw8ysw4q83vk-libpqxx-7.5.2/include/pqxx/binarystring.hxx:21,
                 from /nix/store/0p9hqklkykhrig45y8r5iw8ysw4q83vk-libpqxx-7.5.2/include/pqxx/binarystring:4,
                 from /nix/store/0p9hqklkykhrig45y8r5iw8ysw4q83vk-libpqxx-7.5.2/include/pqxx/pqxx:3,
                 from ../libhydra/db.hh:3,
                 from hydra-evaluator.cc:1:
../../version:1:1: error: expected unqualified-id before numeric constant
    1 | 0.1
      | ^~~
In file included from /nix/store/0p9hqklkykhrig45y8r5iw8ysw4q83vk-libpqxx-7.5.2/include/pqxx/util.hxx:37,
                 from /nix/store/0p9hqklkykhrig45y8r5iw8ysw4q83vk-libpqxx-7.5.2/include/pqxx/result.hxx:25,
[...]

PR #983, opened 2021-07-05
Fix build with libpqxx 7

Taneb:
This compiles with libpqxx HEAD but due to a known error does not work with the latest released version, 7.5. It also works with the 6.X version currently in use.

This does all the work on this side for #978 but I don't think that should be closed until libpqxx 7.6 is released and we can fix Hydra in nixpkgs


grahamc:
Thanks!

PR #984, opened 2021-07-19
Jobset: test `type`-related constraints

cole-h:
Prior to https://github.com/NixOS/hydra/pull/856, the `type`-related constraints
were a bit too loose, allowing some variations that should have been seen as
invalid. This commit adds a test (inspired by
https://github.com/NixOS/hydra/pull/856#issuecomment-772820929) to ensure these
constraints stay unviolated.

### DIFF
--- a/t/Controller/Jobset/type-constraints.t
+++ b/t/Controller/Jobset/type-constraints.t
@@ -1,0 +1,122 @@
+use strict;
+use Setup;
+my %ctx = test_init();
+
+require Hydra::Schema;
+require Hydra::Model::DB;
+require Hydra::Helper::Nix;
+
+use Data::Dumper;
+use Test2::V0;
+use Test2::Compare qw(compare strict_convert);
+
+my $db = Hydra::Model::DB->new;
+hydra_setup($db);
+
+# This test checks a matrix of jobset configuration options for constraint violations.
+
+my @types = ( 0, 1, 2 );
+my @nixexprinputs = ( undef, "input" );
+my @nixexprpaths = ( undef, "path" );
+my @flakes = ( undef, "flake" );
+
+my @expected_failing;
+my @expected_succeeding = (
+    {
+        "name" => "test",
+        "emailoverride" => "",
+        "type" => 0,
+        "nixexprinput" => "input",
+        "nixexprpath" => "path",
+        "flake" => undef,
+    },
+    {
+        "name" => "test",
+        "emailoverride" => "",
+        "type" => 1,
+        "nixexprinput" => undef,
+        "nixexprpath" => undef,
+        "flake" => "flake",
+    },
+);
+
+# Checks if two Perl hashes (in scalar context) contain the same data.
+# Returns 0 if they are different and 1 if they are the same.
+sub test_scenario_matches {
+    my ($first, $second) = @_;
+
+    my $ret = compare($first, $second, \&strict_convert);
+
+    if (defined $ret == 1) {
+        return 0;
+    } else {
+        return 1;
+    }
+}
+
+# Construct a matrix of parameters that should violate the Jobsets table's constraints.
+foreach my $type (@types) {
+    foreach my $nixexprinput (@nixexprinputs) {
+        foreach my $nixexprpath (@nixexprpaths) {
+            foreach my $flake (@flakes) {
+                my $hash = {
+                    "name" => "test",
+                    "emailoverride" => "",
+                    "type" => $type,
+                    "nixexprinput" => $nixexprinput,
+                    "nixexprpath" => $nixexprpath,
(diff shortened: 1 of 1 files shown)


t/Controller/Jobset/constraints.t
```diff
@@ -0,0 +1,165 @@
...
+            foreach my $flake (@flakes) {
+                my $hash = {
+                    "jobset_type" => $jobset_type,
+                    "nixexprinput" => $nixexprinput,
+                    "nixexprpath" => $nixexprpath,
+                    "flake" => $flake,
+                };
+
+                push(@expected_failing, $hash)
+                    if (compare_hashes($hash, $expected_succeeding[0]) && compare_hashes($hash, $expected_succeeding[1]));
```
grahamc, line 80: How about turning this in to a foreach loop, so if there are additional cases later this bit of code doesn't need to be updated?

cole-h:
Converted to draft because I want to investigate how to validate this against the actual database / table, rather than one created solely for this test.

t/Controller/Jobset/type-constraints.t
```diff
@@ -0,0 +1,128 @@
...
+my %ctx = test_init();
+
+require Hydra::Schema;
+require Hydra::Model::DB;
+require Hydra::Helper::Nix;
+
+use Test2::V0;
+require Catalyst::Test;
+use HTTP::Request::Common;
+Catalyst::Test->import('Hydra');
```
grahamc, line 13: I think you probably don't need to import these.

t/Controller/Jobset/type-constraints.t
```diff
@@ -0,0 +1,128 @@
...
+hydra_setup($db);
+
+# This test checks a matrix of jobset configuration options for constraint violations.
+
+my @types = ( 0, 1, 2 );
+my @nixexprinputs = ( undef, "input" );
+my @nixexprpaths = ( undef, "path" );
+my @flakes = ( undef, "flake" );
+my @emailoverrides = ( undef, "" );
+my @names = ( undef, "test" );
```
grahamc, line 25: Looks like a couple of dead variables.

t/Controller/Jobset/type-constraints.t
```diff
@@ -0,0 +1,128 @@
...
+                # Necessary, otherwise cases will fail because the `->create`
+                # will throw an exception due to an expected constraint failure
+                # (which will cause the `ok()` to be skipped, leading to no
+                # assertions in the subtest).
+                is(1, 1);
+
+                ok(
+                    $project->jobsets->create($case),
+                    "Expected jobset to violate constraints"
+                );
```
grahamc, line 98: Is the `ok` here doing something important? Is the `ok` the part that is failing, or is the `create` call dying?

t/Controller/Jobset/type-constraints.t
```diff
@@ -0,0 +1,128 @@
...
+        $count++;
+    };
+};
+
+# Validate that the list of parameters that should not fail the constraints do indeed succeed.
+subtest "Expected constraint successes" => sub {
+    my $count = 1;
+    foreach my $case (@expected_succeeding) {
+        subtest "Case $count" => sub {
+            lives {
```
grahamc, line 111: Do we need the `lives` block here? Not saying we don't, genuinely curious.

grahamc (commented):
Looks pretty good. Just a few nits inline.

Two other things:

1. How about using Data::Dumper to print the test scenario data in to the `subtest` name?
1. Can you fill in the PR body, maybe by filling out the commit body as well, with a bit about the goal of this PR, and a reference to https://github.com/NixOS/hydra/pull/856 and in particular https://github.com/NixOS/hydra/pull/856#issuecomment-772820929?

t/Controller/Jobset/type-constraints.t
cole-h: The `create` should have been prefixed with a `!` to invert the case: if, for whatever reason, `create` doesn't throw an exception, the test should not pass (because we want every test here to have failed).

t/Controller/Jobset/type-constraints.t
cole-h: Nope -- I just liked the symmetry :P Will remove.

t/Controller/Jobset/type-constraints.t
cole-h: Hehe, oops.

t/Controller/Jobset/type-constraints.t
cole-h: I had copied over the HTTP API test as scaffolding initially :sweat_smile:

grahamc (approved):
squash then looks good!

PR #985, opened 2021-07-22
Make hydra-server honor apache-style includes in hydra.conf like all the other components

kreisys:
After learning that I can `Include` from `hydra.conf` I tried to leverage that to get some secrets that shouldn't be there into `/run/keys`. I then noticed that while GitHub commit status reporting worked just fine, `Sign in with Google` did not. It took a few hours of head-scratching followed by a perl deep(-ish)-dive to figure out what's going on...

In short(...ish, for future reference):
- `hydra-server` loads `hydra.conf` via `Catalyst::Plugin::ConfigLoader` which in turn uses its `General` driver to drive `Config::General` module which finally parses `hydra.conf`. `Config::General` is also used directly by `Helpers::Nix` to load `hydra.conf`.
- Everything else uses the `getHydraConfig` sub in `Helpers::Nix`, which uses `Config::General` directly but passes it the options `-UseApacheInclude =>  1, -IncludeAgain => 1`, thereby making it respect `Include` statements in `hydra.conf`.
- __GitHub commit status reporting__ happens in `hydra-notify`, which uses `Helpers::Nix::getHydraConfig()` so everything was fine there.
- __Sign in with Google__ happens in `hydra-server`, which loads `hydra.conf` via `Catalyst::Plugin::ConfigLoader` with no special parameters, so no go there.

Confusing. Luckily, the only example in the `ConfigLoader` cpan docs covered precisely what I needed so I only had to copy and paste https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Helper/Nix.pm#L46-L47 into, well, the few lines that are around it in the diff...

Aight that's already far more prose than code kthxbye 😌


src/lib/Hydra.pm
```diff
@@ -48,6 +48,14 @@ __PACKAGE__->config(
             file($ENV{'HYDRA_LDAP_CONFIG'})
         ) : undef
     },
+    'Plugin::ConfigLoader' => {
+        driver => {
+            'General' => {
```
grahamc, line 53: Pretty good! Could you add a comment in this file, around this line, to also update https://github.com/NixOS/hydra/blob/2f0dc294afc512bd3db1c3476b8d46954214af53/src/lib/Hydra/Helper/Nix.pm#L45-L48 if these General options change, and also edit https://github.com/NixOS/hydra/blob/2f0dc294afc512bd3db1c3476b8d46954214af53/src/lib/Hydra/Helper/Nix.pm#L45-L48 to say to edit this files' options to match? This way future changes don't get out of sync again.

grahamc: PS I don't suppose you would be up for writing a bit of documentation about this?

`foreman start manual` will run a live-reloading docs webserver. Maybe a snippet in https://github.com/NixOS/hydra/blob/master/doc/manual/src/installation.md, or a new configuration.md file.

Note if you add a new file (I sort of prefer this...) you'll need to add it to the SUMMARY: https://github.com/NixOS/hydra/blob/master/doc/manual/src/SUMMARY.md

kreisys: Done. I'm segfaulting on whether factoring out two lines is worth a new file.

kreisys: ... or we can export it from `Nix.pm`. I'll look into it in the morning. I'm perled-out for the day.

doc/manual/src/configuartion.md
```diff
@@ -0,0 +1,130 @@
...
+```
+should **NOT** be in `hydra.conf`.
+
+`hydra.conf` is rendered in the Nix store and is therefore world-readable.
+
+Instead, the above should be written to a file outside the Nix store by
+other means (manually, using Nixops' secrets feature, etc) and included
+like so:
+```
+Include github_authorizations.conf
```
grahamc, line 28:
```suggestion
Include /run/keys/hydra_github_authorizations.conf
```

grahamc (approved):
Looks great! I'm leaving this open for a bit in case you want to factor out those two lines, and also to ask if my suggestion to incuding the keys makes sense to you.

doc/manual/src/configuartion.md
kreisys: Oh I haven't been able to use absolute paths. It tries to append them to `/var/lib/hydra`. That's why I mentioned symlinks as a workaround. I haven't checked relative paths tbh. I'll expand the example to demonstrate that.

Regarding the config refactor, from this side of the night it looks to me like `getHydraConfig` doesn't really belong (at least semantically) in `Helpers::Nix`. How about splitting that to a `Helpers::Config` and expose a sub that just returns the `Config::General` options? That would make it more akin to a semi-real refactor. Though perhaps it increases the risk of breaking something so you tell me 😌

grahamc: Oh dear, oops, I guess I missed the part about symlinks. I'm glad I waited for your review!

I would be comfortable with you doing that refactor for sure. Would you like this PR to merge first? I'd say this is a notable improvement in docs and behavior already, and I don't want you to feel like this is becoming annoying gatekeeping.

kreisys:
So trying to factor out more than necessary was a mistake. The resulting `Config.pm` which included a sizable chunk of `Helper/Nix.pm`, would cause `make check` fail job 13. So I backtracked on that and instead created a `Config.pm` that only exports the `Config::General` options and nothing more. Seems to work fine, but I want to double check with what I was working on when I needed this. I'll add a working example to the doc based on my final, known-to-work config.

Also you might want to have a look at the other available options for [Config::General](https://metacpan.org/pod/Config::General). There are some options there that might be useful. Specifically, in the includes department, I think this might be important for this PR:

> ### `-IncludeRelative`
> If set to a true value, included files with a relative path (i.e. "cfg/blah.conf") will be opened from within the location of the configfile instead from within the location of the script($0). This works only if the configfile has a absolute pathname (i.e. "/etc/main.conf").
> 
> If the variable -ConfigPath has been set and if the file to be included could not be found in the location relative to the current config file, the module will search within -ConfigPath for the file. See the description of -ConfigPath for more details.

The hydra components all use the same `hydra.conf` but their systemd services have different `WorkingDirectory` so it might get confusing if we don't add it...

kreisys:
PS is there a way to run just one test?

grahamc:
> PS is there a way to run just one test?

Yep: `yath test ./t/foo/bar.t`. I also like to use `-v` sometimes.

grahamc:
This looks pretty good to me right now, but I've asked a friendly perlmonger to review it for perliness. Thanks!

src/lib/Hydra/Config.pm
```diff
@@ -0,0 +1,11 @@
+package Hydra::Config;
+
+use strict;
+use Exporter;
+
+our @ISA = qw(Exporter);
+our @EXPORT = qw(%configGeneralOpts);
+
+my %configGeneralOpts = (-UseApacheInclude => 1, -IncludeAgain => 1);
```
grahamc, line 9: my perl friend suggests:

```suggestion

our %generalOptions = (-UseApacheInclude => 1, -IncludeAgain => 1);
```

then in other places you can refer to %Hydra::Config::generalOptions`

also referred to https://metacpan.org/pod/Exporter#What-Not-to-Export

kreisys: Do I still need to `use Hydra::Config;`?

grahamc: I think yes, but worth an experiment.

kreisys: yeah no need to `use`. also it looks like adding `-IncludeRelative` made absolute paths work so I updated the doc.
also added a rudimentary-ish test that verifies nested includes work.
I think we're good.

max-privatevoid:
Just for reference, I believe `Config::General` unconditionally (and thus Hydra even before this gets merged) supports includes when using this syntax:
```
<<include /run/secrets/secrets.conf>>
```

kreisys:
> Just for reference, I believe `Config::General` unconditionally (and thus Hydra even before this gets merged) supports includes when using this syntax:
> 
> ```
> <<include /run/secrets/secrets.conf>>
> ```

Yeah, I also figured that out somewhere amidst my adventures... You think it's worth noting in the doc?

kreisys:
@grahamc are we waiting for anything on my end?

grahamc (approved):

grahamc:
Looks great from here! Let's run the tests once more and verify. Thanks for the docs and test work :).

PR #995, opened 2021-08-12
Declarative jobsets: move event handling to a plugin

grahamc:
Declarative jobsets were sort of tucked in to the event handler
itself. It turned out that it could have been implemented as a
plugin without much trouble.

Declarative jobsets have no tests, so this should be tested
by hand.


grahamc:
Just fixup'd a missing import.

cole-h (approved):
Diff looks sane to me.

grahamc:
@dasJ gave a :+1: on testing on Matrix.

PR #996, opened 2021-08-12
hydra-notify: move buildFinished query in to the function impl

grahamc:
This is more consistent with the other event handlers, of dealing
with IDs and not objects.


cole-h (approved):
Diff looks sane to me.

PR #997, opened 2021-08-12
Abstract over postgres' LISTEN/NOTIFY

grahamc:
This lets us test the event loop if we wanted, and lets us
test the listening behavior in isolation.


cole-h (approved):
Diff LGTM, ran the test locally and it passed!

Especially liking the test :heart_eyes:

PR #1001, opened 2021-08-18
hydra-notify: export prometheus metrics

grahamc:


src/script/hydra-notify
```diff
@@ -3,24 +3,91 @@
...
+    my $fork_manager = Parallel::ForkManager->new(1);
+    $fork_manager->start_child("metrics_exporter", sub {
+        my $server = HTTP::Server::PSGI->new(
+            host => $promCfg->{"listen_address"},
+            port => $promCfg->{"port"},
+            timeout => 1,
+        );
+
+        $server->run($prom->psgi);
+    });
```
grahamc, line 80: This is a pretty ugly thing. In order to export metrics we basically need a background thread / process serving the data. Perl's threads are "do not use" so here we are. We're using Prometheus::Tiny::Shared which uses a mmap'd file to share the state from the event loop process with the exporter process. This is doubly not great as every write is a json encode, and every read is a json decode. However, again, here we are.

src/script/hydra-notify
```diff
@@ -3,24 +3,91 @@
...
+use Time::HiRes qw( gettimeofday tv_interval );
 
 STDERR->autoflush(1);
 STDOUT->autoflush(1);
 binmode STDERR, ":encoding(utf8)";
 
+my $config = getHydraConfig();
+
+my $prom = Prometheus::Tiny::Shared->new;
+# Note: It is very important to pre-declare any metrics before using them.
```
cole-h, line 24: Maybe explain why it's important? Though I personally would acquiesce, I could imagine somebody who might ignore this because there's no reason given.

grahamc: :+1: thanks!

src/lib/Hydra/Helper/Nix.pm
```diff
@@ -70,6 +70,29 @@ sub getStatsdConfig {
...
+sub getHydraNotifyPrometheusConfig {
+    my ($config) = @_;
+    my $cfg = $config->{hydra_notify};
+
+    if (!defined($cfg) || ref $cfg ne "HASH") {
+        return undef;
+    }
+
+    my $cfg = $cfg->{prometheus};
+    if (!defined($cfg) || ref $cfg ne "HASH") {
```
andir, line 82: Shouldn't this log an error/warning if the config isn't a hash? That smells like a user error.

src/script/hydra-notify
andir: if this is as gross as you describe: Couldn't be write a text file with the metrics instead? Any webserver is able to serve that and the prometheus node exporter is able to pick up text files.

grahamc: Good question. `Net::Prometheus` (an objectively better Prometheus implementation) supports `->render`ing to text, which we could write out to a file.

However:

1. Writing to a text file means we don't get valid `up` metrics: if hydra-notify crashes we're still getting metrics as if things are fine. This could be resolved by having the second process serve it.
2. The metrics in the file are only as up to date as the most recent write. If we don't write out to the file on every metric update, we lose valuable hints as to what is going on.

I didn't see a way to have Net::Prometheus write out on every update, but perhaps that could be done?

What do you think?

src/lib/Hydra/Helper/Nix.pm
grahamc: My kingdom for serde.

grahamc: How about this?

```perl
sub getHydraNotifyPrometheusConfig {
    my ($config) = @_;
    my $cfg = $config->{hydra_notify};

    if (!defined($cfg)) {
        return undef;
    }

    if (ref $cfg ne "HASH") {
        print STDERR "Error reading Hydra's configuration file: hydra_notify should be a block.\n";
        return undef;
    }

    my $cfg = $cfg->{prometheus};
    if (!defined($cfg)) {
        return undef;
    }

    if (ref $cfg ne "HASH") {
        print STDERR "Error reading Hydra's configuration file: hydra_notify.prometheus should be a block.\n";
        return undef;
    }

    if (defined($cfg->{"listen_address"}) && defined($cfg->{"port"})) {
        return {
            "listen_address" => $cfg->{'listen_address'},
            "port" => $cfg->{'port'},
        };
    } else {
        print STDERR "Error reading Hydra's configuration file: hydra_notify.prometheus should include listen_address and port.\n";
        return undef;
    }

    return undef;
}
```

src/script/hydra-notify
andir: Regarding the 1. point: We could always add a timestamp to the file.
2. How many writes / second are we talking here? Writing out 9 metrics sounds relatively cheap but we will have to do it atomically (so new file and then move it).

To make that work we could probably wrap our usage of the library such that it writes to a file whenever we set a metric? Another option would be to update the file after each time we try to execute a notification?

PR #1014, opened 2021-08-28
gitea-plugin: fix test

Ma27:
The test seems to be broken for a while[1]. The cause for this is that
in gitea 1.14 the `create-user` command got renamed to `user create`.

[1] https://hydra.nixos.org/build/151092299

cc @grahamc


grahamc:
Thanks, looks good!

PR #1017, opened 2021-09-08
jobset-eval: pass `full` parameter to "Compare to..."

Ma27:
When I take a look at *all* failing builds (by clicking at `[...] more
jobs omitted`) and I try to compare the failures to another jobset, I'd
like to still view *all* failing builds in the compare-view.

This wasn't the case before since the `full=`-param was ignored by the
compare-buttons.

cc @grahamc


grahamc:
Nice catch!

PR #1025, opened 2021-09-22
Produce better errors for failing jobsets

lukegb:
Recently nixpkgs had a problem where evaluations were failing with the dreaded `[json.exception.type_error.302] type must be string, but is null` nlohmann/json error.

It turns out this is because some of the NixOS tests included in the `tested` job could no longer be evaluated.

No more! With this change, not only will we produce _an_ evaluation, but we'll also generate more useful evaluation errors for people to use when debugging, rather than trying to manually bisect nixpkgs to work out at what point evaluations broke.

More context in the commits included in this PR. It even includes a test (which appears to be the first test of the aggregate job functionality, and I'm... asserting that it breaks correctly. Oh well.)


src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -440,11 +440,21 @@ int main(int argc, char * * argv)
...
+                        printError("aggregate job '%s' references non-existent job '%s'", jobName, jobName2);
+                        brokenJobs[jobName2] = "does not exist";
+                        continue;
+                    }
+                    if (job2->find("error") != job2->end()) {
+                        std::string error = (*job2)["error"];
+                        printError("aggregate job '%s' references broken job '%s': %s", jobName, jobName2, error);
+                        brokenJobs[jobName2] = error;
+                        continue;
+                    }
```
grahamc, line 457: Would it be possible to make this its own function? I'm not sure it wouldbe straight forward to do so, but if it were ...

grahamc (approved):
Looks good to me, thank you. Let me know about that question and I'll happily merge.

PR #1044, opened 2021-10-20
Perlcritic level 3

grahamc:
WIP until I can fix remaining two-arg opens.


src/lib/Hydra/Helper/Nix.pm
```diff
@@ -141,7 +141,7 @@ sub registerRoot {
     my ($path) = @_;
     my $link = gcRootFor $path;
     return if -e $link;
-    open my $root, ">$link" or die "cannot create GC root `$link' to `$path'";
+    open(my $root, ">", $link) or die "cannot create GC root `$link' to `$path'";
```
sternenseemann, line 144: This is the only `open` call using parens?

PR #1062, opened 2021-11-19
RunCommand: store a log of execution and render it on the Build page

grahamc:
- [x] Don't display the tab if there are no RunCommand  tasks configured

I'm probably not going to do these since I'm not excited to add an opt-in-feature for this and support that ~forever:

- [ ] Make displaying the log data optional and default off in case someone has secrets in the commands
- [ ] Possibly make accessing the log require logging in

Includes:

- #1073 

image

### DIFF
--- a/src/lib/Hydra/Controller/Build.pm
+++ b/src/lib/Hydra/Controller/Build.pm
@@ -37,6 +37,7 @@ sub buildChain :Chained('/') :PathPart('build') :CaptureArgs(1) {
     $c->stash->{project} = $c->stash->{build}->project;
     $c->stash->{jobset} = $c->stash->{build}->jobset;
     $c->stash->{job} = $c->stash->{build}->job;
+    $c->stash->{runcommandlogs} = [$c->stash->{build}->runcommandlogs->search({}, {order_by => ["id DESC"]})];
 }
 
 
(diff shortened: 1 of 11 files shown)


SuperSandro2000:
>  Make displaying the log data optional and default off in case someone has secrets in the commands

Would it be feasible to filter out secrets like GitLab does it=

cole-h:
> > Make displaying the log data optional and default off in case someone has secrets in the commands
> 
> Would it be feasible to filter out secrets like GitLab does it?

I imagine one issue would be: how do you decide what's a secret? There's no magic regex or state machine that can catch them all.

grahamc:
There are some libraries to find high entropy strings: https://github.com/tehryanx/entro.py which might be useful, but I worry about people using RunCommand today and then one day they upgrade and stuff which has never been public is suddenly public.

dasJ:
A question mark for builds  that are not finished yet would be nice to keep consistency with nix build

grahamc:
@dasJ regarding the `?` I had that at one point but a UX developer I've worked with in the past advised against it to avoid visual noise.

grahamc:
I'm probably not going to do these:

> Make displaying the log data optional and default off in case someone has secrets in the commands
> Possibly make accessing the log require logging in

unless someone comes asking for it.

PR #1065, opened 2021-11-22
project:jobset:job: make clickable links in titles

grahamc:
image

image

image

### DIFF
--- a/src/root/build.tt
+++ b/src/root/build.tt
@@ -1,4 +1,7 @@
-[% WRAPPER layout.tt title="Build $id of job $project.name:$jobset.name:$job" %]
+[% WRAPPER layout.tt
+  title="Build $id of job " _ makeNameTextForJob(jobset, job)
+  titleHTML="Build $id of job " _ linkToJob(jobset, job)
+%]
 [% PROCESS common.tt %]
 [% PROCESS "product-list.tt" %]
 [% USE HTML %]
--- a/src/root/job.tt
+++ b/src/root/job.tt
@@ -1,5 +1,6 @@
 [% WRAPPER layout.tt
-    title="Job $project.name:$jobset.name:$job"
+    title=makeNameTextForJob(jobset, job)
+    titleHTML=makeNameLinksForJob(jobset, job)
     starUri=c.uri_for(c.controller('Job').action_for('star'), c.req.captures)
 %]
 [% PROCESS common.tt %]
--- a/src/root/layout.tt
+++ b/src/root/layout.tt
@@ -81,7 +81,8 @@
 
       [% IF !hideHeader %]
         <div class="page-header">
-          [% IF c.user_exists && starUri; INCLUDE makeStar; " "; END; HTML.escape(title) %]
+          [% IF c.user_exists && starUri; INCLUDE makeStar; " "; END %]
+          [% IF titleHTML != ""; titleHTML; ELSE; HTML.escape(title); END %]
         </div>
       [% ELSE %]
         [% IF first %]<br />[% first = 0; END; %]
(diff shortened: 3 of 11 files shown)


t/View/TT.t
```diff
@@ -0,0 +1,77 @@
+use feature 'unicode_strings';
```
cole-h, line 1: For some reason, the full name of this file just tickles me.

t/View/TT.t
```diff
@@ -0,0 +1,77 @@
...
+
+is(
+    Hydra::View::TT::linkToProject(undef, $c, $project),
+    '<a href="http://localhost/project/tests">tests</a>',
+    "linkToProject"
+);
+is(
+    Hydra::View::TT::linkToJobset(undef, $c, $jobset),
+    '<a href="http://localhost/project/tests">tests</a>:'
+    . '<a href="http://localhost/jobset/tests/example">example</a>',
```
cole-h, line 41: If I understand the implementation correctly, it looks like the various `linkTo*` subs create the link with the colon leading -- I think it would make sense to also follow this convention here:

```suggestion
    '<a href="http://localhost/project/tests">tests</a>'
    . ':<a href="http://localhost/jobset/tests/example">example</a>',
```

And in all the following `is` tests.

src/root/build.tt
```diff
@@ -1,4 +1,7 @@
-[% WRAPPER layout.tt title="Build $id of job $project.name:$jobset.name:$job" %]
+[% WRAPPER layout.tt
+  title="Build $id of job " _ makeNameTextForJob(jobset, job)
+  titleHTML="Build $id of job " _ linkToJob(jobset, job)
+   %]
```
samueldr, line 4:
```suggestion
%]
```

Or is it wanted, if so why?

src/root/log.tt
```diff
@@ -1,4 +1,7 @@
-[% WRAPPER layout.tt title="Log of " _ (step ? " step $step.stepnr of " : "") _ "build ${build.id} of job $build.project.name:$build.jobset.name:$build.job.name" %]
+[% WRAPPER layout.tt
+  titleHTML="Log of " _ (step ? " step $step.stepnr of " : "") _ "build ${build.id} of job " _ linkToJob(build.jobset, job)
+  title="Log of " _ (step ? " step $step.stepnr of " : "") _ "build ${build.id} of job " _ makeNameTextForJob(build.jobset, job)
+  %]
```
samueldr, line 4:
```suggestion
%]
```

Similarly, unless it is desired.

src/root/layout.tt
```diff
@@ -81,7 +81,8 @@
 
       [% IF !hideHeader %]
         <div class="page-header">
-          [% IF c.user_exists && starUri; INCLUDE makeStar; " "; END; HTML.escape(title) %]
+          [% IF c.user_exists && starUri; INCLUDE makeStar; " "; END %]
+          [% IF titleHTML != ""; titleHTML; ELSE; HTML.escape(title); END %]
```
samueldr, line 85: I know it's "technically better" to do it that way, but in an entirely different context than perl-based template, I previously used an HTML sanitizer to "stringify" HTML strings into string titles. In case this can spark an idea for reducing exhausting repetitiveness.

src/root/job.tt
```diff
@@ -1,5 +1,5 @@
 [% WRAPPER layout.tt
-    title="Job $project.name:$jobset.name:$job"
+    titleHTML=makeNameLinksForJob(jobset, job)
```
samueldr, line 2: Isn't this now missing a text-only title?

src/root/layout.tt
grahamc: Unfortunately, I don't think this is possible with our current tools. I like the idea though :).

src/root/job.tt
grahamc: :+1: fixup'd thanks!

src/root/build.tt
grahamc: Thanks, fixed, good catch!

PR #1068, opened 2021-11-30
Fix some warnings

cole-h:
It's always bothered me, seeing warnings pop up while running Hydra locally (while testing something or another). So, I fixed a lot of them here. Some fixes may need extra attention, though.

---

TODO:
- [x] add a test for the `narinfo` thing

### DIFF
--- a/src/lib/Hydra.pm
+++ b/src/lib/Hydra.pm
@@ -8,7 +8,6 @@ use Hydra::Plugin;
 use Hydra::Model::DB;
 use Catalyst::Runtime '5.70';
 use Catalyst qw/ConfigLoader
-                Unicode::Encoding
                 Static::Simple
                 StackTrace
                 Authentication
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -15,6 +15,9 @@ use JSON;
 use List::Util qw[min max];
 use List::MoreUtils qw{any};
 use Net::Prometheus;
+use Types::Standard qw/StrMatch/;
+
+use constant NARINFO_REGEX => qr{^([a-z0-9]{32})\.narinfo$};
 
 # Put this controller at top-level.
 __PACKAGE__->config->{namespace} = '';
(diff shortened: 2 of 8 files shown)


src/lib/Hydra.pm
```diff
@@ -8,7 +8,6 @@ use Hydra::Plugin;
 use Hydra::Model::DB;
 use Catalyst::Runtime '5.70';
 use Catalyst qw/ConfigLoader
-                Unicode::Encoding
```
cole-h, line 11: The warning notes that we should define some encoding config -- uh, what should that look like?

src/script/hydra-eval-jobset
```diff
@@ -644,7 +644,7 @@ sub checkJobsetWrapped {
     # Hash the arguments to hydra-eval-jobs and check the
     # JobsetInputHashes to see if the previous evaluation had the same
     # inputs.  If so, bail out.
-    my @args = ($jobset->nixexprinput, $jobset->nixexprpath, inputsToArgs($inputInfo));
+    my @args = ($jobset->nixexprinput // "", $jobset->nixexprpath // "", inputsToArgs($inputInfo));
```
cole-h, line 647: Should this maybe also take into account the flake-related arguments? Seems a bit strange to hash on things that are only defined for so-called "legacy" input types.

src/lib/Hydra/Controller/Root.pm
```diff
@@ -15,6 +15,9 @@ use JSON;
 use List::Util qw[min max];
 use List::MoreUtils qw{any};
 use Net::Prometheus;
+use Types::Standard qw/StrMatch/;
+
+use constant NARINFO_REGEX => qr{^([a-z0-9]+)\.narinfo$};
```
cole-h, line 20: Would be nice if we could make this match the length of the hash as well, but adding `{32}` made the server hang or something (in my brief testing).

src/lib/Hydra/Helper/CatalystUtils.pm
```diff
@@ -279,7 +279,7 @@ sub requirePost {
 
 
 sub trim {
-    my $s = shift;
+    my $s = shift // "";
```
grahamc, line 282: This would be a good opportunity to add a test case for this function, add a failing case for an uninitialized variable which should fail, and then be fixed by this test.

src/lib/Hydra/Controller/Root.pm
grahamc: Hmm... how did you try for it? I used: 

```
use constant NARINFO_REGEX => qr{^([a-z0-9]{32})\.narinfo$};
```

and it seemed good from here. If we wanted to be Very Precise we could use the exact character set of a store path: 0123456789abcdfghijklmnpqrsvwxyz but we probably don't need to be.

src/lib/Hydra.pm
grahamc: I think we're good: https://metacpan.org/pod/Catalyst#ENCODING

src/script/hydra-eval-jobset
grahamc: Honestly I don't know, so I'm hesitant to do so :grimacing:.

src/lib/Hydra/Controller/Root.pm
cole-h: Heh, whoops. I think I placed it OUTSIDE the capture group.

cole-h:
I added the narinfo test in dca27f637a5680b6786f59c2949aa4c1018f993c and made sure that it passed, and then I made the change removing `LocalRegex` in 7978bb607e35a05b0f5e206085f42fe63230c90e and validated that it still passed.

grahamc (approved):

PR #1069, opened 2021-12-01
TT.t: fixup $_, since _ is a special variable, not a throwaway

grahamc:


Ericson2314:
This is the most Perl bug to fix, ever, haha. Glad you spotted it!

PR #1071, opened 2021-12-08
hydra-eval-jobset: log fetches and evaluations

grahamc:
Sometimes evaluations will print errors about fetches. These can be hard to diagnose if you have many inputs. These log lines should help.


cole-h (approved):
Always a fan of more trace-like messages that aid debugging.

PR #1073, opened 2021-12-08
Runcommand: Expand the test coverage

grahamc:
My goal here is to make the RunCommand code that actually runs commands very simple, and make the policy choices (is the plugin enabled? does this runcommand match the job?) rest of it easy to test.

Each commit here passes tests individually and should be fairly simple to follow.


src/lib/Hydra/Plugin/RunCommand.pm
```diff
@@ -38,6 +38,64 @@ sub eventMatches {
     return 0;
 }
 
+sub makeRunCommandJson {
```
cole-h, line 72: Technically, this doesn't _actually_  create the json, but the hash that will be encoded into a json. Maybe there's a better name for it (or, move the `encode_json` call in)?

grahamc: Gosh. I don't know what to call it! I don't want to move the encode_json inside it, however it is also not a generic Perl hash: it has JSON-specific values (`true` and `false`) in it.

cole-h (approved):
LGTM. Tests ❤️

PR #1079, opened 2021-12-16
foreman: run the dev server with --restart and --debug

grahamc:
Debug enables info and debug log levels and provides quite a lot of useful dev-time information.

Restart automatically restarts the app when the .pm files change.


cole-h (approved):
That would be very nice indeed! No more "^C -> up arrow"!

PR #1083, opened 2021-12-20
Faster notifications and buildQueued support

grahamc:
In May 2017 plugin support for executing a buildQueued event was introduced in 582c39942097539aa595d6945a82c4e25ca16ff1. At the time, hydra-notify was executed every time an event was to be delivered.

A month later in June 2017, the buildQueued hook was commented out in 803833aba77e1082c14857aa26933fc7fe5ae190:

    Disable the build-queued hook
    
    This can take an excessive amount of time. For example, on
    hydra.nixos.org, a call to hydra-notify takes 0.7s even if there are
    no plugins. So for an eval with ~45K new builds, the calls to
    hydra-notify add up to about 9 hours.
    
    The proper fix would be to pass a list of build IDs, or an eval ID.

A couple of years later in August of 2019, buildQueued support was deleted altogether from hydra-notify when it was converted from an exec-per-event to a Postgres LISTEN/NOTIFY event: 29468995040ae21e0e1c14c1bdbb16ccb514caa8. 

Meanwhile, lots of plugins implement behavior for `buildQueued`.

I note that a proper fix would be passing a list of build IDs or an eval ID, however performance testing today showed that Posgres can process 100,000 NOTIFY's in about 2 seconds and 1,000,000 in about 20 seconds:

```perl
use strict;
use warnings;
use Setup;
use Time::HiRes qw( gettimeofday tv_interval );
use Hydra::PostgresListener;
use Test2::V0;

my $ctx = test_context();
my $db = $ctx->db(0);
my $dbh = $db->storage->dbh;

my $listener = Hydra::PostgresListener->new($dbh);

$listener->subscribe("build_queued");

my $messages_to_send = 1000000;

subtest "Sending a lot of messages" => sub {
    my $start_time = [gettimeofday()];
    # Send a lot of messages
    for (1 .. $messages_to_send) {
        $dbh->do("notify build_queued, ?", undef, 1);
    }
    my $runtime = tv_interval($start_time);
    is($runtime, 0, "The runtime is...");
};

done_testing;
```

Once I hooked this up to the current hydra-notify code it took 2 minutes and 45 seconds to do nothing with 1 million events. This is because every plugin gets executed on every event using the base plugin's "noop" behavior.

~To fix this I introduced the concept of "interest": a plugin can define what events it is interested in, and the Event handlers also define their interests, and the Event is executed with the Plugin if there is an intersection.~

To fix this I deleted the default noop implementations in the base plugin, and each Event handler checks to see if the plugin has defined the subs necessary to run.

With this change, hydra-notify can process 100,000 messages in about 7-10 seconds.

I don't particularly like this approach of "interest" but I do want to maintain some backwards compatibility with the plugins API, and this seemed like a cheap way to experiment.

To that end, this is an RFC PR at the moment. It is not suitable for merging because it contains no tests.

Todo:

- [x] Vet the idea is a good one
- [x] Write tests and documentation
- [x] Perhaps try it out in hydra.nixos.org for a minute

### DIFF
--- a/src/lib/Hydra/Event.pm
+++ b/src/lib/Hydra/Event.pm
@@ -38,6 +38,12 @@ sub new_event {
     }, $self;
 }
 
+sub interested {
+    my ($self, $plugin) = @_;
+
+    return $self->{"event"}->interestedIn($plugin);
+}
+
 sub execute {
     my ($self, $db, $plugin) = @_;
     return $self->{"event"}->execute($db, $plugin);
--- a/src/lib/Hydra/Event.pm
+++ b/src/lib/Hydra/Event.pm
@@ -3,10 +3,12 @@ package Hydra::Event;
 use strict;
 use warnings;
 use Hydra::Event::BuildFinished;
+use Hydra::Event::BuildQueued;
 use Hydra::Event::BuildStarted;
 use Hydra::Event::StepFinished;
 
 my %channels_to_events = (
+  build_queued => \&Hydra::Event::BuildQueued::parse,
   build_started => \&Hydra::Event::BuildStarted::parse,
   step_finished => \&Hydra::Event::StepFinished::parse,
   build_finished => \&Hydra::Event::BuildFinished::parse,
--- a/src/lib/Hydra/Event.pm
+++ b/src/lib/Hydra/Event.pm
@@ -38,7 +38,7 @@ sub new_event {
     }, $self;
 }
 
-sub interested {
+sub interestedIn {
     my ($self, $plugin) = @_;
 
     return $self->{"event"}->interestedIn($plugin);
--- a/doc/manual/src/notifications.md
+++ b/doc/manual/src/notifications.md
@@ -8,6 +8,12 @@ Notifications are passed from `hydra-queue-runner` to `hydra-notify` through Pos
 
 Note that the notification format is subject to change and should not be considered an API. Integrate with `hydra-notify` instead of listening directly.
 
+### `build_queued`
+
+* **Payload:** Exactly one value, the ID of the build.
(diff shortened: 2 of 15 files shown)


src/lib/Hydra/Event.pm
```diff
@@ -36,6 +38,12 @@ sub new_event {
     }, $self;
 }
 
+sub interested {
```
andir, line 41: why not call this `interestedIn`?

grahamc: Good idea, fixed.

grahamc:
I've had this deployed on Hydra for a bit. We recently had an evaluation register ~50k new builds, and hydra-notify was able to process all of them within less than 1 sample period (ie: 15s):

image

andir (approved):
Looks good and clean to me. Good work!

grahamc:
Another bit of data to look at: the CompressLogs is _intended_ to run at the end of every step being finished. You can see that happening here:

image

Compare to GithubPulls which used to run on every step finishing too, but no longer does so since it has nothing to do:

image

Note in both of these the Y axis is the total # of invocations.

grahamc:
I added a new metric to count how much time is spent processing events in total, see: https://monitoring.nixos.org/prometheus/graph?g0.expr=rate(notify_event_runtime_sum[10m])&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h this will let us know how this PR is affecting the notify daemon.

grahamc:
I got about 16h of data collected from that change and re-deployed this PR at 2021-12-22T17:32:43+00:00.

grahamc:
I think the data is pretty compelling and that we're not drastically harming the performance of hydra-notify by turning build-queued on again.

image

PR #1089, opened 2022-01-05
Fetch the tracker HTML from the config file, remove HYDRA_TRACKER

grahamc:
The indentation in the hydra.conf makes it possible to include multi-line
strings without it being likely that the contents of the tracker
is mis-parsed or interrupts tho config parser.

It isn't impossible / foolproof probably, but it shouldn't be likely.


cole-h (approved):
Big +1 for not using env vars where they have no business being used...

PR #1091, opened 2022-01-07
hydra-queue-runner: support store URIs declaring an alternate store location

Ma27:
When having a builder like this in `/etc/nix/machines`

    ssh://mfbuild?remote-store=/home/bosch/store

Hydra cannot build there since it tries to pass the entire value to
`ssh(1)` which doesn't work. Also, an alternate store-location is e.g.
used if the user isn't a trusted user on the remote system and thus
cannot use `/nix/store`.

If such a URI is given, Hydra will now add a `--store /home/bosch/store`
to the `ssh`-command to select the appropriate location remotely.

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -10,6 +10,7 @@
 #include "util.hh"
 #include "worker-protocol.hh"
 #include "finally.hh"
+#include "url.hh"
 
 using namespace nix;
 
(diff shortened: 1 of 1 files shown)


grahamc:
This looks like a pretty good idea to me. I'm trying to not add / change / fix code without tests, but it isn't really reasonable to ask for that for the C++ codebase since there _are_ no tests. However, I managed to trick @andir in to setting up the infrastructure to make tests, which should be available soonish. To make this easier to test, could you break this out in to a smaller function?

Ma27:
@grahamc updated the PR accordingly :)

grahamc:
Looks pretty good! I hope we can add some C++ tests in soon. Thanks!

PR #1092, opened 2022-01-07
Allow restarting failed builds even with no eval to compare to

cole-h:
This is the smallest change that will make this work. All my other ideas would have resulted in a larger refactoring.

---

Closes https://github.com/NixOS/hydra/issues/1088.

---

TODO:

- [x] write a test

### DIFF
--- a/src/lib/Hydra/Controller/JobsetEval.pm
+++ b/src/lib/Hydra/Controller/JobsetEval.pm
@@ -64,17 +64,8 @@ sub view_GET {
 
     $c->stash->{otherEval} = $eval2 if defined $eval2;
 
-    sub cmpBuilds {
-        my ($left, $right) = @_;
-        return $left->get_column('job') cmp $right->get_column('job')
-            || $left->get_column('system') cmp $right->get_column('system')
-    }
-
     my @builds = $eval->builds->search($filter, { columns => [@buildListColumns] });
     my @builds2 = defined $eval2 ? $eval2->builds->search($filter, { columns => [@buildListColumns] }) : ();
-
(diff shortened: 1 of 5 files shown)


src/lib/Hydra/Controller/JobsetEval.pm
```diff
@@ -119,6 +120,9 @@ sub view_GET {
         } else {
             push @{$c->stash->{new}}, $build if !$found;
         }
+        if ($build->buildstatus != 0) {
+            push @{$c->stash->{failed}}, $build;
+        }
```
grahamc, line 125: Since in the end we just count how many we have, we could make `failed = 0` above and increment it here.

We may also want to extend this check: if buildstatus is defined AND non-zero. It is extremely edge-case-ey but consider a scenario where builds are queued and none are failed: I think `failed` would be >0 in that scenario.

grahamc:
This looks like it'll be hard to test. I think an easier way to test this would be to do the hard thing: extract this in to a function which buckets out the builds. Then we do the easy thing: write a test for the bucketing function.

src/lib/Hydra/Helper/BuildDiff.pm
```diff
@@ -0,0 +1,82 @@
...
+        }
+        if (defined $build->buildstatus && $build->buildstatus != 0) {
+            push @{$ret->{failed}}, $build;
+        }
+    }
+
+    return $ret;
+}
+
+1;
```
grahamc, line 82: d'you mind installing the editorconfig plugin? It'll add these newlines automatically (also, if you wouldn't mind fixing this one :).

PR #1099, opened 2022-01-10
Provide a dark theme when the user prefers it

dasJ:
Somewhat inspired by the dark GitLab theme.
Screenshots (don't worry about the wideness, the screenshots also include #965):

image
image
image
image

### DIFF
--- a/src/root/static/css/hydra.css
+++ b/src/root/static/css/hydra.css
@@ -138,10 +138,227 @@ td.step-status span.warn {
 }
 
 .date {
-	cursor: help;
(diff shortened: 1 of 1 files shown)


grahamc:
oh my!

dasJ:
If people want to test this in their local Hydra instances remember you might need to purge the cache (force-reload) and you need your browser to be configured to use dark mode on websites

rbvermaa:
@dasJ Thanks a lot for this PR. Our colleagues really like it!

PR #1101, opened 2022-01-10
hydra-eval-jobset: test emitted notifications

grahamc:
This is an addition of a simple test which compares the NOTIFY/LISTEN events received vs. expected. This is an attempt at putting some tests around the behavior of hydra-eval-jobset without changing its structure much.

### DIFF
--- a/t/lib/CliRunners.pm
+++ b/t/lib/CliRunners.pm
@@ -21,10 +21,12 @@ sub evalSucceeds {
     my ($jobset) = @_;
     my ($res, $stdout, $stderr) = captureStdoutStderr(60, ("hydra-eval-jobset", $jobset->project->name, $jobset->name));
     $jobset->discard_changes;  # refresh from DB
-    chomp $stdout; chomp $stderr;
-    print STDERR "Evaluation errors for jobset ".$jobset->project->name.":".$jobset->name.": \n".$jobset->errormsg."\n" if $jobset->errormsg;
-    print STDERR "STDOUT: $stdout\n" if $stdout ne "";
-    print STDERR "STDERR: $stderr\n" if $stderr ne "";
+    if ($res) {
+        chomp $stdout; chomp $stderr;
+        print STDERR "Evaluation errors for jobset ".$jobset->project->name.":".$jobset->name.": \n".$jobset->errormsg."\n" if $jobset->errormsg;
+        print STDERR "STDOUT: $stdout\n" if $stdout ne "";
+        print STDERR "STDERR: $stderr\n" if $stderr ne "";
+    }
     return !$res;
 }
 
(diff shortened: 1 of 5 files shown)


t/lib/CliRunners.pm
```diff
@@ -21,13 +25,28 @@ sub evalSucceeds {
...
+    my ($jobset) = @_;
+    my ($res, $stdout, $stderr) = captureStdoutStderr(60, ("hydra-eval-jobset", $jobset->project->name, $jobset->name));
+    $jobset->discard_changes;  # refresh from DB
+    if (!$res) {
+        chomp $stdout; chomp $stderr;
+        print STDERR "Evaluation errors for jobset ".$jobset->project->name.":".$jobset->name.": \n".$jobset->errormsg."\n" if $jobset->errormsg;
+        print STDERR "STDOUT: $stdout\n" if $stdout ne "";
+        print STDERR "STDERR: $stderr\n" if $stderr ne "";
+    }
+    return !!$res;
```
cole-h, line 47: :eyes: :question: 

At first glance, I'm thinking "maybe this is used to coerce it to an `int`...?", so maybe an explicit `int()` would be better than a double negative?

t/lib/CliRunners.pm
```diff
@@ -21,13 +25,28 @@ sub evalSucceeds {
...
+        print STDERR "STDERR: $stderr\n" if $stderr ne "";
+    }
     return !$res;
 }
 
+sub evalFails {
+    my ($jobset) = @_;
+    my ($res, $stdout, $stderr) = captureStdoutStderr(60, ("hydra-eval-jobset", $jobset->project->name, $jobset->name));
+    $jobset->discard_changes;  # refresh from DB
+    if (!$res) {
```
cole-h, line 41: Do we need this block? If so, I think the negation is wrong. If the eval fails, the exit code will (probably?) be non-zero, and the `if (!255) {` (`if (0) {`) block won't be entered. Thus, it'll only reach this when the eval succeeds, in which case `Evaluation errors for jobset` is incorrect, too.

grahamc: The goal is to print evaluation output if the evaluation succeeds. I think the message should be changed to say "Evaluation unexpectedly succeeded...." instead of "Evaluation errors". But otherwise I _think_ the code is right?

cole-h: Yeah, it was just the clash between those two that threw me off. "Evaluation unexpectedly succeeded" would be much clearer.

t/lib/CliRunners.pm
grahamc: Yep! Forcing it to be a simple bool.

cole-h (approved):
Nice, much clearer, IMHO!

PR #1105, opened 2022-01-11
hydra-eval-jobset: send cached_build_finished, cached_build_queued

grahamc:
These new notifications allow plugins to operate on existing builds being reused by a new evaluation. A key example of this is for sending "queued" and "finished" notifications to systems like GitHub. See: https://github.com/DeterminateSystems/hydra/pull/6

Note this does not reuse build_finished or build_queued because those might perform tasks that are very noisy assuming a low execution rate per build. Furthermore, those events operate on _all_ evaluations the build is part of, and these events operate only on the specific evaluation which is new.

This will send many thousand notifications per evaluation for Nixpkgs / NixOS on hydra.nixos.org, so before merging this I want to deploy it and see how it impacts the database and hydra-notify's performance.

### DIFF
--- a/src/script/hydra-eval-jobset
+++ b/src/script/hydra-eval-jobset
@@ -399,7 +399,7 @@ sub getPrevJobsetEval {
 
 # Check whether to add the build described by $buildInfo.
 sub checkBuild {
-    my ($db, $jobset, $inputInfo, $buildInfo, $buildMap, $prevEval, $jobOutPathMap, $plugins) = @_;
+    my ($db, $jobset, $eval, $inputInfo, $buildInfo, $buildMap, $prevEval, $jobOutPathMap, $plugins) = @_;
 
     my @outputNames = sort keys %{$buildInfo->{outputs}};
     die unless scalar @outputNames;
@@ -435,10 +435,17 @@ sub checkBuild {
                 # the Nixpkgs jobset with PostgreSQL.
                 { jobset_id => $jobset->get_column('id'), job => $jobName,
                   name => $firstOutputName, path => $firstOutputPath },
-                { rows => 1, columns => ['id'], join => ['buildoutputs'] });
+                { rows => 1, columns => ['id', 'finished'], join => ['buildoutputs'] });
             if (defined $prevBuild) {
                 #print STDERR "    already scheduled/built as build ", $prevBuild->id, "\n";
                 $buildMap->{$prevBuild->id} = { id => $prevBuild->id, jobName => $jobName, new => 0, drvPath => $drvPath };
+
+                if ($prevBuild->finished) {
+                    $db->storage->dbh->do("notify cached_build_finished, ?", undef, "${\$eval->id}\t${\$prevBuild->id}");
+                } else {
+                    $db->storage->dbh->do("notify cached_build_queued, ?", undef, "${\$eval->id}\t${\$prevBuild->id}");
+                }
+
                 return;
             }
         }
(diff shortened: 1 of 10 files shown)


src/script/hydra-eval-jobset
```diff
@@ -723,31 +730,35 @@ sub checkJobsetWrapped {
...
-            , hasnewbuilds => $jobsetChanged ? 1 : 0
-            , nrbuilds => $jobsetChanged ? scalar(keys %buildMap) : undef
-            , flake => $flakeRef
-            , nixexprinput => $jobset->nixexprinput
-            , nixexprpath => $jobset->nixexprpath
-            });
+        $ev->update({
+            hasnewbuilds => $jobsetChanged ? 1 : 0,
+            nrbuilds => $jobsetChanged ? scalar(keys %buildMap) : undef
+        });
```
grahamc, line 761: I think this changing to an update instead of a single, correct creation is okay. This all takes place within a transaction, and I don't think there are triggers that this is going to confuse.

t/Event/CachedBuildFinished.t
```diff
@@ -0,0 +1,114 @@
...
+
+        is($event->interestedIn($plugin), 1, "The plugin is interesting.");
+    };
+};
+
+subtest "load" => sub {
+    my ($build) = $db->resultset('Builds')->search(
+      { },
+      { limit => 1 }
+    )->single;
```
cole-h, line 84: Really minor nit, but is there any reason this needs to be split across lines like so?

t/Event/CachedBuildQueued.t
```diff
@@ -0,0 +1,114 @@
...
+
+        is($event->interestedIn($plugin), 1, "The plugin is interesting.");
+    };
+};
+
+subtest "load" => sub {
+    my ($build) = $db->resultset('Builds')->search(
+      { },
+      { limit => 1 }
+    )->single;
```
cole-h, line 84: Same here.

t/scripts/hydra-eval-jobset/notifications.t
```diff
@@ -69,7 +69,23 @@ subtest "on a fresh evaluation with changed sources" => sub {
...
+    # so we can't expect the list in any specific order here.
+    is(
+        [sort(
+            $listener->block_for_messages(0)->()->{"channel"},
+            $listener->block_for_messages(0)->()->{"channel"},
+            $listener->block_for_messages(0)->()->{"channel"},
+            $listener->block_for_messages(0)->()->{"channel"}
+        )],
+        [
+            "build_queued",
```
cole-h, line 82: Would it be useful to document which job each of these events corresponds to? I imagine the two cached finished builds are both the stable passing and failing jobs. Is the cached queued build for the variable job? Or is that the queued build?

grahamc: Hmmm... yeah, I think that would be a good idea! It is a bit tricky because the events for builds are sent (on purpose by the hydra-eval-jobset code) in a randomized order, but we could piece it together here.

grahamc: I added comments about what the events correspond to. Writing the test to actually validate the content and deal with random order is, uh, frustrating.

cole-h: Much better and clearer, thanks!

grahamc:
I deployed this PR at Jan 11 20:01 CET, let's check back in a few days.

grahamc:
It looks like a big new evaluation took place, and hydra-notify processed 111,000 cached_build_finished events in 14.1s. Postgres looks unaffected.

grahamc:
After a few days I'm noticing the amount of time processing events is not significantly impacted, despite clearly handling a much larger volume of messages more regularly:

image

t/Event/CachedBuildFinished.t
grahamc: :+1: Nope, probably was trying different queries and got here. Fixed!

PR #1093, opened 2022-01-09
Builds: drop the jobset and project columns

grahamc:
This is a work in progress PR which is not carefully validated. The tests do pass which is nice to see. However, I've barely even started the webserver or clicked around.

The project and jobset columns were the only foreign key to a jobset and project. #969 added a jobset_id as a replacement, but didn't go all the way to dropping the now duplicate columns. This PR finishes that.

I think a good way to apply this PR is to pull out individual commits, updating individual systems and merging / deploying them one at a time, validating in production that the change didn't break anything. This way, we can easily revert changes if it is the easiest solution.

Several commits in this PR implemented the easiest possible solution to make a test pass. Those patches should be undone, and reimplemented. One good example of this is the json_hints change which critically breaks the API. Another example is the dropping of several indexes without consideration for how it impacts the performance properties of Hydra.

--

before this PR:

```
hydra=> select pg_table_size('builds');
 pg_table_size 
---------------
   72219074560
(1 row)
```

after this PR followed by a `vacuum full builds` (which took 3734085.204 ms):

```
hydra=> select pg_table_size('builds');
 pg_table_size 
---------------
   56651767808
(1 row)
```

a reduction of 14GB.

PR #1108, opened 2022-01-14
Builds table: replace some project and jobset references in the Projects and Builds controllers

grahamc:
Extracted from #1093.

My methodology here is to:

1. Pick a commit from #1093

The picked commit in this case was `Builds: LatestSucceeded*`.

### DIFF
--- a/t/Hydra/Controller/Job/builds.t
+++ b/t/Hydra/Controller/Job/builds.t
@@ -1,0 +1,31 @@
+use strict;
+use warnings;
+use Setup;
+use Test2::V0;
+use Catalyst::Test ();
+use HTTP::Request::Common;
+
+my $ctx = test_context();
+
+Catalyst::Test->import('Hydra');
+
+my $builds = $ctx->makeAndEvaluateJobset(
+    expression => "basic.nix",
+    build => 1
+);
+
+my $build = $builds->{"empty_dir"};
+my $jobset = $build->jobset;
+my $project = $build->project;
+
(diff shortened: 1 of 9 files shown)


t/Hydra/Controller/Job/builds.t
```diff
@@ -0,0 +1,31 @@
...
+my $builds = $ctx->makeAndEvaluateJobset(
+    expression => "basic.nix",
+    build => 1
+);
+
+my $build = $builds->{"empty_dir"};
+my $jobset = $build->jobset;
+my $project = $build->project;
+
+subtest "/job/PROJECt/JOBSET/JOB/all" => sub {
```
cole-h, line 21: Minor nit: any reason why the final `t` in `PROJECt` is lower-case?

t/Hydra/Controller/Job/builds.t
```diff
@@ -0,0 +1,31 @@
...
+    my $response = request(GET '/job/' . $project->name . '/' . $jobset->name . '/' . $build->job . '/all');
+    ok($response->is_success, "The page showing the job's builds returns 200.");
+};
+
+subtest "/job/PROJECt/JOBSET/JOB/channel/latest" => sub {
```
cole-h, line 26: Here, too.

cole-h (commented):
2 very minor nits, otherwise the diffs and tests all look sane to me.

t/Hydra/Controller/Job/builds.t
grahamc: Typo + copypasta I'm sure!

PR #1110, opened 2022-01-14
Project jobset: api: update endpoints to stop using build.{project,jobset}

grahamc:
Extracted from #1093.

My methodology here is to:

1. Pick a commit from #1093

The picked commit in this case was `Fixup project name, jobset name, in buildToHash`. After implementing a test for that behavior I noticed a handful of other codepaths that used project and jobset, so I fixed those too. I thought the `push*` APIs used them, but they don't.


cole-h (approved):
Tests look sane to me.

PR #1111, opened 2022-01-14
Project jobset columns: fixup /queue-summary and /machines

grahamc:
Extracted from #1093.

My methodology here is to:

1. Pick a commit from #1093

The picked commit in this case was `queue summary: fix refs` and `machines: fixup refs`.


cole-h (approved):

vcunat:
Hydra should now be running including this PR (says 9cc3e42), but in https://hydra.nixos.org/machines the jobset column is (still?) broken and the corresponding job links as well.  I believe that's quite a recent regression.

PR #1112, opened 2022-01-14
Project jobset: update search

grahamc:
Extracted from #1093.

My methodology here is to:

1. Pick a commit from #1093

The picked commit in this case was `search: fix references to jobset / project info`.

### DIFF
--- a/t/Hydra/Controller/Root/status.t
+++ b/t/Hydra/Controller/Root/status.t
@@ -2,26 +2,18 @@ use strict;
 use warnings;
 use Setup;
 use Data::Dumper;
-use JSON::MaybeXS qw(decode_json);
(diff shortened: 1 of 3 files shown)


t/Hydra/Controller/Root/status.t
```diff
@@ -50,6 +50,12 @@ subtest "/queue" => sub {
         print STDERR Dumper $response->content;
     }
 };
+
+subtest "/search" => sub {
+    my $response = request(GET '/search?query=one');
+    ok($response->is_success, "The search page 200's.");
+};
```
grahamc, line 57: This is insufficient. Needs test coverage on matching:

- project names
- jobset names
- job names
- builds' output path
- derivation path

PR #1117, opened 2022-01-15
queue-runner: track jobsets by ID

grahamc:
Extracted from #1093.

My methodology here is to:

1. Pick a commit from #1093

In this case, a lot of tests failed before, and they all pass now. However, as little as I know Perl, I know C++ even less. I'd love to have actual C++-level tests if possible. I'd also like it if someone who knows C++ reviewed this.

### DIFF
--- a/src/hydra-queue-runner/queue-monitor.cc
+++ b/src/hydra-queue-runner/queue-monitor.cc
@@ -96,8 +96,11 @@ bool State::getQueuedBuilds(Connection & conn,
         pqxx::work txn(conn);
 
         auto res = txn.exec_params
-            ("select id, project, jobset, job, drvPath, maxsilent, timeout, timestamp, globalPriority, priority from Builds "
-             "where id > $1 and finished = 0 order by globalPriority desc, id",
+            ("select builds.id, builds.jobset_id, jobsets.project as project, "
+             "jobsets.name as jobset, job, drvPath, maxsilent, timeout, timestamp, "
+             "globalPriority, priority from Builds "
+             "inner join jobsets on builds.jobset_id = jobsets.id "
+             "where builds.id > $1 and finished = 0 order by globalPriority desc, builds.id",
             lastBuildId);
 
         for (auto const & row : res) {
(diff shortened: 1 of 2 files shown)


src/hydra-queue-runner/state.hh
```diff
@@ -489,7 +490,7 @@ private:
         bool & stepFinished);
 
     Jobset::ptr createJobset(pqxx::work & txn,
-        const std::string & projectName, const std::string & jobsetName);
+        const std::string & projectName, const std::string & jobsetName, const int & jobsetID);
```
pennae, line 493: `const int &` isn't very common, usually that'd be an `int` or `const int`

src/hydra-queue-runner/state.hh
```diff
@@ -123,6 +123,7 @@ struct Build
     BuildID id;
     nix::StorePath drvPath;
     std::map<std::string, nix::StorePath> outputs;
+    int jobsetId;
```
pennae, line 126: `int` is probably an odd type here. `BuildID` is an `unsigned int`, might be better to use that here as well, maybe even duplicate the typedef for `JobsetID` should we ever have to switch that to long.

src/hydra-queue-runner/state.hh
grahamc: Thanks! Do you have a recommendation re const int vs int?

pennae: we'd use `const int` because immutability by default is nice.

src/hydra-queue-runner/state.hh
grahamc: Worth noting that the builds table has hundreds of millions of rows and the jobsets table has ~20, though that is definitely not universally true.

src/hydra-queue-runner/queue-monitor.cc
```diff
@@ -569,7 +573,7 @@ Step::ptr State::createStep(ref<Store> destStore,
 
 
 Jobset::ptr State::createJobset(pqxx::work & txn,
-    const std::string & projectName, const std::string & jobsetName)
+    const std::string & projectName, const std::string & jobsetName, JobsetID jobsetID)
```
pennae, line 576:
```suggestion
    const std::string & projectName, const std::string & jobsetName, const JobsetID jobsetID)
```
to match the declaration, everything else looks fine C++-wise :)

grahamc: Thanks, I fixed this and squashed them all in to one commit. I appreciate your review!

PR #1127, opened 2022-01-21
Create a bug report issue template

grahamc:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Hydra Server:**
+- OS and version: [e.g. NixOS 22.05.20211203.ee3794c]
+- Version of Hydra
+- Version of Nix Hydra is built against
+- Version of the Nix daemon
```
samueldr, line 30: Any commands to get these infos?

grahamc: I can't think of an easy way to do any of these today!

ajs124: `$(systemctl show nix-daemon.service|grep ExecStart=|grep -Po "(?<=path=).*?(?= )") --version`

The other two are harder. Maybe hydra should just expose them through the API somehow. AFAIR even what's displayed in the web UI is a lie and just taken from the storepaths or someplace like that. In fact, my hydra says "Hydra 2021-08-11 (using nix-2.5pre20211206_d1aaa7e).", which is _definitely_ wrong.

grahamc: @samueldr What if I put a thing before this saying "It is fine to leave these fields empty if you don't know how to answer them." ?

mohe2015: Just (optional) in the same line at the end?

samueldr: Sure, don't block on that. It was if possible :).

grahamc: :+1:  thanks!

PR #1130, opened 2022-01-21
hydra-create-user: support prompting for password

grahamc:
Original body:

> I'm not sure this is a good implementation as-is. It does work,
> but the password gets echo'd to the screen. I tried to use IO::Prompt
> but IO::Prompt really seems to want to read the password from ARGV.

but now it is noecho, so I think it should be good to go.

### DIFF
--- a/src/script/hydra-create-user
+++ b/src/script/hydra-create-user
@@ -5,6 +5,7 @@ use warnings;
 use Hydra::Schema;
 use Hydra::Helper::Nix;
 use Hydra::Model::DB;
+use Term::ReadKey;
 use Getopt::Long qw(:config gnu_getopt);
 
 sub showHelp {
@@ -143,13 +144,17 @@ $db->txn_do(sub {
         }
 
         if (defined $passwordPrompt) {
+            ReadMode 2;
             print STDERR "Password: ";
             my $password = <STDIN> // "";
             chomp $password;
 
-            print STDERR "Password Confirmation: ";
+            print STDERR "\nPassword Confirmation: ";
             my $passwordConfirm = <STDIN> // "";
             chomp $passwordConfirm;
+            ReadMode 0;
+
+            print STDERR "\n";
 
             if ($password ne $passwordConfirm) {
                 die "Passwords don't match."
--- a/src/script/hydra-create-user
+++ b/src/script/hydra-create-user
@@ -14,6 +14,7 @@ Usage: hydra-create-user NAME
   [--type hydra|google|github]
   [--full-name FULLNAME]
   [--email-address EMAIL-ADDRESS]
+  [--password-prompt]
   [--password-hash HASH]
   [--password PASSWORD (dangerous)]
   [--wipe-roles]
@@ -26,6 +27,14 @@ specified.  If --rename-from is given, the specified account is
 renamed.
 
 * Specifying Passwords
+
+** Interactively
+
+Pass `--password-prompt` to collect the password on stdin.
+
+Example:
+
+  $ hydra-create-user alice --password-prompt --role admin
 
 ** Specifying a Hash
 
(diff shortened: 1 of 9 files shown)


src/script/hydra-create-user
```diff
@@ -71,6 +92,11 @@ GetOptions("rename-from=s" => \$renameFrom,
 die "$0: one user name required\n" if scalar @ARGV != 1;
 my $userName = $ARGV[0];
 
+my $chosenPasswordOptions = grep { defined($_) }  ($passwordPrompt, $passwordHash, $password);
+if ($chosenPasswordOptions > 1) {
+    die "$0: please specify one of --password-prompt or --password-hash. See --help for more information.\n";
```
cole-h, line 97: Maybe this would make it more obvious that you can only use one or the other (or the unmentioned other).

```suggestion
    die "$0: please specify only one of --password-prompt or --password-hash. See --help for more information.\n";
```

src/script/hydra-create-user
```diff
@@ -27,6 +28,14 @@ renamed.
 
 * Specifying Passwords
 
+** Interactively
+
+Pass `--password-prompt` to collect the password on stdin.
```
cole-h, line 34: IMO, this should also note that it will be hashed with argon2 (and also that it's not stored in plaintext, maybe?).

src/script/hydra-create-user
```diff
@@ -129,6 +141,24 @@ $db->txn_do(sub {
         if (defined $passwordHash) {
```
cole-h, line 142: Should the `!(defined $passwordHash)` just above this be changed to also take into account `$passwordPrompt`?

Or maybe that conditional shoud be removed altogether? We already `die` if more than one of `$passwordHash`, `$password`, and `$passwordPrompt` are defined around line 95...

src/script/hydra-create-user
```diff
@@ -25,27 +25,37 @@ exists, roles are added to the existing roles unless --wipe-roles is
...
-Examples:
+SHA1 is also accepted, but SHA1 support is deprecated and the user's
+password will be upgraded to Argon2id on first login.
 
-Create a user with an argon2 password:
+** Specifying a plain-text password as an argument (dangerous)
 
-  $ hydra-create-user alice --password-hash '$argon2id$v=19$m=262144,t=3,p=1$NFU1QXJRNnc4V1BhQ0NJQg$6GHqjqv5cNDDwZqrqUD0zQ' --role admin
+This option is dangerous and should not be used: it exposes passwords to
+other users on the system. This option only exists for backwards
```
cole-h, line 65: I wonder, should we worry about backwards compatibility with a dangerous option like this? Would it not be better and safer to just remove this altogether? Or maybe (somehow) only make it usable in a dev server...? (If that's even possible to detect)

src/script/hydra-create-user
grahamc: Good catch, we can delete that extra condition.

src/script/hydra-create-user
grahamc: I don't want to drop it right away, but I did add an admonition:

```
[nix-shell:~/projects/github.com/NixOS/hydra]$ hydra-create-user foobar --password oeusnth
creating new user `foobar'
Submitting plaintext passwords as arguments is deprecated and will be removed. See --help for alternatives.
```

cole-h: Should we also change the invocation in `foreman/start-hydra.sh` to not use the raw `--password` (so we don't see that when running hydra in a dev shell)?

grahamc:
https://github.com/NixOS/hydra/pull/1130/commits/da1af1ce68d56b638aa8a0fdabe00b77ff5fedd5 updates existing docs to recommend --password-prompt.

PR #1131, opened 2022-01-21
Fix invalid YAML in documentation

kenranunderscore:
When trying to get LDAP integration to work I naively copied the example YAML config file, but later saw that there was a typo, making it invalid. This adds the missing colon.


grahamc:
Thank you!

PR #1132, opened 2022-01-21
LDAP support: require the prefix 'hydra_' to match documentation

grahamc:


cole-h (approved):

grahamc:
I pushed a commit verifying the prefix must match exactly.

PR #1134, opened 2022-01-21
Redo LDAP config in the main configuration and add role mappings

dasJ:
The ldap test passes, I didn't do any more testing.
cc @ajs124

### DIFF
--- a/doc/manual/src/configuration.md
+++ b/doc/manual/src/configuration.md
@@ -101,47 +101,70 @@ Using LDAP as authentication backend (optional)
 Instead of using Hydra\'s built-in user management you can optionally
 use LDAP to manage roles and users.
 
-The `hydra-server` accepts the environment variable
-*HYDRA\_LDAP\_CONFIG*. The value of the variable should point to a valid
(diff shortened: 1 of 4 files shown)


ajs124:
That's what I intended for somebody to do, back when I opened the draft PR, way back when.

The only issue is that this now is a backwards-incompatible change, so the 3 people that use LDAP with their hydra will need to change their config.

It would also be nice if the example included how to include the password from an external file.

Ma27:
Very nice! Happy to test it rather soonish, also then there'd only be two people left who'd have to upgrade ;)

grahamc:
This looks good to me. Since this is a fairly, uh, _significant_ change for people who depend on it, I'm inclined to go the extra mile in supporting backwards compatibility with loading YAML from an env var, and providing a default role mapping.

I'm happy to take this work on myself, though.

This looks good to me, though, and I'm glad to see the documentation, thank you!

grahamc:
See #1151 which should be 100% BC and also resolves an issue with this PR in the case that a single group grants multiple roles.

PR #1106, opened 2022-01-12
 Store and display the output of run commands

dasJ:
Replaces https://github.com/DeterminateSystems/hydra/pull/7

image

PR #1137, opened 2022-01-26
Store and display the output of RunCommands

cole-h:
Contains https://github.com/NixOS/hydra/pull/1106.

TODO:

- [x] Make `raw`, `tail`, and `pretty` buttons for the `Logs` link to reflect its capabilities

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -275,7 +275,11 @@ in
 
           mkdir -m 0700 -p ${baseDir}/queue-runner
           mkdir -m 0750 -p ${baseDir}/build-logs
-          chown hydra-queue-runner.hydra ${baseDir}/queue-runner ${baseDir}/build-logs
+          mkdir -m 0750 -p ${baseDir}/runcommand-logs
+          chown hydra-queue-runner.hydra \
+            ${baseDir}/queue-runner \
+            ${baseDir}/build-logs \
+            ${baseDir}/runcommand-logs
 
           ${optionalString haveLocalDB ''
             if ! [ -e ${baseDir}/.db-created ]; then
@@ -457,6 +461,7 @@ in
         script =
           ''
             find /var/lib/hydra/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            find /var/lib/hydra/runcommand-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
           '';
         startAt = "Sun 01:45";
       };
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -436,12 +436,12 @@ in
                 if [ $(systemctl is-active $service) == active ]; then
                   echo "stopping $service due to lack of free space..."
                   systemctl stop $service
-                  date > /var/lib/hydra/.$service-stopped-minspace
+                  date > ${baseDir}/.$service-stopped-minspace
                 fi
               else
                 if [ $spaceleft -gt $(( ($minFreeGB + 10) * 1024**3)) -a \
-                     -r /var/lib/hydra/.$service-stopped-minspace ] ; then
-                  rm /var/lib/hydra/.$service-stopped-minspace
+                     -r ${baseDir}/.$service-stopped-minspace ] ; then
+                  rm ${baseDir}/.$service-stopped-minspace
                   echo "restarting $service due to newly available free space..."
                   systemctl start $service
                 fi
@@ -460,8 +460,8 @@ in
       { path = [ pkgs.bzip2 ];
         script =
           ''
-            find /var/lib/hydra/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
-            find /var/lib/hydra/runcommand-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            find ${baseDir}/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            find ${baseDir}/runcommand-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
           '';
         startAt = "Sun 01:45";
       };
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -461,7 +461,7 @@ in
         script =
           ''
             find ${baseDir}/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
-            find ${baseDir}/runcommand-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            find ${baseDir}/runcommand-logs -type f -mtime +3 -size +0c | xargs -r bzip2 -v -f
           '';
         startAt = "Sun 01:45";
       };
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -461,7 +461,6 @@ in
         script =
           ''
             find ${baseDir}/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
-            find ${baseDir}/runcommand-logs -type f -mtime +3 -size +0c | xargs -r bzip2 -v -f
           '';
         startAt = "Sun 01:45";
       };
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -538,7 +538,7 @@ sub runcommandlog :Local :Args(1) {
     die if defined $tail && $tail !~ /^[0-9]+$/;
 
     my $runlog = $c->model('DB')->resultset('RunCommandLogs')->find({ uuid => $uuid });
-    my $logFile = constructRunCommandLogPath($runlog);
+    my $logFile = constructRunCommandLogPath($runlog) or notFound($c, "RunCommandLog not found.");
     if (-f $logFile) {
         serveLogFile($c, $logFile, $tail);
         return;
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -537,8 +537,10 @@ sub runcommandlog :Local :Args(1) {
 
     die if defined $tail && $tail !~ /^[0-9]+$/;
 
-    my $runlog = $c->model('DB')->resultset('RunCommandLogs')->find({ uuid => $uuid });
-    my $logFile = constructRunCommandLogPath($runlog) or notFound($c, "RunCommandLog not found.");
+    my $runlog = $c->model('DB')->resultset('RunCommandLogs')->find({ uuid => $uuid })
+        or notFound($c, "The RunCommand log is not available.");
+
+    my $logFile = constructRunCommandLogPath($runlog);
     if (-f $logFile) {
         serveLogFile($c, $logFile, $tail);
         return;
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -537,7 +537,8 @@ sub runcommandlog :Local :Args(1) {
 
     die if defined $tail && $tail !~ /^[0-9]+$/;
 
-    my $logFile = constructRunCommandLogPath($uuid);
+    my $runlog = $c->model('DB')->resultset('RunCommandLogs')->find({ uuid => $uuid });
+    my $logFile = constructRunCommandLogPath($runlog);
     if (-f $logFile) {
         serveLogFile($c, $logFile, $tail);
         return;
--- a/src/lib/Hydra/Controller/Root.pm
+++ b/src/lib/Hydra/Controller/Root.pm
@@ -531,13 +531,13 @@ sub log :Local :Args(1) {
 }
 
 sub runcommandlog :Local :Args(1) {
-    my ($self, $c, $filename) = @_;
+    my ($self, $c, $uuid) = @_;
 
     my $tail = $c->request->params->{"tail"};
 
     die if defined $tail && $tail !~ /^[0-9]+$/;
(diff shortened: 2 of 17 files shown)


src/lib/Hydra/Controller/Root.pm
```diff
@@ -529,5 +529,20 @@ sub log :Local :Args(1) {
...
     }
 }
+sub runcommandlog :Local :Args(1) {
+    my ($self, $c, $uuid) = @_;
+
+    my $tail = $c->request->params->{"tail"};
+
+    die if defined $tail && $tail !~ /^[0-9]+$/;
+
+    my $logFile = constructRunCommandLogPath($uuid);
```
grahamc, line 539: I think we should probably pass the `RunCommandLog` Result class to `constructRunCommandLogPath`, and let constructRunCommandLogPath pick the UUID out. This will make sure that the record must exist for us to try and calculate a file for it. This helps avoid any spookiness about filesystem traversal and user input (which is still a problem with this bit of code.)

src/lib/Hydra/Helper/Nix.pm
```diff
@@ -589,4 +591,13 @@ sub isLocalStore {
 }
 
 
+sub constructRunCommandLogPath {
+    my ($filename) = @_;
```
grahamc, line 595: I'm missing a test for this sub in Hydra/Helper/Nix.t. I think we should also test what happens if the UUID is empty, and possibly a brief check of what we expect it to look like, similar to https://github.com/NixOS/hydra/pull/1137/files#diff-56076a02b6158367f4b32f3aa1acb339ed87a357e5223503eb2e3f3a2e427daeR26. Also note my comment in Root.pm about taking the record here.

hydra-module.nix
```diff
@@ -171,7 +171,7 @@ in
       buildMachinesFiles = mkOption {
         type = types.listOf types.path;
         default = optional (config.nix.buildMachines != []) "/etc/nix/machines";
-        example = [ "/etc/nix/machines" "/var/lib/hydra/provisioner/machines" ];
+        example = [ "/etc/nix/machines" "${baseDir}/provisioner/machines" ];
         description = "List of files containing build machines.";
```
grahamc, line 175: We probably shouldn't use a variable in the example value. However, this is overall a nice cleanup :+1:.

grahamc (commented):
Looking pretty good. I added a handful of comments. Since we're mucking around in the view_log and nixlog function in Build.pm, we might want to add some smoke tests (fetch the page, make sure it succeeds) on those to be sure we didn't break it too obviously.

src/lib/Hydra/Helper/Nix.pm
```diff
@@ -19,6 +19,8 @@ our @EXPORT = qw(
     cancelBuilds
     captureStdoutStderr
     captureStdoutStderrWithStdin
+    constructRunCommandLogFilename
```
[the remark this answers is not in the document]
cole-h: Good catch -- was removed when we switched to uuid's, but I forgot to remove the export ^^;

src/root/runcommand-log.tt
```diff
@@ -0,0 +1,52 @@
...
+  Below
+  [% IF tail %]
+  are the last lines of
+  [% ELSE %]
+  is
+  [% END %]
+  the output of a RunCommand execution of the command <tt>[% runcommandlog.command %]</tt>
+  on <a href="[% c.uri_for('/build', build.id) %]">Build [% build.id %]</a>.
+  [% IF step && step.machine %]
+    It was built on <tt>[% step.machine %]</tt>.
```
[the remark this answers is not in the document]
cole-h: Oh yeah, it'll always be run on whatever the `hydra-notify` service is running, right?

src/root/runcommand-log.tt
```diff
@@ -0,0 +1,52 @@
...
+[% PROCESS common.tt %]
+
+<p>
+  Below
+  [% IF tail %]
+  are the last lines of
+  [% ELSE %]
+  is
+  [% END %]
+  the output of a RunCommand execution of the command <tt>[% runcommandlog.command %]</tt>
```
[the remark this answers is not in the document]
cole-h: Very good catch:

image

fixed:

image

src/lib/Hydra/Controller/Root.pm
cole-h: Done in 591ae4a0fa02a35ad354ab615cedd1458761af8e -- could you confirm that my implementation does indeed avoid said spookiness?

t/Hydra/Plugin/RunCommand/http.t
```diff
@@ -0,0 +1,60 @@
...
+
+my $get2 = request(GET '/runcommandlog/' . $runlog->uuid);
+ok($get2->is_success, "GET /runcommandlog/{uuid} succeeded.");
+
+my $get3 = request(GET '/runcommandlog/some-invalid-or-nonexistent-uuid');
+ok(!$get3->is_success, "GET'ing invalid uuid failed.");
+
+# NOTE: This is untestable because the runcommand-log.tt page uses JS to fetch the log.
+# my $get4 = request(GET '/build/' . $build->id . '/runcommandlog/some-invalid-or-nonexistent-uuid');
+# ok(!$get4->is_success, "GET /build/{id}/runcommandlog/{uuid} succeeded.");
```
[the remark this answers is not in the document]
cole-h: To get a 404, we'd have to give a properly-formatted UUID to the request. Although the chances of collision are lower than astronomically low, it's still possible that a test fails due to this collision... WDYT?

The rest is easily doable, though.

src/lib/Hydra/Helper/Nix.pm
cole-h: I've added a test for this, but unsure what you mean about testing an empty UUID, now that we've switched to giving the `RunCommandLog` class to the helper function.

src/root/runcommand-log.tt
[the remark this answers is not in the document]
grahamc: Yep!

grahamc (approved):

src/sql/upgrade-81.sql
```diff
@@ -0,0 +1,5 @@
+alter table runcommandlogs add column uuid uuid;
+update runcommandlogs set uuid = gen_random_uuid() where uuid is null;
+alter table runcommandlogs alter column uuid set not null;
+alter table runcommandlogs add constraint RunCommandLogs_uuid_unique unique(uuid);
```
cole-h, line 4: I think the only real blocker at the moment is testing that this migration works as expected -- I added the unique constraint so that we're sure the UUID is indeed unique, but it's been a hot minute since I last used SQL. That's pretty much the only part I haven't tested.

Should be able to get to that on Monday.

hydra-module.nix
```diff
@@ -456,7 +460,8 @@ in
       { path = [ pkgs.bzip2 ];
         script =
           ''
-            find /var/lib/hydra/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            find ${baseDir}/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            find ${baseDir}/runcommand-logs -type f -mtime +3 -size +0c | xargs -r bzip2 -v -f
```
grahamc, line 464: Have we actually tried reading a log which has been bzip2'd? If not, we should probably just drop this line.

src/root/build.tt
```diff
@@ -524,11 +524,29 @@ END;
...
+                [% ELSE %]
+                  Running for [% INCLUDE renderDuration duration = curTime - runcommandlog.start_time %]
+                    [% IF runcommandlog.uuid != undef %]
+                      [% runLog = c.uri_for('/build', build.id, 'runcommandlog', runcommandlog.uuid) %]
+                      <div>
+                        <a class="btn btn-secondary btn-sm" href="[% runLog %]">pretty</a>
+                        <a class="btn btn-secondary btn-sm" href="[% runLog %]/raw">raw</a>
+                        <a class="btn btn-secondary btn-sm" href="[% runLog %]/tail">tail</a>
+                      </div>
+                    [% END %]
```
grahamc, line 547: It looks like this code block can be moved outside this `IF / ELSE / END` and then not duplicated in both branches. I might be missing something, though.

src/lib/Hydra/Plugin/RunCommand.pm
```diff
@@ -160,9 +163,22 @@ sub buildFinished {
...
+        my $logPath = Hydra::Helper::Nix::constructRunCommandLogPath($runlog) or die "RunCommandLog not found.";
+        my $dir = dirname($logPath);
+        my $oldUmask = umask();
+
+        # file: 640, dir: 750
+        umask(0027);
+        make_path($dir);
+
+        open(my $f, '>', $logPath);
+        umask($oldUmask);
```
grahamc, line 175: If we get a die or whatever in make_path / open / etc. we could leave this code path and enter another without resetting the umask. That could create some bizarre bug reports.

How about we wrap this up with Try::Tiny:

```
```
```suggestion
        my $oldUmask = umask();
        try {
            # file: 640, dir: 750
            umask(0027);
            make_path($dir);

            open(my $f, '>', $logPath);
        } finally {
            umask($oldUmask);
        }
```

src/sql/upgrade-81.sql
```diff
@@ -0,0 +1,6 @@
...
+# the unique uuid constraint adds a unique, btree index, so we don't need to create out own:
+#    "runcommandlogs_uuid_unique" UNIQUE CONSTRAINT, btree (uuid)
```
grahamc, line 6:
```suggestion
```

we probably don't need to remember this :).

grahamc (commented):
This is basically ready to merge, but here's a couple final nits.

src/lib/Hydra/Plugin/RunCommand.pm
cole-h: Should the entire  `run3` logic be moved into the try block as well (including the later `close`)?

grahamc (approved):

src/sql/upgrade-81.sql
```diff
@@ -0,0 +1,4 @@
+alter table runcommandlogs add column uuid uuid;
+update runcommandlogs set uuid = gen_random_uuid() where uuid is null;
+alter table runcommandlogs alter column uuid set not null;
+alter table runcommandlogs add constraint RunCommandLogs_uuid_unique unique(uuid);
```
grahamc, line 4: I just noticed that Postgres doesn't automatically create indexes on the _source_ end of a foreign key. A while ago I experimented with adding 120 million runcommandlogs to the table, and this is causing significant performance problems:

```
hydra=# explain analyze select * from runcommandlogs where build_id = 2;
                                                             QUERY PLAN                                                             
------------------------------------------------------------------------------------------------------------------------------------
 Gather  (cost=1000.00..1626001.73 rows=1 width=50) (actual time=3718.813..3719.702 rows=5 loops=1)
   Workers Planned: 2
   Workers Launched: 2
   ->  Parallel Seq Scan on runcommandlogs  (cost=0.00..1625001.62 rows=1 width=50) (actual time=3656.869..3717.351 rows=2 loops=3)
         Filter: (build_id = 2)
         Rows Removed by Filter: 39999999
 Planning Time: 0.251 ms
 Execution Time: 3719.737 ms
(8 rows)
```

We can see there isn't an existing index on the build_id:

```
hydra=# \d+ runcommandlogs;
                                                   Table "public.runcommandlogs"
    Column    |  Type   | Collation | Nullable |                  Default                   | Storage  | Stats target | Description 
--------------+---------+-----------+----------+--------------------------------------------+----------+--------------+-------------
 id           | integer |           | not null | nextval('runcommandlogs_id_seq'::regclass) | plain    |              | 
 job_matcher  | text    |           | not null |                                            | extended |              | 
 build_id     | integer |           | not null |                                            | plain    |              | 
 command      | text    |           | not null |                                            | extended |              | 
 start_time   | integer |           |          |                                            | plain    |              | 
 end_time     | integer |           |          |                                            | plain    |              | 
 error_number | integer |           |          |                                            | plain    |              | 
 exit_code    | integer |           |          |                                            | plain    |              | 
 signal       | integer |           |          |                                            | plain    |              | 
 core_dumped  | boolean |           |          |                                            | plain    |              | 
 uuid         | uuid    |           | not null |                                            | plain    |              | 
Indexes:
    "runcommandlogs_pkey" PRIMARY KEY, btree (id)
    "runcommandlogsuuid" btree (uuid)
Check constraints:
    "runcommandlogs_end_time_has_start_time" CHECK (end_time IS NULL OR start_time IS NOT NULL)
    "runcommandlogs_not_started_no_exit_time_no_code" CHECK (start_time IS NOT NULL OR end_time IS NULL AND error_number IS NULL AND exit_code IS NULL AND signal IS NULL AND core_dumped IS NULL)
Foreign-key constraints:
    "runcommandlogs_build_id_fkey" FOREIGN KEY (build_id) REFERENCES builds(id) ON DELETE CASCADE
Access method: heap
```

I tried adding one with:

```suggestion
alter table runcommandlogs add constraint RunCommandLogs_uuid_unique unique(uuid);
create index IndexRunCommandLogsOnBuildID on RunCommandLogs(build_id);
```

and now we're getting very efficient queries:

```
hydra=# explain analyze select * from runcommandlogs where build_id = 2;
                                                                  QUERY PLAN                                                                  
----------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using indexruncommandlogsonbuildid on runcommandlogs  (cost=0.57..4.58 rows=1 width=50) (actual time=0.037..0.039 rows=5 loops=1)
   Index Cond: (build_id = 2)
 Planning Time: 0.172 ms
 Execution Time: 0.051 ms
(4 rows)
```

Can you alter hydra.sql and this migration to add this index while we're in here?

grahamc (approved):

cole-h:
Thanks @dasJ for the initial work on this :)

PR #1138, opened 2022-01-29
Add cache_control header when saving NARs to S3

andreabedini:
Otherwise S3 will set no cache headers at all.

The max-age is quite generous (1 year) and the content is marked as
immutable (so clients doesn't not have to revalidate).


grahamc:
Unfortunately this is probably not doing what you expect. This S3Backup script is a tool for backing up Hydra to an additional S3 bucket. It isn't responsible for generally uploading build results to a cache. I am not sure anyone uses it or has used it in years, and I somewhat doubt it works today. These changes would probably need to be made in Nix itself. Somewhere in here: https://github.com/NixOS/nix/blob/master/src/libstore/s3-binary-cache-store.cc

grahamc:
(reopened in case you actually do use the S3Backup plugin)

andreabedini:
Right, thanks @grahamc, I'll dig into `s3-binary-cache-store.cc` and perhaps open another PR.

Issue #606, opened 2018-10-25
Gitlab support


Ericson2314:
It would be nice if hydra supported GitLab in all the ways (3 plugins, I believe) it supports GitHub.

nlewo:
@Ericson2314 I have a PR to add the Gitlab MR support. Do not hesitate to review it:)

Ericson2314:
@nlewo oh, great!

Ericson2314:
https://github.com/NixOS/hydra/pull/592 is said PR. (I hope to get to this quite soon, but linking it here for posterity.)

globin:
#649 for gitlab status support.

tfc:
Hey @globin i have a question about the status support plugin:

https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Plugin/GitlabStatus.pm#L63 this line does not work with SSH checkout URLs (like "git@mygitlab.com:foo/bar.git") but only with http(s) URLs.

So i have the impression that you authenticate via HTTP(S), but i would need that for private repositories. Is that a use case that we share?
For now, i am running a patched version of the plugin that simply hardcodes the gitlab server as we have only a single one. But i'd prefer to not patch it and authenticate properly via HTTPS.

hacker1024:
> Hey @globin i have a question about the status support plugin:
> 
> https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Plugin/GitlabStatus.pm#L63 this line does not work with SSH checkout URLs (like "[git@mygitlab.com](mailto:git@mygitlab.com):foo/bar.git") but only with http(s) URLs.
> 
> So i have the impression that you authenticate via HTTP(S), but i would need that for private repositories. Is that a use case that we share? For now, i am running a patched version of the plugin that simply hardcodes the gitlab server as we have only a single one. But i'd prefer to not patch it and authenticate properly via HTTPS.

I just ran into this myself. The offending line is [here](https://github.com/NixOS/hydra/blob/c1a5ff3959f8d3d0f13295d29a02084e14dff735/src/lib/Hydra/Plugin/GitlabStatus.pm#L69C37-L69C37) - it tries to parse the SSH host as a URI.

An easy fix would be to use another string input for the domain.

PR #1142, opened 2022-02-06
GitlabPulls plugin rework

### LINKED ISSUE #655: gitlabpulls: Project Not Found
I've configured a jobset to use the github pulls api and I've used the value syntax specified in the (inline) documentation: `https://gitlab.com projectId`.  

The project is non-public, however, the hydra user has access to the repository and successfully executes other jobsets that run on branch updates.

This job returns:
`Error pulling from the gitlab pulls API: {"message":"404 Project Not Found"}`
Is there a workaround?

amerocu:
I have done a bit of rework on the plugin because
- it doesn't work out of the box, even for public repositories.
- it was not clear where to overlay without deep knowledge of Hydra and GitLab API.

I know that the change is pretty big, any feedback is really welcome.

- fixes #655
- allow cloning with ssh for private repositories as noted in #606
- improves towards a full integration as asked in #606

### DIFF
--- a/doc/manual/src/plugins/README.md
+++ b/doc/manual/src/plugins/README.md
@@ -162,7 +162,51 @@ Create jobs based on open gitlab pull requests.
 
 ### Configuration options
 
-- `gitlab_authorization.<projectId>`
+Gitlab need you to authenticate to call the Gitalb API (even if the repository is public).
+
+Access token can be specified as global for all Gitlab call or for a specific project or group.
+
+- `gitlab_authorization.access_token`
+- `gitlab_authorization.projects.<projectId>.access_token`
+
+How to get the token at [Gitlab API Authorization](https://docs.gitlab.com/ee/api/#authentication).
+
+Example:
+
+```xml
+<gitlab_authorization>
+  access_token = "<gitlab_secret_token>"
+  <projects>
+    <31319625>
+      access_token = "<project_secret_token>"
+    </31319625>
+  </projects>
+</gitlab_authorization>
+```
+
+### Project configuration
+
+The declarative project has to be cofigured with:
+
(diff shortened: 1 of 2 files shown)


amerocu:
Some images to easily reproduce, the "project_id" is public, but the token is expired.

Screenshot from 2022-02-06 21-34-09
Screenshot from 2022-02-06 21-34-23

nlewo:
> it doesn't work out of the box, even for public repositories.

Hm, i'm suprised because it works for me on a Gitlab public repository:
- https://gitlab.com/simple-nixos-mailserver/nixos-mailserver
- https://hydra.nix-community.org/project/simple-nixos-mailserver

This doesn't mean i'm against a refactoring! I initially added this plugin without any perl knowledge by copying the GitHub plugin.

amerocu:
I think two are the cases
1. You have some custom patch over the plugin.
2. I'm still missing some key part on how make the plugin work.

Can you share the project configuration? 
Where do you specify how the job is declared?

nlewo:
We are not using any specific patch. 
The project configuration is in the [nix-community infra repository](https://github.com/nix-community/infra/blob/c44dc42b697d9abc594555bbb56270c80b93c30e/services/hydra/declarative-projects.nix#L19).
The declarative project configuration is in the [nixos-mailserver repository](https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/tree/master/.hydra).

grahamc:
I can't review this properly today, but a couple things I'm noticing:

1. Documentation, yes! Thank you!
2. Tests: I typically look to see an increase in test coverage for PRs that are changing existing code. Today's Hydra coverage is much better than it was a year ago, but plugins in particular are suffering from low coverage. If you could please add test suite coverage to GitlabPulls,that would be great.

PR #1144, opened 2022-02-07
Eval event cleanup

grahamc:
Document the eval_* events, sort some imports and subscriptions.

Note that I'm immediately breaking the documented API in #1145.

PR #1145, opened 2022-02-07
`eval_*` events: change the API to include IDs and use \t separators

grahamc:
Needs:

- [ ] changelog

The old format for these events used the two-character string `\t` as a separator between the fields. I wasn't going to break these event's message format until I saw this and assumed nobody used it. I assume if nobody reported the bug nobody used it. It is obviously a bug that it isn't a tab character. If somebody did depend on such an obvious bug without reporting it, well, oops.

This PR lets plugins attach to evalStarted, evalCached, evalFinished, evalFailed, evalAdded events. This is useful for sending status events to, eg, GitHub (PR soon.)

Note this PR includes #1144 which is fixups originally part of this branch.

### DIFF
--- a/doc/manual/src/notifications.md
+++ b/doc/manual/src/notifications.md
@@ -75,7 +75,7 @@ It is possible for subsequent deliveries of the same `build_finished` data to im
 
 ### `eval_failed`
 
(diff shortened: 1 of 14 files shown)


cole-h (approved):
I really like these eval events -- gonna be some cool stuff coming up!

PR #1146, opened 2022-02-08
GitHubStatus: Gently refactor

grahamc:
Every commit here should be fairly obvious to verify, even though that doesn't mean much with Perl :).

I tried to move independent chunks of code to its own sub and write a test for that sub. I have not tried it for real.

### DIFF
--- a/t/Hydra/Plugin/GithubStatus/GithubStatus.t
+++ b/t/Hydra/Plugin/GithubStatus/GithubStatus.t
@@ -86,4 +86,41 @@ subtest "statusBody" => sub {
     );
 };
 
+subtest "extractGithubArgsFromFlake" => sub {
+    my @failingInputs = (
+        '',
+        'git@github.com:NixOS/hydra.git'
+    );
+    for my $input (@failingInputs) {
+        my $githubArgs = Hydra::Plugin::GithubStatus::extractGithubArgsFromFlake($input);
+        ok(!defined($githubArgs), "$input is not an extractable flake ref");
+    };
+
+    my $passingInputs = {
+        'github:NixOS/hydra/90aa8592925a688c37e8d93822d19bd74e25fca7' => {
+            owner => "NixOS",
+            repo => "hydra",
+            rev => "90aa8592925a688c37e8d93822d19bd74e25fca7"
+        },
+        'git+ssh://git@github.com/NixOS/hydra?rev=90aa8592925a688c37e8d93822d19bd74e25fca7' => {
+            owner => "NixOS",
+            repo => "hydra",
+            rev => "90aa8592925a688c37e8d93822d19bd74e25fca7"
+        },
+        'git+ssh://git@github.com/NixOS/hydra?randomjunkseemsokayhererev=90aa8592925a688c37e8d93822d19bd74e25fca7' => {
+            owner => "NixOS",
+            repo => "hydra",
+            rev => "90aa8592925a688c37e8d93822d19bd74e25fca7"
+        },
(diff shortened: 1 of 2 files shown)


t/Hydra/Plugin/GithubStatus/GithubStatus.t
```diff
@@ -12,6 +12,41 @@ my $builds = $ctx->makeAndEvaluateJobset(
...
+        "continuous-integration/hydra:my-job-name:${build_id}",
+        "An empty configuration produces a default result"
+    );
+
+    is(
+        Hydra::Plugin::GithubStatus::calculateContext($build, "my-job-name", {
+            useShortContext => 1
+        }),
+        "ci/hydra:my-job-name:${build_id}",
+        "useShortContext=1 uses abbreviates continuous-integration"
```
cole-h, line 30: Nit:

```suggestion
        "useShortContext=1 abbreviates continuous-integration"
```

t/Hydra/Plugin/GithubStatus/GithubStatus.t
```diff
@@ -12,6 +12,41 @@ my $builds = $ctx->makeAndEvaluateJobset(
...
+        "continuous-integration/hydra:my-job-name",
+        "excludeBuildFromContext=1 removes the build ID"
+    );
+
+    is(
+        Hydra::Plugin::GithubStatus::calculateContext($build, "my-job-name", {
+            context => "my special context"
+        }),
+        "my special context",
+        "context=... replaces  any other context"
```
cole-h, line 46: Maybe "setting context replaces..."?

```suggestion
        "Manually setting the context replaces any other context"
```

t/Hydra/Plugin/GithubStatus/GithubStatus.t
```diff
@@ -86,4 +86,41 @@ subtest "statusBody" => sub {
...
 };
 
+subtest "extractGithubArgsFromFlake" => sub {
+    my @failingInputs = (
+        '',
+        'git@github.com:NixOS/hydra.git'
+    );
+    for my $input (@failingInputs) {
+        my $githubArgs = Hydra::Plugin::GithubStatus::extractGithubArgsFromFlake($input);
+        ok(!defined($githubArgs), "$input is not an extractable flake ref");
```
cole-h, line 96: Just to make it more obvious e.g. in the case of the empty string input:

```suggestion
        ok(!defined($githubArgs), "input '$input' is not an extractable flake ref");
```

t/Hydra/Plugin/GithubStatus/GithubStatus.t
```diff
@@ -123,4 +123,48 @@ subtest "extractGithubArgsFromFlake" => sub {
...
 };
 
+subtest "extractGithubArgsFromInput" => sub {
+    my @failingURIs = (
+        '',
+        'a static string without slashes'
+    );
+    for my $uri (@failingURIs) {
+        my $githubArgs = Hydra::Plugin::GithubStatus::extractGithubArgsFromInput($uri, "the-revision");
+        is($githubArgs, undef, "$uri is not a parsable GitHub URI");
```
cole-h, line 133: Same here: to make the empty string more noticeable in the output:

```suggestion
        is($githubArgs, undef, "uri '$uri' is not a parsable GitHub URI");
```

Also, maybe this should be `ok(!defined($githubArgs), ....)` as earlier? IIRC, comparing to `undef` and using `defined()` have slightly different semantics.

PR #1147, opened 2022-02-08
JobsetInputAlts: Drop, storing the input's value directly in the Input table

grahamc:
This most certainly breaks the API(s), and potentially other things. We
may want to add compatibility shims to the API.

I haven't tried using the UI manually, though the tests do pass. Also,
I haven't tried writing a migration for this.

Another thing to consider is this will actually cause data loss for older instances. Hydra's newest jobsetevals record with a non-zero altNr is from October of 2017. I'm not sure what we should do about this: this is a pretty nice cleanup, and enables quite a lot more. However, I'm not eager to delete user's data.


src/root/edit-jobset.tt
```diff
@@ -20,18 +20,9 @@
     <td id="[% baseName %]">
       [% IF createFromEval %]
         [% value = (input.uri or input.value); IF input.revision; value = value _ " " _ input.revision; END;
```
cole-h, line 22: Can this entire branch be removed? I can't tell if it's used or not, but it doesn't appear so to me...

cole-h (commented):
I can understand the desire to not delete user's data, but at the same time, it seems like everything that used it has been necessarily fixed.

Maybe we could just make an announcement somewhere about this PR and let it steep for a week or so (or more)?

PR #1148, opened 2022-02-09
captureStdoutStderr*: move to Hydra::Helper::Exec which helps avoid some environment variable fixation problems

grahamc:
Using Hydra::Helper::Nix in tests causes environment variables like HYDRA_DBI to be fixated too early, making us `require` it. A common reason to do this in a test is to access captureStdoutStderrWithStdin. This PR moves those functions to a new helper.


cole-h (approved):
Honestly, it would probably be nice to (in a future PR) move out everything that does (or doesn't) cause this fixation issue. But this is strictly an improvement!

PR #1149, opened 2022-02-09
flake.nix: format with nixpkgs-fmt

grahamc:
I personally use nixpkgs-fmt and am a bit annoyed every time I edit this file and find myself explicitly having to _not_ format the file.


cole-h (approved):
Yes, please!

PR #1150, opened 2022-02-09
LDAP: Test without a VM

grahamc:
Previously, iterating on the LDAP test required many minutes to build hydra, start a VM, and execute. This PR introduces a LDAPContext class which lets us start a LDAP server and test against it.

Includes #1149 #1148


t/lib/LDAPContext.pm
```diff
@@ -0,0 +1,234 @@
...
+    my ($self) = @_;
+
+    expectOkay(1, ("slaptest", "-u", "-F", $self->{"_slapd_dir"}));
+}
+
+sub _spawn {
+    my ($self) = @_;
+
+    my $pid = fork;
+    die "When starting the LDAP server:failed to fork." if not defined $pid;
```
cole-h, line 181:
```suggestion
    die "When starting the LDAP server: failed to fork." if not defined $pid;
```

PR #1151, opened 2022-02-10
ldap: support configurable roles with backwards compatibility

grahamc:
I happen to know an org is using the current LDAP implementation, and they'd be in a bad spot if it broke suddenly. Therefore I took #1134 and ran with it to preserve backwards compatibility. While I did that I noticed #1134 had an issue where a single group could not grant more than one Hydra role. I also took some care to validate that the role mappings map to actually valid roles.

Includes #1150, #1134.


flake.nix
```diff
@@ -520,7 +520,6 @@
                 TextTable
                 UUID4Tiny
                 XMLSimple
-                YAML
```
dasJ, line 523: That's probably not good when you want to keep the legacy behaviour

dasJ (changes requested):

doc/manual/src/configuration.md
```diff
@@ -108,47 +108,70 @@ Using LDAP as authentication backend (optional)
...
+    </credential>
+    <store>
+      class = LDAP
+      ldap_server = localhost
+      <ldap_server_options>
+        timeout = 30
+        debug = 2
+      </ldap_server_options>
+      binddn = "cn=root,dc=example"
+      bindpw = notapassword
```
cole-h, line 140: Maybe there could be a commented out option here that shows readers how exactly to include the bind password from a file outside of the Nix store?

t/Hydra/Config/ldap_role_map.t
```diff
@@ -0,0 +1,242 @@
...
+my $emptyHydraConf = Hydra::Config::loadConfig($emptyHydraConfFile);
+
+my $ldapYamlFile = "$tmpdir/ldap.yaml";
+write_file($ldapYamlFile, <<YAML);
+credential:
+    class: Password
+YAML
+
+subtest "getLDAPConfig" => sub {
+     subtest "No ldap section and an env var gets us legacy data" => sub {
```
cole-h, line 40: A nit, but some of the spacing in this sub looks a bit wonky, e.g. this first line is indented by 5 spaces.

doc/manual/src/configuration.md
```diff
@@ -108,47 +108,70 @@ Using LDAP as authentication backend (optional)
...
-      role_basedn: "ou=groups,dc=example"
-      role_filter: "(&(objectClass=groupOfNames)(member=%s))"
-      role_scope: one
-      role_field: cn
-      role_value: dn
-      role_search_options:
-        deref: always
+This is configured by defining the `<ldap>` block in the configuration file.
+In this block it\'s possible to configure the authentication plugin in the
+`<config>` block, all options are directly passed to `Catalyst::Authentication
```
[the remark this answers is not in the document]
grahamc: Good catch, thanks!

doc/manual/src/configuration.md
grahamc: Good idea. I updated the docs to suggest this, and also made ldap.t run this way.

src/lib/Hydra/Config.pm
```diff
@@ -2,7 +2,165 @@ package Hydra::Config;
...
+        eval {
+            validate_roles($mapping->{$group});
+        };
+        if ($@) {
+            push @errors, "On group '$group': $@";
+        }
+    }
+
+    if (@errors) {
+        die join "\n", @errors;
```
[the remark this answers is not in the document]
grahamc: Done, good idea.

t/Hydra/Config/ldap_role_map.t
grahamc: D'oh.

dasJ (approved):
Worked with both the old and the new config

PR #1154, opened 2022-02-12
display flakeref

tomberek:
Beginning to expose more flake associated details. Did not yet figure out the right approach to expose lockfile information in the inputs tab.


grahamc:
Hmm... it looks like this isn't always right, because Hydra's evaluator will fall back to `.checks` if `hydraJobs` isn't available:

https://github.com/NixOS/hydra/blob/1124230d9f6d482481ddb5316d14e19828a0ab57/src/hydra-eval-jobs/hydra-eval-jobs.cc#L118-L122

I wonder if there is something we could do to deal with that. For example, record the root of the evaluation. This might dovetail nicely with #1143.

PR #1162, opened 2022-02-20
Make `copyClosureTo` take a regular C++ ref to the store

Ericson2314:
This is syntactically lighter wait, and demonstates there are no weird
dynamic lifetimes involved, just regular passing reference to callee
which it only borrows for the duration of the call.


Ericson2314:
CC @grahamc

Ericson2314:
Shall we merge this too? It is a subset of #1180 by coincidence.

Ericson2314:
Thanks!!

PR #1163, opened 2022-02-20
Constituents: Add a GC root for indirect aggregates

grahamc:
@vcunat has noticed periodically aggregate jobs get aborted, it turns out the drv was being GC'd:

```
Feb 20 06:02:22 ceres hydra-queue-runner[3018800]: aborting GC'ed build 167881994
Feb 20 06:03:23 ceres hydra-queue-runner[3018800]: aborting GC'ed build 167882030
Feb 20 06:31:18 ceres hydra-queue-runner[3018800]: aborting GC'ed build 167883929
Feb 20 09:40:05 ceres hydra-queue-runner[2629347]: aborting GC'ed build 167880357
Feb 20 15:17:17 ceres hydra-queue-runner[2629347]: aborting GC'ed build 167893300
Feb 20 15:17:19 ceres hydra-queue-runner[2629347]: aborting GC'ed build 167893322
```

A few days ago @cole-h and I wrote t/queue-runner/direct-indirect-constituents.t and t/jobs/constituents.nix to test direct and indirect aggregates.

Building on top of that test, I wrote t/evaluator/evaluate-constituents-gc.t which evaluates those aggregates and tries to garbage collect the aggregates. That revealed the "indirect" aggregate could be freely garbage collected.

This PR adds a GC root for those indirect aggregates.


src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -504,6 +504,16 @@ int main(int argc, char * * argv)
...
+            /* Register the derivation as a GC root.  !!! This
+                registers roots for jobs that we may have already
+                done. */
+            auto localStore = store.dynamic_pointer_cast<LocalFSStore>();
+            if (gcRootsDir != "" && localStore) {
+                Path root = gcRootsDir + "/" + std::string(baseNameOf((std::string) job["drvPath"]));
+                if (!pathExists(root))
+                    localStore->addPermRoot(localStore->parseStorePath((std::string) job["drvPath"]), root);
+            }
+
```
grahamc, line 516: Pretty much pasta'd from https://github.com/NixOS/hydra/blob/290e0653ad0121b80e53dfc0a5f9eed7540049ff/src/hydra-eval-jobs/hydra-eval-jobs.cc#L201-L209 but I'm anxious about creating a bigger diff here without any test framework on the C++ code.

cole-h: I wonder what it would take to implement a test framework...

grahamc: Yeah, good question. Here's a "My First Test" level of implementation: https://github.com/andir/hydra/commit/ba51556573e983c0baac84cf009b725769a1cac1

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -504,6 +504,16 @@ int main(int argc, char * * argv)
...
             job.erase("namedConstituents");
 
+            /* Register the derivation as a GC root.  !!! This
+                registers roots for jobs that we may have already
+                done. */
+            auto localStore = store.dynamic_pointer_cast<LocalFSStore>();
+            if (gcRootsDir != "" && localStore) {
+                Path root = gcRootsDir + "/" + std::string(baseNameOf((std::string) job["drvPath"]));
+                if (!pathExists(root))
+                    localStore->addPermRoot(localStore->parseStorePath((std::string) job["drvPath"]), root);
```
firestack, line 514: https://github.com/nlohmann/json#implicit-conversions
this says that we shouldn't use implicit casting for getting json values out of a json object, so likely this should be changed to `job["drvPath"].get<std::string>()`


Maybe something like the following?
```suggestion
                auto drvPath = job["drvPath"].get<std::string>();
                Path root = gcRootsDir + "/" + std::string(baseNameOf(drvPath));
                if (!pathExists(root))
                    localStore->addPermRoot(localStore->parseStorePath(drvPath), root);
```

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -504,6 +504,16 @@ int main(int argc, char * * argv)
 
             job.erase("namedConstituents");
 
+            /* Register the derivation as a GC root.  !!! This
+                registers roots for jobs that we may have already
+                done. */
+            auto localStore = store.dynamic_pointer_cast<LocalFSStore>();
+            if (gcRootsDir != "" && localStore) {
+                Path root = gcRootsDir + "/" + std::string(baseNameOf((std::string) job["drvPath"]));
```
firestack, line 512: I can't find where the `Path` type comes from, but I'd ask if this is the proper way to concatenate paths for this type, as there may be a default path separator character we'd want to use? 

if this path came from `std::filesystem::path` we'd be able to use the `/ operator` overload to concatenate the path elements and I'd assume that this path type would have something similar, or some expectation of what "proper data" would look like.

grahamc: Good question. It looks like this is a `nix::Path`, but I couldn't find where that was defined.

grahamc: It looks like `gcRootsDir` could be made a std::filesystem::path, though? I don't know the right way to do that.

src/hydra-eval-jobs/hydra-eval-jobs.cc
grahamc: I applied this patch locally and everything looks good!

src/hydra-eval-jobs/hydra-eval-jobs.cc
cole-h: (FWIW, `Path` is defined here: https://github.com/NixOS/nix/blob/5b809f9e0e0fe84304c2ae0f5f7b2d4db02565ad/src/libutil/types.hh#L25)

src/hydra-eval-jobs/hydra-eval-jobs.cc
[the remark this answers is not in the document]
grahamc: I'm pretty happy where this code is after accepting the suggestion in the other comment. How would you check the variable? I don't mind adding one.

src/hydra-eval-jobs/hydra-eval-jobs.cc
Ericson2314: Can we pull this out in a function?

Ericson2314: If the issue is the variables it closes over, we can do a C++ lambda too I am happy to help.

Issue #1164, opened 2022-02-20
Deduplicate code with Nix


Ericson2314:
Currently Hydra duplicates lots of functionality that is in Nix. This is bad because it leads to either the Nix or Hydra versions of a feature being undertested and broken (e.g. remote builds are very finicky), or missing features on one side (Hydra doesn't support `ssh-ng://`).

The solution is to dedup, fixing both problems, and lowing our maintenance burden across the board.

I will try to break this down into steps, updating this list accordingly

- [ ] Generic SSH logic
   - Nix: `src/libstore/ssh.cc` in Nix
   - Hydra: top of `src/hydra-queue-runner/build-remote.cc`

   Notable differences: Hydra passes a number of extra flags to ssh, evidently based on sysadmin war stories.

- [ ] Legacy SSH protocol
   - Nix: `src/libstore/legacy-ssh-store.cc` in Nix
   - Hydra: Much of `src/hydra-queue-runner/build-remote.cc`

   Notable differences: Hydra doesn't issue builds blocking and synchronously, so it can cancel them and handle logs as they occur. Probably other things, TBD.

CC @rickynils @Ma27 @grahamc

Ericson2314:
Note to fix this, it would be very convenient to use a local build of Nix during Hydra development. I will try to hack that up.

Ericson2314:
https://discourse.nixos.org/t/developing-a-system-that-replaces-nix-remote-build/22388/11 lists a huge number of issues with the current remote building situation that makes using Nix in large institutional contexts difficult. CC @elaforge 

The immediate goals of this issue are to shrink the amount of the code an increase testing coverage by moving us towards a world where hydra and remote building both battle-test the same code paths. But a further goal would be to, be improving the quality and flexibility of the build scheduling code, make experimentation with the alternative architectures suggested in that that thread easier.

Ericson2314:
#1340 #1339 got us a big closer on this too.

PR #1165, opened 2022-02-20
Begin factoring out the protocol code

Ericson2314:
After a few preparatory moves, this does `cmdQueryValidPaths`

See each commit for details and easier reviewing.

Progress towards https://github.com/NixOS/hydra/issues/1164

This targets `nix-next` not `master`, so it is fine that this depends on an unreleased version of Nix. Only the Nix PR needs to be merged first.

Depends on https://github.com/NixOS/nix/pull/6134
~~Depends on #1324~~

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -8,6 +8,7 @@
 #include "build-result.hh"
 #include "path.hh"
 #include "serve-protocol.hh"
+#include "serve-protocol-impl.hh"
 #include "state.hh"
 #include "current-process.hh"
 #include "processes.hh"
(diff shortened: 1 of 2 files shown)


edolstra:
I think it's better to start using the `Store` abstraction. I started work on this a while ago, here is how far I got: https://pastebin.com/SA5TStt9. The remaining problems are lack of cancellation and some issues with logging.

Ericson2314:
@edolstra I agree in principle, but I am very afraid of making too many changes too quickly.

With this approach, I can *and preserve* things like the fact the the hydra `queryPathInfo` locks the paths while the `LegacySSHStore` one doesn't.

Long term, yes, I absolutely do want to use the store abstraction! But I want to get there very incrementally to avoid any regressions.

Ericson2314:
It's passing locally! (CI should soon catch up :))

src/hydra-queue-runner/build-remote.cc
```diff
@@ -215,6 +215,9 @@ static void handshake(Machine::Connection & conn, unsigned int repeats)
         throw Error("unsupported ‘nix-store --serve’ protocol version on ‘%1%’", conn.machine->sshName);
     if (GET_PROTOCOL_MINOR(conn.remoteVersion) < 3 && repeats > 0)
         throw Error("machine ‘%1%’ does not support repeating a build; please upgrade it to Nix 1.12", conn.machine->sshName);
+
+    // Do not attempt to speak a newer version of the protocol
+    conn.remoteVersion = std::min(conn.remoteVersion, our_version);
```
Ericson2314, line 220: I am shocked we got away with not having this...ever! If the two sides don't agree on a version, how can anything work?!

PR #1161, opened 2022-02-19
flake: inputs.nix: 2.6.0 -> 2.6.1

Ma27:
cc @grahamc (was about to push it to the previous PR when you merged it :sweat_smile: )

Issue #1166, opened 2022-02-23
Staging branch for hydra


Ericson2314:
I have been thinking, how to land #1164?

 - Hydra just got on a stable version of Nix, it would be a shame to get off
 - It would be risky to make all the nix changes, and just *hope* they were what Hydra needs
 - Could backport Nix changes to stable branch, but that is a bit sketch.

I think the right solution is to have a staging branch of hydra that points to master Nix. Whenever a version of Nix is released, `staging` merged, and Hydra `master` fast-forwards to the new Nix release. Of course, `master` can be merged into `staging` at any time.

Ericson2314:
CC @rickynils

NickCao:
A staging branch would also be nice for scenarios like https://github.com/NixOS/nixpkgs/pull/158455#issuecomment-1038532832.

Ericson2314:
Per #1171, we need staging branches for each release of Nix ahead of the one hydra uses, in addition one for Nix master, too.

PR #1171, opened 2022-03-10
fix build against nix 2.7

ajs124:
Includes  #1161 (@Ma27)

not particularly well tested, but builds and tests pass (I think)


src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -157,7 +157,8 @@ static void worker(
                 if (drv->querySystem() == "unknown")
                     throw EvalError("derivation must have a 'system' attribute");
 
-                auto drvPath = drv->queryDrvPath();
+                auto localStore = state.store.dynamic_pointer_cast<LocalFSStore>();
+                auto drvPath = localStore->printStorePath(drv->requireDrvPath());
```
Ericson2314, line 161: `localStore` might be `null`, please do not do this way.
```suggestion
                auto drvPath = state.store->printStorePath(drv->requireDrvPath());
```

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -201,7 +202,6 @@ static void worker(
                 /* Register the derivation as a GC root.  !!! This
                    registers roots for jobs that we may have already
                    done. */
-                auto localStore = state.store.dynamic_pointer_cast<LocalFSStore>();
```
Ericson2314, line 204: Please put back

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -210,7 +210,7 @@ static void worker(
 
                 nlohmann::json out;
                 for (auto & j : outputs)
-                    out[j.first] = j.second;
+                    out[j.first] = localStore->printStorePath(j.second);
```
Ericson2314, line 213:
```suggestion
                    out[j.first] = state.store->printStorePath(j.second);
```

Ma27 (approved):
Deployed to my own Hydra, seems to work fine :+1:

Ericson2314 (approved):

Ericson2314:
@edolstra Feel comfortable merging this one?

cole-h:
I feel like we should hold off on this, due to the recent issues with Hydra that maybe are connected to the bump to 2.6 (e.g. see the Infra and Infra alerts channels on Matrix, where this has been a ~hot topic).

Ericson2314:
Oh, I didn't know about that. Bummer. Perhaps we need a way to disable non-blocking GC until the kinks are worked out? (Note, they could be Hydra side, e.g. a lack of temp roots, too.)

Per #1166 I would say we need a staging branch for every version of Nix newer than the on on `master`, in addition to Nix HEAD.

So we could revert 2,6 on master, but then revert the revert on `staging-nix-2.6`, and merge that and this into `staging-nix-2.7`, and then make `staging-nix-head` or so on top of that.

edolstra:
@cole-h Do you have more info on those issues?

grahamc:
@edolstra could you, @rbvermaa, and I do a call some time to go over them? They're a bit hard to track down and I think we might have an easier time sorting through them together.

rbvermaa:
I would prefer it if we can allow Hydra to build against 2.4 and up. We had to revert back to 2.4, because we were getting builds that kept hanging, which significantly disrupted our builds.

Ericson2314:
@rbvermaa The Nix internals change too much for that to be possible at this time. But we could keep the branches tied to older Nix around afterwords.

rbvermaa:
It seems surprising to me to not support the versions of nix that are provided in NixOS stable channel.

Ericson2314:
@rbvermaa Hydra has always been built against a single version of Nix. The daemon protocol is meant to be interoperable, so other Nix versions for the builders will work. (And I have unmerged PRs to test that interoperability.)

Hydra previously was pinned to random Nix versions. In recent months, it is pinned to a stable version of Nix. This is definitely an improvement.

grahamc:
Applied in #1188.

PR #1173, opened 2022-03-10
WIP: hydra-queue-runner metrics

cole-h:
There are probably better ways to achieve this (and will likely need to
be refactored a bit to support further metrics.

---

This uses the https://github.com/jupp0r/prometheus-cpp library.

---

```
$ curl http://localhost:8080/metrics
# HELP exposer_transferred_bytes_total Transferred bytes to metrics services
# TYPE exposer_transferred_bytes_total counter
exposer_transferred_bytes_total 1639.000000
# HELP exposer_scrapes_total Number of times metrics were scraped
# TYPE exposer_scrapes_total counter
exposer_scrapes_total 2.000000
# HELP exposer_request_latencies Latencies of serving scrape requests, in microseconds
# TYPE exposer_request_latencies summary
exposer_request_latencies_count 2
exposer_request_latencies_sum 157.000000
exposer_request_latencies{quantile="0.500000"} 70.000000
exposer_request_latencies{quantile="0.900000"} 70.000000
exposer_request_latencies{quantile="0.990000"} 70.000000
# HELP hydra_queue_runner_running Whether the queue runner is currently running
# TYPE hydra_queue_runner_running gauge
hydra_queue_runner_running 1.000000
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -565,6 +565,7 @@
               (if lib.versionAtLeast lib.version "20.03pre"
               then nlohmann_json
               else nlohmann_json.override { multipleHeaders = true; })
+              prometheus-cpp
             ];
 
           checkInputs = [
--- a/flake.nix
+++ b/flake.nix
@@ -1,10 +1,13 @@
 {
   description = "A Nix-based continuous build system";
 
+  # FIXME: All the pinned versions of nix/nixpkgs have a broken foreman (yes,
+  # even 2.7.0's Nixpkgs pin).
+  inputs.newNixpkgs.url = "github:NixOS/nixpkgs/nixos-21.11";
   inputs.nixpkgs.follows = "nix/nixpkgs";
   inputs.nix.url = github:NixOS/nix/2.6.0;
 
-  outputs = { self, nixpkgs, nix }:
+  outputs = { self, newNixpkgs, nixpkgs, nix }:
     let
 
       version = "${builtins.readFile ./version.txt}.${builtins.substring 0 8 self.lastModifiedDate}.${self.shortRev or "DIRTY"}";
@@ -566,7 +569,9 @@
 
           checkInputs = [
             cacert
-            foreman
+            # FIXME: foreman is broken on all nix/nixpkgs pin, up to and
+            # including 2.7.0
+            newNixpkgs.legacyPackages.${final.system}.foreman
             glibcLocales
             netcat-openbsd
             openldap
--- a/src/hydra-queue-runner/hydra-queue-runner.cc
+++ b/src/hydra-queue-runner/hydra-queue-runner.cc
@@ -6,6 +6,10 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 
+#include <prometheus/exposer.h>
+#include <prometheus/registry.h>
+#include <prometheus/gauge.h>
+
 #include "state.hh"
 #include "build-result.hh"
 #include "store-api.hh"
@@ -853,6 +857,25 @@ int main(int argc, char * * argv)
 {
     return handleExceptions(argv[0], [&]() {
         initNix();
+
+        /* Export a simple "up" metric, to allow monitoring that we're
+           still alive. */
+        std::thread([&]() {
+            prometheus::Exposer exposer{"127.0.0.1:8080"};
+
+            // @note it's the users responsibility to keep the object alive
+            auto registry = std::make_shared<prometheus::Registry>();
+
+            auto& running = prometheus::BuildGauge()
+                .Name("hydra_queue_runner_running")
+                .Help("Whether the queue runner is currently running")
+                .Register(*registry);
+
+            exposer.RegisterCollectable(registry);
+            running.Add({}).Set(1);
+
+            while (true) { }
+        }).detach();
 
         signal(SIGINT, SIG_DFL);
         signal(SIGTERM, SIG_DFL);
--- a/flake.lock
+++ b/flake.lock
@@ -16,6 +16,22 @@
         "type": "github"
       }
     },
+    "newNixpkgs": {
+      "locked": {
+        "lastModified": 1646588256,
+        "narHash": "sha256-ZHljmNlt19nSm0Mz8fx6QEhddKUkU4hhwFmfNmGn+EY=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "2ebb6c1e5ae402ba35cca5eec58385e5f1adea04",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-21.11",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
     "nix": {
       "inputs": {
         "lowdown-src": "lowdown-src",
@@ -69,6 +85,7 @@
     },
     "root": {
       "inputs": {
+        "newNixpkgs": "newNixpkgs",
         "nix": "nix",
         "nixpkgs": [
           "nix",
--- a/src/hydra-queue-runner/Makefile.am
+++ b/src/hydra-queue-runner/Makefile.am
@@ -4,5 +4,5 @@ hydra_queue_runner_SOURCES = hydra-queue-runner.cc queue-monitor.cc dispatcher.c
  builder.cc build-result.cc build-remote.cc \
  build-result.hh counter.hh state.hh db.hh \
  nar-extractor.cc nar-extractor.hh
-hydra_queue_runner_LDADD = $(NIX_LIBS) -lpqxx
+hydra_queue_runner_LDADD = $(NIX_LIBS) -lpqxx -lprometheus-cpp-pull -lprometheus-cpp-core
 hydra_queue_runner_CXXFLAGS = $(NIX_CFLAGS) -Wall -I ../libhydra -Wno-deprecated-declarations


cole-h:
Tests are failing due to port conflicts, probably (since it's hardcoded).

flake.nix
```diff
@@ -562,11 +565,14 @@
...
               then nlohmann_json
               else nlohmann_json.override { multipleHeaders = true; })
+              prometheus-cpp
             ];
 
           checkInputs = [
             cacert
-            foreman
+            # FIXME: foreman is broken on all nix/nixpkgs pin, up to and
+            # including 2.7.0
```
grahamc, line 574: Ouch, does this mean foreman is broken today on hydra master? :/. Is there a nixpkgs issue about it?

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -854,6 +858,25 @@ int main(int argc, char * * argv)
     return handleExceptions(argv[0], [&]() {
         initNix();
 
+        /* Export a simple "up" metric, to allow monitoring that we're
+           still alive. */
+        std::thread([&]() {
+            prometheus::Exposer exposer{"127.0.0.1:8080"};
```
grahamc, line 864: I don't think we need to put this in a thread: exposer appears to make its own thread for serving metrics. I think we probably want to put the exposer in to the State struct, along with the various metrics we end up creating.

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -854,6 +858,25 @@ int main(int argc, char * * argv)
...
+            // @note it's the users responsibility to keep the object alive
+            auto registry = std::make_shared<prometheus::Registry>();
+
+            auto& running = prometheus::BuildGauge()
+                .Name("hydra_queue_runner_running")
+                .Help("Whether the queue runner is currently running")
+                .Register(*registry);
+
+            exposer.RegisterCollectable(registry);
+            running.Add({}).Set(1);
```
grahamc, line 875: Prometheus automatically creates an `up` metric for every scrape target, so just by virtue of having an exposer at all will create this `up` :).

grahamc (commented):
Looking like a good start. Let's not start the exposer unless we're actually `state.run`ing: we don't want to start a server just to show the --help output for example (if there were some.).

Maybe in fact we want to move the exposer's starting up in to `state.run`?

cole-h:
The reason I didn't initially put it in `State::run` is because it's (maybe obviously) not running when there's nothing to do. In the h.n.o case, if we were to set up alerting, it wouldn't be a simple `absent(up{job="queue-runner"} == 1)` (or similar) to validate that the queue-runner is actually alive. I seem to recall you saying somewhere that, in a recent incident, the queue-runner service was just not running at all -- creating the metrics this early would tell you that the service is actually living.

Although, maybe this could be supplanted by an alert using the systemd exporter, checking if that service is `running` or `activating` and alerting if not...

flake.nix
cole-h: Yes it does, and not that I know of. However, it is fixed in Nixpkgs master -- but we're pinned to the Nixpkgs that Nix 2.6.0 is using (I tried 2.6.1 and 2.7.0, and their Nixpkgs pins have the same problem).

cole-h:
I've added some commits that:

1) updates prometheus-cpp to 1.0.0 -- this has the benefit of letting us get random ports when setting the port to 0 (hopefully tests should pass now)
2) replaces all aliases with their proper names, because newer nixpkgs have turned them intro `throw`s (this can probably be a separate PR, FWIW -- it was necessary while I was tracking down the glibc issue)
3) moves `registry` to be a member of `State` and creates it upon instantiation of `State`, as well as makes the metrics port configurable via `--port 10001`

Maybe this is in a decent spot? But I'm not a C++ pro, so there are probably some things that can be improved.

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -754,6 +759,15 @@ void State::run(BuildID buildOne)
     if (!lock)
         throw Error("hydra-queue-runner is already running");
 
+    /* Set up simple exporter, to show that we're still alive. */
+    std::string metricsAddress{"127.0.0.1:" + std::to_string(metricsPort)};
+    prometheus::Exposer exposer{metricsAddress};
+    exposer.RegisterCollectable(registry);
+
+    std::cout << "Starting the Prometheus exporter, listening on "
+        << "http://" << metricsAddress << "/metrics"
```
cole-h, line 768: This is wrong in the case that we use port 0, since a random port will be picked from those available (it'll show `http://127.0.0.1:0/metrics`, I think).

cole-h:
Now that https://github.com/NixOS/nixpkgs/pull/163695 has been merged, I think this is in a good place (I've updated the `newNixpkgs` to point to `nixos-unstable-small`, since that has updated to contain that PR).

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -754,6 +759,16 @@ void State::run(BuildID buildOne)
     if (!lock)
         throw Error("hydra-queue-runner is already running");
 
+    /* Set up simple exporter, to show that we're still alive. */
+    std::string metricsAddress{"127.0.0.1"};
+    prometheus::Exposer exposer{metricsAddress + ":" + std::to_string(metricsPort)};
+    auto exposerPort = exposer.GetListeningPorts().front();
+    exposer.RegisterCollectable(registry);
+
+    std::cout << "Starting the Prometheus exporter, listening on "
```
cole-h, line 768:
```suggestion
    std::cout << "Started the Prometheus exporter, listening on "
```

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -875,6 +891,16 @@ int main(int argc, char * * argv)
                     buildOne = *b;
                 else
                     throw Error("‘--build-one’ requires a build ID");
+            } else if (*arg == "--port") {
```
cole-h, line 894: This should also be configurable via the config file, but also override-able here (e.g. the flag should take precedence).

cole-h (commented):
I also want to write a test for this.

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -39,21 +39,25 @@ std::string getEnvOrDie(const std::string & key)
...
+State::State(std::optional<uint16_t> metricsPortOpt)
     : config(std::make_unique<HydraConfig>())
     , maxUnsupportedTime(config->getIntOption("max_unsupported_time", 0))
     , dbPool(config->getIntOption("max_db_connections", 128))
     , maxOutputSize(config->getIntOption("max_output_size", 2ULL << 30))
     , maxLogSize(config->getIntOption("max_log_size", 64ULL << 20))
     , uploadLogsToBinaryCache(config->getBoolOption("upload_logs_to_binary_cache", false))
     , rootsDir(config->getStrOption("gc_roots_dir", fmt("%s/gcroots/per-user/%s/hydra-roots", settings.nixStateDir, getEnvOrDie("LOGNAME"))))
-    , metricsPort(config->getIntOption("queue_runner_metrics_port", metricsPort))
+    , metricsPort(config->getIntOption("queue_runner_metrics_port", 9099))
```
cole-h, line 50: Is there a "better" default port to use?

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -880,7 +884,7 @@ int main(int argc, char * * argv)
         bool unlock = false;
         bool status = false;
         BuildID buildOne = 0;
-        uint16_t metricsPort = 0;
+        std::optional<uint16_t> metricsPortOpt = std::nullopt;
```
cole-h, line 895: Happy to bikeshed on names hehe.

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -742,6 +750,24 @@ void State::unlock()
...
+    std::string metricsAddress{"127.0.0.1"};
+    prometheus::Exposer exposer{metricsAddress + ":" + std::to_string(metricsPort)};
+    auto exposerPort = exposer.GetListeningPorts().front();
+    exposer.RegisterCollectable(registry);
+
+    std::cout << "Started the Prometheus exporter, listening on "
+        << "http://" << metricsAddress << ":" << exposerPort << "/metrics"
+        << std::endl;
+
+    while (true) {};
```
cole-h, line 767: This needs to run forever, otherwise the `prometheus::Exposer` gets dropped at the end of this scope.

src/hydra-queue-runner/hydra-queue-runner.cc
grahamc: Let's use 9198, since hydra-notify uses 9199.

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -37,19 +39,25 @@ std::string getEnvOrDie(const std::string & key)
...
-State::State()
+State::State(std::optional<uint16_t> metricsPortOpt)
     : config(std::make_unique<HydraConfig>())
     , maxUnsupportedTime(config->getIntOption("max_unsupported_time", 0))
     , dbPool(config->getIntOption("max_db_connections", 128))
     , maxOutputSize(config->getIntOption("max_output_size", 2ULL << 30))
     , maxLogSize(config->getIntOption("max_log_size", 64ULL << 20))
     , uploadLogsToBinaryCache(config->getBoolOption("upload_logs_to_binary_cache", false))
     , rootsDir(config->getStrOption("gc_roots_dir", fmt("%s/gcroots/per-user/%s/hydra-roots", settings.nixStateDir, getEnvOrDie("LOGNAME"))))
+    , metricsPort(config->getIntOption("queue_runner_metrics_port", 9099))
```
grahamc, line 50:
```suggestion
    , metricsPort(config->getIntOption("queue_runner_metrics_port", 9198))
```

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -36,20 +38,73 @@ std::string getEnvOrDie(const std::string & key)
...
 {
     hydraData = getEnvOrDie("HYDRA_DATA");
 
     logDir = canonPath(hydraData + "/build-logs");
 
+    if (metricsAddrOpt.has_value()) {
+        metricsAddr = metricsAddrOpt.value();
+    }
+
+
```
cole-h, line 107:
```suggestion
```

PR #1174, opened 2022-03-16
Fixup std::string usage

jonringer:
This fails with nixpkgs' default hardening

```
../libhydra/hydra-config.hh:20:29: error: 'string' was not declared in this scope
   20 |                 line = trim(string(line, 0, line.find('#')));
      |
```


ajs124:
This is also addressed in #1171

jonringer:
Yea, that looks much more thorough of a fixup

PR #1175, opened 2022-03-18
Fix flake.nix to use the non-deprecated perl function

AmandaCameron:
Hydra will fail to build on recent nixpkgs versions due to buildPerlPackage being removed, with a throw saying to use perlPackages.buildPerlPackage.

This is a simple find/replace for final.buildPerlPackage -> final.perlPackages.buildPerlPackage


AmandaCameron:
Bah, this needs some more massaging, now it's complaining about netcat-openbsd being removed. Will push with the relevent changes shortly.

cole-h:
FWIW, these fixes are already included as part of https://github.com/NixOS/hydra/pull/1173.

flake.nix
```diff
@@ -568,7 +568,7 @@
             cacert
             foreman
             glibcLocales
-            netcat-openbsd
+            libressl.nc
```
dasJ, line 571: What's the reason for this? `netcat-openbsd` still exists on master.
Also not too thrilled to add anything from libressl when the rest works with the well-maintained nixpkgs openssl

dasJ (changes requested):

flake.nix
AmandaCameron: netcat-openbsd did not still exist on master when I made this PR, I'm guessing that's since changed. It was an alias to libressl.nc before the alias removal.

AmandaCameron: Looks like it [was removed in 2018, replaced with an alias](https://github.com/NixOS/nixpkgs/commits/nixos-unstable/pkgs/tools/networking/netcat-openbsd/default.nix) then re-added this may

PR #1178, opened 2022-03-20
hydra-update-gc-roots: allow cached refs to the build's jobset

grahamc:
Re-executing this search_related on every access turned out to
create very problematic performance. If a jobset had a lot of
error output stored in the jobset, and there were many hundreds
or thousands of active jobs, this could easily cause >1Gbps of
network traffic.

### DIFF
--- a/src/script/hydra-update-gc-roots
+++ b/src/script/hydra-update-gc-roots
@@ -39,10 +39,7 @@ sub keepBuild {
 
     $build->finished;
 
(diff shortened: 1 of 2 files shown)


grahamc:
This program would be chewing through approximately 1Gbps on h.n.o, and is now taking less than 1GB in total.

cole-h (approved):
55TB in 7 days holy guacamole...

Hats off to you. I don't think I'd have been able to figure out a way to fix that 😆

PR #1180, opened 2022-03-21
Split the `buildRemote` function

thufschmitt:
In an attempt to understand everything that’s going on here − and to make it easier to eventually simplify it − extract a handful of auxiliary functions from the `State::buildRemote` method.

I (hopefully) didn’t substantially change any logic yet, just moved things out of this method, so the behavior should remain exactly the same.


Ericson2314:
Should we merge https://github.com/NixOS/hydra/pull/1162 first?

Need to read not on my phone, but I like this incrementalism. Good first steps before other cleanups!

src/hydra-queue-runner/build-remote.cc
```diff
@@ -175,24 +175,307 @@ StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo> & paths
...
+                inputs.insert(*outPath);
+                basicDrv.inputSrcs.insert(*outPath);
+            }
+        }
+    }
+
+    /* Ensure that the inputs exist in the destination store. This is
+        a no-op for regular stores, but for the binary cache store,
+        this will copy the inputs to the binary cache from the local
+        store. */
```
edolstra, line 238:
```suggestion
    /* Ensure that the inputs exist in the destination store. This is
       a no-op for regular stores, but for the binary cache store,
       this will copy the inputs to the binary cache from the local
       store. */
```

edolstra (approved):
Looks good, though I am reminded of this post by John Carmack on splitting big functions: http://number-none.com/blow/john_carmack_on_inlined_code.html. I.e. having a lot of small functions obfuscates the control flow.

Ericson2314:
@edolstra I think it's instructive per the top of that linked page that John Carmack eventually changed his mind :). These functions are are basically the ones to dedup with Nix per the very cautious incremental plan in https://github.com/NixOS/hydra/issues/1164 ensuring we have no regressions, before finally getting Hydra to use the Store abstraction properly as the last step. I therefore like this a lot.

thufschmitt:
> I am reminded of this post by John Carmack on splitting big functions: http://number-none.com/blow/john_carmack_on_inlined_code.html. I.e. having a lot of small functions obfuscates the control flow.

Thanks for the reference, that’s indeed a very valid point (and a very interesting read :) ) I think we’re fine(ish) here, because most of these functions are conceptually pure (for a very broad definition of pure), or at least idempotent.

thufschmitt:
Mh, just noticed an issue with this, we’re sending the original derivation to the remote builders and not the basic one. Will fix asap

thufschmitt:
> Mh, just noticed an issue with this, we’re sending the original derivation to the remote builders and not the basic one. Will fix asap

OK, fixed in c5922932b4e4308a60c8fa5b671fa18c08f4dbe1

Ericson2314 (approved):

Ericson2314:
I would still like to do this. @edolstra did start a more aggressive rewrite to go directly to using the store abstraction. But that PR is still draft, and to me that adds evidence to my theory that this approach (there is no disagreement on the end goal!) is better:

1. Smaller steps
2. No behavior changes to start
3. Easier to review those smaller steps
4. Easier to revert/bisect if someone goes wrong.

We are all pressed for time, and no one wants to break Hydra / hydra.nixos,org, so I think this is the prudent way forward.

Please, let's do this.

src/hydra-queue-runner/build-remote.cc
```diff
@@ -175,24 +175,301 @@ StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo> & paths
...
+
+    createDirs(dirOf(logFile));
+
+    AutoCloseFD logFD = open(logFile.c_str(), O_CREAT | O_TRUNC | O_WRONLY, 0666);
+    if (!logFD) throw SysError("creating log file ‘%s’", logFile);
+
+    return {std::move(logFile), std::move(logFD)};
+}
+
+void handshake(Machine::Connection & conn, unsigned int repeats)
```
dasJ, line 191: I think it would either be better to rename the function (`nixServeV1Handshake` or something like that) or move it into a namespace/class to make its use more obvious

src/hydra-queue-runner/build-remote.cc
```diff
@@ -175,24 +175,301 @@ StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo> & paths
...
+    if (magic != SERVE_MAGIC_2)
+        throw Error("protocol mismatch with ‘nix-store --serve’ on ‘%1%’", conn.machine->sshName);
+    conn.remoteVersion = readInt(conn.from);
+    if (GET_PROTOCOL_MAJOR(conn.remoteVersion) != 0x200)
+        throw Error("unsupported ‘nix-store --serve’ protocol version on ‘%1%’", conn.machine->sshName);
+    if (GET_PROTOCOL_MINOR(conn.remoteVersion) < 3 && repeats > 0)
+        throw Error("machine ‘%1%’ does not support repeating a build; please upgrade it to Nix 1.12", conn.machine->sshName);
+}
+
+BasicDerivation sendInputs(
```
dasJ, line 206: Same with this function

dasJ (changes requested):
The change looks good itself apart from the naming issue and the conflict that arose

src/hydra-queue-runner/build-remote.cc
thufschmitt: Yup' fair point. I've moved all that new logic to its own namespace to make things clearer

edolstra:
I'm not super in favor of this PR, since it is just a refactoring but it will make #1200 harder to merge.

Ericson2314:
@edolstra I am happy to fix all #1200 conflicts if we merge this. I am very much a fan of this PR!

Ericson2314:
https://github.com/NixOS/hydra/compare/master...obsidiansystems:hydra:split-buildRemote I fixed merge conflicts in this one here, but I could not yet due to limited internet.

Ericson2314:
@dasJ Is the new namespacing sufficient to you to clarify the identifiers?

dasJ:
Yeah I'm happy with it. Could you resolve the conflict? I will pick this PR into my downstream Hydra then and run it for a couple of days to make sure nothing odd happens (although I don't expect a lot of impact)

src/hydra-queue-runner/build-remote.cc
```diff
@@ -175,24 +177,305 @@ StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo> & paths
...
+    Store & destStore,
+    Machine::Connection & conn,
+    unsigned int & overhead,
+    counter & nrStepsWaiting,
+    counter & nrStepsCopyingTo
+)
+{
+    BasicDerivation basicDrv(*step.drv);
+
+    for (auto & input : step.drv->inputDrvs) {
```
teh, line 221: Can this be `const auto&` ? (in line 223 as well maybe?)

src/hydra-queue-runner/build-remote.cc
```diff
@@ -175,24 +177,305 @@ StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo> & paths
     return sorted;
 }
 
+std::pair<Path, AutoCloseFD> openLogFile(const std::string & logDir, const StorePath & drvPath)
```
teh, line 180: `openFile` and some others aren't exported in a header right? In that case maybe move them into an anonymous namespace:

```c++
namespace {
std::pair<Path, AutoCloseFD> openLogFile(const std::string & logDir, const StorePath & drvPath) {
...
}
}
````

See https://stackoverflow.com/questions/4422507/how-are-unnamed-namespaces-superior-to-the-static-keyword

File: src/hydra-queue-runner/build-remote.cc [hunk 6]
```diff
@@ -175,24 +177,305 @@ StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo> & paths
...
+    while (true) {
+        auto storePathS = readString(conn.from);
+        if (storePathS == "") break;
+        auto deriver = readString(conn.from); // deriver
+        auto references = worker_proto::read(localStore, conn.from, Phantom<StorePathSet> {});
+        readLongLong(conn.from); // download size
+        auto narSize = readLongLong(conn.from);
+        auto narHash = Hash::parseAny(readString(conn.from), htSHA256);
+        auto ca = parseContentAddressOpt(readString(conn.from));
+        readStrings<StringSet>(conn.from); // sigs
```
teh, line 342: is the result of sigs discarded intentionally?

Ericson2314: It was already done that way. After we upgrade to Nix 2.19 we can deduplicate this code better.

src/hydra-queue-runner/build-remote.cc
Ericson2314: It's in a namespace now, and I marked it `static` in https://github.com/NixOS/hydra/pull/1313

src/hydra-queue-runner/build-remote.cc
Ericson2314: Done in https://github.com/NixOS/hydra/pull/1313

PR #1191, opened 2022-03-31
Prepare for nixos-search integration

ncfavier:
As suggested in https://github.com/NixOS/nixos-search/issues/125#issue-653965820

Makes `nix run github:nixos/nixos-search#flake-info -- flake github:nixos/hydra` succeed.

The `lastModifiedDate` fill is just so that we can run that command as `flake-info flake ./hydra` too.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,7 +7,7 @@
   outputs = { self, nixpkgs, nix }:
     let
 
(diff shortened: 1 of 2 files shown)


grahamc:
Thanks!

PR #1198, opened 2022-04-08
queue runner: monitor: replace the build load counter with a histogram…

grahamc:
… of build load timings.

Looking to start investigating why sometimes the queue runner loads only between 1-10 builds per second.


grahamc:
I don't know if it works, but it compiles.

src/hydra-queue-runner/metrics.cc
```diff
@@ -0,0 +1,75 @@
...
+            .Add({{"disposition", "cached-failure"}}, prometheus::Histogram::BucketBoundaries{0.05, 0.1, 0.25, 0.5, 1, 2.5})
+    )
+    , queue_build_load_cached_success(
+        queue_build_load_family
+            .Add({{"disposition", "cached-success"}}, prometheus::Histogram::BucketBoundaries{0.05, 0.1, 0.25, 0.5, 1, 2.5})
+    )
+    , queue_build_load_added(
+        queue_build_load_family
+            .Add({{"disposition", "added"}}, prometheus::Histogram::BucketBoundaries{0.05, 0.1, 0.25, 0.5, 1, 2.5})
+    )
```
grahamc, line 44: These are the new metrics introduced in this PR.

src/hydra-queue-runner/metrics.hh
```diff
@@ -0,0 +1,97 @@
...
+
+    PromTimerExactlyOneExit(prometheus::Histogram& default_metric)
+        : histogram(default_metric)
+    {
+    }
+
+    ~PromTimerExactlyOneExit() {
+        if (histogram) {
+            finish(histogram->get());
+        }
```
grahamc, line 65: This produces:

```
In file included from state.hh:19,
                 from queue-monitor.cc:1:
metrics.hh: In lambda function:
metrics.hh:24:26: warning: ‘*((void*)(& timer)+8).std::reference_wrapper<prometheus::Histogram>::_M_data’ may be used uninitialized in this function [-Wmaybe-uninitialized]
   24 |         histogram.Observe(std::chrono::duration<float>(std::chrono::high_resolution_clock::now() - created).count());
      |         ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
queue-monitor.cc:145:14: note: ‘*((void*)(& timer)+8).std::reference_wrapper<prometheus::Histogram>::_M_data’ was declared here
  145 |         auto timer = PromTimerExactlyOneExit(prom.queue_build_load_missed_exit);
      |              ^~~~~
```

src/hydra-queue-runner/metrics.hh
```diff
@@ -0,0 +1,143 @@
...
+
+template<class T>
+class PromTimerSyncLock
+{
+    std::optional<PromTimer> blocked;
+    typename nix::Sync<T, std::mutex>::Lock contained;
+    PromTimer blocking;
+
+public:
+    PromTimerSyncLock(nix::Sync<T>& data, std::string location, prometheus::Family<prometheus::Histogram>& metric_family)
```
andir, line 92: You probably don't want to construct a new string here for every instance. Consider a std::string_view instead or a raw char* depending on what you need it for.

src/hydra-queue-runner/metrics.hh
```diff
@@ -0,0 +1,143 @@
...
+            finish(histogram);
+        }
+    }
+
+public:
+    void finish(prometheus::Histogram& metric)
+    {
+        // Assert that we have not reported any metric yet,
+        // as this timer should only be finish()'d once
+        assert(unsubmitted);
```
andir, line 78: It doesn't seem right to assert here. Log a warning or make it only assert while you execute the unit tests. Usually you don't want to kill the process because some of the metrics arent "correct". Especially given how sad the unit test story is.

edolstra: If this is a logic error that shouldn't happen, then an assertion is fine.

PR #1200, opened 2022-04-13
Support arbitrary stores for building

edolstra:
This removes the duplicate implementation of the `nix-store --serve` client in `buildRemote()`. Instead we now use the `Store` API directly. To be able to support build cancellation, the actual build is performed by a helper process (`hydra-build-step`), which can be killed by `hydra-queue-runner` if the corresponding step is cancelled.

To do:

[ ] Actually support build step cancellation.
[ ] Set step statuses like `ssConnecting`.
[ ] Rip out nar-extractor. Since hydra-build-step is a separate process, this doesn't work anymore. Instead hydra-build-step should populate the local NAR cache.
[ ] Get rid of `Machine` duplication between Hydra and libstore.
[ ] Get rid of duplication between `RemoteResult` and `BuildResult`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -68,6 +68,7 @@ AC_CONFIG_FILES([
   src/hydra-evaluator/Makefile
   src/hydra-eval-jobs/Makefile
   src/hydra-queue-runner/Makefile
+  src/hydra-build-step/Makefile
   src/sql/Makefile
   src/ttf/Makefile
   src/lib/Makefile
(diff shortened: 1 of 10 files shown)


Ericson2314:
https://github.com/NixOS/hydra/compare/use-store-api...obsidiansystems:hydra:use-store-api Fixed conflicts (rather trivially) here.

Ericson2314:
The conflicts between the build remote split PR (now merged into master) and this are fixed.

As expected, this PR obviates most of the split out functions by instead moving functionality to the new executable, so the split out functions are mostly just deleted and the diff is fairly small.

PR #1203, opened 2022-04-29
hydra-eval-jobs: Report forked worker process status information on exception

firestack:
When evaluating jobs using `hydra-eval-jobs` it's possible for the evaluation to fail in various unexpected ways, such as consuming more than the available memory, which in turn can trigger `OOMKiller` on child worker processes of `hydra-eval-jobs`. When `hydra-eval-jobs` attempts to interact with a dead child process by reading or writing to the file descriptors `to`/`stdin` and `from`/`stdout` pipes attached to the child process, [`readLine`](https://github.com/NixOS/nix/blob/ee57f91413c9d01f1027eccbe01f7706c94919ac/src/libutil/util.cc#L387-L405) reaches and throws [`EndOfFile`](https://github.com/NixOS/nix/blob/ee57f91413c9d01f1027eccbe01f7706c94919ac/src/libutil/util.cc#L399) and [the thread which is responsible for that process cleans up and exits, while also notifying any other workers](https://github.com/NixOS/hydra/blob/5c90edd19f1787141ae3d9751f567b4df11fc0fa/src/hydra-eval-jobs/hydra-eval-jobs.cc#L416-L420) to [wakeup and continue execution](https://github.com/NixOS/hydra/blob/5c90edd19f1787141ae3d9751f567b4df11fc0fa/src/hydra-eval-jobs/hydra-eval-jobs.cc#L378), in which they should [exit because the shared state shows that an exception has been raised](https://github.com/NixOS/hydra/blob/5c90edd19f1787141ae3d9751f567b4df11fc0fa/src/hydra-eval-jobs/hydra-eval-jobs.cc#L368-L371).

This snippet is the error information reported in the hydra UI currently, when a worker process dies for any reason.
```
hydra-eval-jobs returned exit code 1:
error: unexpected EOF reading a line
```

This PR attempts to make errors in the worker visible in the hydra UI by running `waitpid` on the child process when there's an error. 
I've attached examples of what the output should look like with this PR.
```
hydra-eval-jobs returned exit code 1:
child process (1354273) exited with status=247
error: unexpected EOF reading a line
```
```
hydra-eval-jobs returned exit code 1:
child process (1353091) killed by signal=9
error: unexpected EOF reading a line
```

### DIFF
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -31,7 +31,7 @@ void check_pid_status_nonblocking(pid_t check_pid) {
 
     int wstatus = 0;
     pid_t pid = waitpid(check_pid, &wstatus, WNOHANG);
-    // -1 = failiure, WNOHANG: 0 = no change
+    // -1 = failure, WNOHANG: 0 = no change
     if (pid <= 0) { return; }
 
     std::cerr << "child process (" << pid << ") ";
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -25,6 +25,28 @@
 
 #include <nlohmann/json.hpp>
 
+void check_pid_status_quick(pid_t check_pid) {
+    // Only check 'initialized' and known PID's
+    if (check_pid <= 0) { return; }
+
+    int wstatus = 0;
+    pid_t pid = waitpid(check_pid, &wstatus, WNOHANG);
+    // -1 = failiure, WNOHANG: 0 = no change
+    if (pid <= 0) { return; }
+
+    std::cerr << "child process (" << pid << ") ";
+
+    if (WIFEXITED(wstatus)) {
+        std::cerr << "exited with status=" << WEXITSTATUS(wstatus) << std::endl;
+    } else if (WIFSIGNALED(wstatus)) {
+        std::cerr << "killed by signal=" << WTERMSIG(wstatus) << std::endl;
+    } else if (WIFSTOPPED(wstatus)) {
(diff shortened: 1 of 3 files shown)


src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -25,6 +25,28 @@
 
 #include <nlohmann/json.hpp>
 
+void check_pid_status_nonblocking(pid_t check_pid) {
+    // Only check 'initialized' and known PID's
+    if (check_pid <= 0) { return; }
+
+    int wstatus = 0;
+    pid_t pid = waitpid(check_pid, &wstatus, WNOHANG);
+    // -1 = failiure, WNOHANG: 0 = no change
```
Mindavi, line 34: Failure

Mindavi (approved):

grahamc (approved):
Looks good, nice work on the test, thanks Kayla!

t/evaluator/evaluate-oom-job.t
```diff
@@ -0,0 +1,33 @@
+use strict;
+use warnings;
+use Setup;
+use Test2::V0;
+use Hydra::Helper::Exec;
+
+my ($systemdrRes) = captureStdoutStderr(3, (
+   "systemd-run", "--user", "--collect", "--scope", "--property", "MemoryMax=25M", "--",
+   "true"
+));
```
grahamc, line 10: Looks like we need to do this to handle the case that systemd-run isn't available:

```suggestion
exec {
  my ($systemdrRes) = captureStdoutStderr(3, (
     "systemd-run", "--user", "--collect", "--scope", "--property", "MemoryMax=25M", "--",
     "true"
  ));
} or do {
  skip_all("systemd-run does not work in this environment");
}
```

Issue #1193, opened 2022-04-01
Feature requests for puppeting Hydra from other CI systems


mikepurvis:
Hello there! My company is on a Nix journey and part of this has been figuring out how Hydra fits into our CI strategy (which is otherwise comprised of Jenkins and GitLab CI pipelines). Overall, we are finding that the Hydra fundamentals are really good, but the frontend isn't a good fit for our needs:

- The projects/jobsets/evals setup gets in the way of the workflow we want. We use flakes exclusively, and currently every new flake tag ends up created as its own jobset, which means that there's only ever a single evaluation of each (since we can't trigger an eval for a particular tag—it's always just "take the latest").
- There's no API for eval progress information, buildstep breakdown, or logs.
- Sometimes it's not convenient to have to explicitly cancel an eval. It would be great to have an option for an API-triggered eval where it relies on an external keepalive, and if that keepalive disappears, it self-cancels.

What we'd _really_ like is to be able to just trigger evals of arbitrary flakes— further bonus if we can upload a flake tarball directly to Hydra over the API rather than it needing to be hosted on a URL or git repo somewhere.

So the question is, is this functionality that Hydra would have any interest in offering, or is it just too far out of scope for Hydra's intended use cases? If out of scope, would it be something that could be reasonable to build ourselves using portions of Hydra? Or would it make more sense to build a whole new thing from scratch and just directly call the `nix` CLI?

Thanks for any thoughts or advice, definitely looking forward to potentially collaborating!

Ericson2314:
I would like Hydra to act as a remote builder, such that one can just submit ad-hoc jobs from the local CLI.

It sounds like that is something that might help you out too.

nlewo:
I'm also interested in such feature (and really like how Hydra is efficient!). 

Currently, i already use Hydra and report status in GitLab/GitHub CI:
1. the repository contains a [declarative Hydra jobset](https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/tree/4ed684481b550a802a9c4af97cdc60354d03a615/.hydra) with the GitLab plugin: each minute, Hydra polls the repository and creates a jobset per pull request (i guess GitLab could use the Hydra API to ask for an eval in order to avoid this 1 minute period)
[...]

PR #1204, opened 2022-05-02
Provide build step information

iwanders:
Hi everyone,

As my colleague @mikepurvis described in https://github.com/NixOS/hydra/issues/1193, we are in the process of incorporating some Nix builds running on Hydra into our Jenkins pipelines. An important aspect of this is making sure that it is clear for developers where to find the logs for build failures, with our transition to Nix we should be able to provide this in a more specific manner than we previously could. However, we could not find any API endpoint that actually provides information about the build steps that make up a build.

I think the (currently undocumented) `<build_id>/api/get-info` [endpoint](https://github.com/NixOS/hydra/blob/master/src/lib/Hydra/Controller/Build.pm#L572-L580) was supposed to provide this information, so in this PR I extended that endpoint to provide information about all the build steps that make up a build. With this newly available information about the results of each individual build step, we will be able to link directly to compilation logs for failed steps.

Overview of changes:
- Extended the `<build_id/api/get-info` endpoint to include build steps, and made it (mostly) a superset of the current `/build/<build_id>` api endpoint, this should allow for uniform handling between the two. The old fields are preserved for backwards compatibility.
- Added a function to convert the build step object into Hash that can be returned through the JSON view.
- Modified `buildToHash` to ensure the `buildstatus` and `priority` fields are always present.
- Used yaml references in the hydra api to ensure fields identical between the build api endpoint and get-info stay in sync.
- [Rendered hydra API](https://editor.swagger.io/?url=https://raw.githubusercontent.com/iwanders/NixOS-hydra/CORE-21733-get-info-buildsteps/hydra-api.yaml) document.

This is the first time I'm using Perl, one of the things I couldn't quite figure out is how the current build api endpoint (so `/build/<build_id`) works. I reused this [buildToHash](https://github.com/NixOS/hydra/blob/5c90edd19f1787141ae3d9751f567b4df11fc0fa/src/lib/Hydra/Controller/API.pm#L23-L43) function, but later noticed that this is not the function that provides that is used in the `/build/<build_id>` endpoint, as the `build_status` key was missing if the build was still ongoing. I _think_ that endpoint is served by this [as_json](https://github.com/NixOS/hydra/blob/5c90edd19f1787141ae3d9751f567b4df11fc0fa/src/lib/Hydra/Schema/Result/Builds.pm#L568-L598) function. But I don't know enough Perl to figure out why we have two, or how I could call the latter to populate the results of the `get-info` endpoint as a starting point and then add the steps to that. It would be the ideal case if `get-info` is a superset of the simple `/build/<build_id>` api, currently it's close, but because there's two conversion functions it's not guaranteed. Ideally, these should be consolidated in the future to ensure the endpoints in the api are as consistent as possible.

Another thing that's worth considering is changing the name from `/build/<build_id>/api/get-info` to `/build/<build_id>/get-info` that would break backwards compatibility, but makes it more in line with the current api naming.

### DIFF
--- a/hydra-api.yaml
+++ b/hydra-api.yaml
@@ -507,6 +507,33 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
+  /build/{build-id}/api/get-info:
+    get:
+      summary: Retrieves info about this build id, providing information about the steps in it.
+      parameters:
+      - name: build-id
+        in: path
+        description: build identifier
+        required: true
+        schema:
+            type: integer
+      responses:
+        '200':
+          description: build
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BuildInfo'
+              examples:
+                build-success:
+                  $ref: '#/components/examples/build-get-info-success'
+        '404':
+          description: build couldn't be found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
   /build/{build-id}/constituents:
     get:
       summary: Retrieves a build's constituent jobs
@@ -673,7 +700,7 @@ components:
           nullable: true
           description: the path to the file to evaluate
           type: string
-        errormsg:
+        errormsg: &schema_errormsg
           nullable: true
           description: contains the stderr output of the nix-instantiate command
           type: string
@@ -813,7 +840,7 @@ components:
             checkinterval:
               description: interval in seconds at which to check the jobset inputs
               type: integer
-            haserrormsg:
+            haserrormsg: &schema_haserrormsg
               description: true if the evaluation had errors
               type: boolean
             nrscheduled:
@@ -896,18 +923,18 @@ components:
           description: The nix store path
           type: string
 
-    Build:
+    Build: 
       type: object
       properties:
         id:
           type: integer
-        starttime:
+        starttime: &schema_starttime
           description: time when build started
           type: integer
-        stoptime:
+        stoptime: &schema_stoptime
           description: time when build ended
           type: integer
-        timestamp:
+        timestamp: &schema_timestamp
           description: time when the build was first created
           type: integer
         jobsetevals:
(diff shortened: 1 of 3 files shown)


ajs124:
Can you maybe write a test for this? I think this would probably fit into `t/Hydra/Controller/API/checks.t`.

Calling `as_json` sounds like it might be useful, yes. I'll try to look into that. As for why we have that and `buildToHash`, that's probably mainly historical reasons. Maybe @grahamc knows, since he only added `ad_json` a few months ago.

iwanders:
Thanks for taking a look at this @ajs124, appreciated!

> Can you maybe write a test for this? I think this would probably fit into `t/Hydra/Controller/API/checks.t`.

Certainly, I've added one to the `t/Hydra/Controller/Build/api.t` file, since that's where the other `build` related api endpoints were being tested. Please see 3f0094d8a7baaeb12e721c74b0f74a958ca34bd7 for this addition.

I also discovered that the swagger didn't like the `<<: *value` expansion type anymore, so I updated the api description to just do direct dereferences in 8423bff765718c16534bf955dac3ee51eff91c91.

iwanders:
I do sometimes see the following warnings
```
( STDERR )  job 87    	(in cleanup) {UNKNOWN}: DBI Exception: DBD::Pg::st DESTROY failed: no connection to the server [for Statement "INSERT INTO users ( emailaddress, password, username) VALUES ( ?, ?, ? )"]  at /nix/store/bc7xdl0w78hmmsa7lfnpvc0rqkcbbcw8-hydra-perl-deps/lib/perl5/site_perl/5.32.1/DBIx/Class/Schema.pm line 1118.
( STDERR )  job 87    	DBIx::Class::Schema::throw_exception(Hydra::Schema=HASH(0x63c70d0), "DBI Exception: DBD::Pg::st DESTROY failed: no connection to t"...) called at /nix/store/bc7xdl0w78hmmsa7lfnpvc0rqkcbbcw8-hydra-perl-deps/lib/perl5/site_perl/5.32.1/DBIx/Class/Storage.pm line 113
( STDERR )  job 87    	DBIx::Class::Storage::throw_exception(DBIx::Class::Storage::DBI::Pg=HASH(0x64374d8), "DBI Exception: DBD::Pg::st DESTROY failed: no connection to t"...) called at /nix/store/bc7xdl0w78hmmsa7lfnpvc0rqkcbbcw8-hydra-perl-deps/lib/perl5/site_perl/5.32.1/DBIx/Class/Storage/DBI.pm line 1623
( STDERR )  job 87    	DBIx::Class::Storage::DBI::__ANON__("DBD::Pg::st DESTROY failed: no connection to the server [for "..., DBI::st=HASH(0x6958f10), undef) called at scripts/hydra-init.t line 0
( STDERR )  job 87    	eval {...} called at scripts/hydra-init.t line 0
```
when running the unit tests, also happens on the `master` branch. Don't think it's particularly concerning, but I thought I'd point it out.

PR #1205, opened 2022-05-02
Move postgres host and port to a single location.

iwanders:
These are some minor changes I did when I started development work on https://github.com/NixOS/hydra/pull/1204. I'm not on NixOS and did most development in a NixOS docker container. Following the steps to bring up the development environment in a Docker container didn't work at first, Postgres doesn't allow running as root, and the NixOS docker image user is root. So in the end I ended up using a second container to run the Postgres server and then pointed the hydra instance running in the NixOS docker container at the external database. This did allow me to easily load a copy of the production database, so in the end it was actually quite convenient for development.

The postgres command line utilities respect the `PGPORT` and `PGHOST` [environment variables](https://www.postgresql.org/docs/current/libpq-envars.html), so we don't need to specify them manually in various files. Putting them in the environment allows overriding them manually when using a different postgres server.

The `LOGNAME` variable is new, but it appeared to be necessary for the queue runner to come up.

This also switches communication to the database during initialisation of the development environment to use tcp instead of a unix domain socket.

I have not tested these changes on a 'real' instance NixOS, only in a NixOS docker container (where Postgres doesn't start), the environment variables looked good inside the nix shell.

fyi @mikepurvis.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -624,7 +624,10 @@
             export HYDRA_HOME="$(pwd)/src/"
             mkdir -p .hydra-data
             export HYDRA_DATA="$(pwd)/.hydra-data"
-            export HYDRA_DBI='dbi:Pg:dbname=hydra;host=localhost;port=64444'
+            export LOGNAME="hydra-dev"
+            export PGPORT=64444
+            export PGHOST="localhost"
+            export HYDRA_DBI="dbi:Pg:dbname=hydra;host=$PGHOST;port=$PGPORT"
 
             popd >/dev/null
           '';
--- a/foreman/start-hydra.sh
+++ b/foreman/start-hydra.sh
@@ -1,15 +1,15 @@
 #!/bin/sh
 
 # wait for postgresql to listen
-while ! pg_isready -h $(pwd)/.hydra-data/postgres -p 64444; do sleep 1; done
+while ! pg_isready ; do sleep 1; done
 
-createdb -h $(pwd)/.hydra-data/postgres -p 64444 hydra
+createdb hydra
 
 # create a db for the default user. Not sure why, but
 # the terminal is otherwise spammed with:
 #
 #     FATAL:  database "USERNAME" does not exist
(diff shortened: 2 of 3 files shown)


Ericson2314:
This looks good to me. Sockets for `PGHOST` should also work.

iwanders:
> Sockets for PGHOST should also work.

I switched it to use tcp to allow easy use of the [postgres](https://hub.docker.com/_/postgres) docker container. Exposing a unix domain socket should work as well yes, but I think exposing tcp ports in Docker is more standard.

Ericson2314:
You did what? This diff makes it seem like it was using TCP before and after the change.

iwanders:
> You did what? This diff makes it seem like it was using TCP before and after the change. I am fine with that, but I would be wary of switching away from UDP if that were the prior default.

My knowledge of postgres is pretty limited, but the original command was `createdb -h $(pwd)/.hydra-data/postgres -p 64444 hydra`. For the `-h` flag the [docs state](https://www.postgresql.org/docs/current/app-createdb.html#:~:text=Specifies%20the%20host%20name%20of%20the%20machine%20on%20which%20the%20server%20is%20running.%20If%20the%20value%20begins%20with%20a%20slash%2C%20it%20is%20used%20as%20the%20directory%20for%20the%20Unix%20domain%20socket.) the following:
> Specifies the host name of the machine on which the server is running. If the value begins with a slash, it is used as the directory for the Unix domain socket.

Since `$(pwd)` starts with a slash, the `-p 64444` is then [interpreted](https://www.postgresql.org/docs/current/app-createdb.html#:~:text=Specifies%20the%20TCP%20port%20or%20the%20local%20Unix%20domain%20socket%20file%20extension%20on%20which%20the%20server%20is%20listening%20for%20connections) as the local Unix domain socket extension.

I think that means that previously communication happened through a unix domain socket? After this change we would be communicating over a tcp socket.

iwanders:
This functionality was implemented in parallel in #1653, since that is merged, I'm closing this stale PR.

PR #1220, opened 2022-06-12
GitInput: Allow setting a umask

dasJ:


edolstra:
Hm, what's the use case for this?

dasJ:
@edolstra probably one of the weirdest use cases tbh.

First off, `initNix()` (from nix) sets a umask which is too restrictive for me, overriding the one that is set from the systemd unit.
I could store the umask before the call, and restore it afterwards but I was too afraid I could break something nix-internal (I really don't know enough about it).

I want a more open umask because I wrote a Hydra plugin that interacts with the git repos as well by fetching into them and creating them if necessary.
This plugin however runs in `hydra-notify` while the regular git fetcher runs in `hydra-evaluator`. For this to work properly, I need to have write permissions for `hydra` group.

There may be a more clever solution I haven't thought of yet, maybe restoring the umask as mentioned above but this is the best I could come up with.

dasJ:
Now that I come to think of it, maybe not as weird as I expected.
My plugin has a `evalAdded` hook that parses local lockfiles and generates git inputs so the SCM diff works.
May be worth supporting for future things like a flake lock extractor?

PR #1221, opened 2022-06-12
sql: migrate timestamps from 4 to 8 bytes

ajs124:


ajs124:
turns out this breaks stuff. who would have thought. luckily it's a long time until 2038.

dasJ:
https://xkcd.com/2697/

PR #1222, opened 2022-06-16
flake: Update Nix to 2.9.1

Ma27:
NOTE: I'm well-aware that we have to be careful with this to avoid new
regressions on hydra.nixos.org, so this should only be merged after
extensive testing from more people.

Motivation: I updated Nix in my deployment to 2.9.1 and decided to also
update Hydra in one go (and compile it against the newer Nix). Given
that this also updates the C++ code in `hydra-{queue-runner,eval-jobs}`
this patch might become useful in the future though.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -39,16 +39,18 @@
         "nixpkgs-regression": "nixpkgs-regression"
       },
       "locked": {
-        "lastModified": 1649172203,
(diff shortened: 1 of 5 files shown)


Ma27:
cc @edolstra @grahamc

PR #1226, opened 2022-06-17
module: only activate pg_trgm extensions on first db setup

SuperSandro2000:
This was a migration step introduced in https://github.com/NixOS/nixpkgs/commit/ce37a040c262aed26e9c6fd63ba527ba1bc028cc almost 3 years ago.
This avoids executing sudo on every start and makes hydra easier to use with sudo's requiretty setting activated.

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -287,9 +287,9 @@ in
             if ! [ -e ${baseDir}/.db-created ]; then
               runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/createuser hydra
               runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/createdb -O hydra hydra
+              echo "create extension if not exists pg_trgm" | runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/psql hydra
               touch ${baseDir}/.db-created
             fi
-            echo "create extension if not exists pg_trgm" | runuser -u ${config.services.postgresql.superUser} -- ${config.services.postgresql.package}/bin/psql hydra
           ''}
 
           if [ ! -e ${cfg.gcRootsDir} ]; then


dasJ:
> This avoids executing sudo

why? I don't see sudo here

SuperSandro2000:
You don't know by chance if runuser is using sudo somehow behind the back?

It was failing for me because I have 

```
Defaults  passwd_timeout=10
Defaults  requiretty
Defaults  use_pty
```

in my sudoers file.

dasJ:
Surely doesn't seem like it:
```
tests/ts/cramfs/doubles:# sudo may use whatever group
tests/ts/cramfs/mkfs-endianness:# sudo may use whatever group
sys-utils/readprofile.8:   sudo readprofile \-M 20
sys-utils/setpriv.1.adoc:If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) (without the *-g* option), try something like:
sys-utils/readprofile.8.adoc:   sudo readprofile -M 20
sys-utils/setpriv.1:If you\(cqre looking for behavior similar to \fBsu\fP(1)/\fBrunuser\fP(1), or \fBsudo\fP(8) (without the \fB\-g\fP option), try something like:
po/cs.po:msgstr " -e, --evenparity        nastaví paritu na sudou\n"
grep: po/cs.gmo: binary file matches
Documentation/howto-tests.txt:  Alternatively using sudo and make:
Documentation/howto-tests.txt:  $ sudo -E make check TS_OPTS="--parallel=1"
Documentation/releases/v2.35-ReleaseNotes:   - fix sudo command line  [Karel Zak]
po-man/pt_BR.po:msgid "   sudo readprofile -M 20\n"
po-man/pt_BR.po:"If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) "
po-man/uk.po:msgid "   sudo readprofile -M 20\n"
po-man/uk.po:msgstr "   sudo readprofile -M 20\n"
po-man/uk.po:"If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) "
po-man/de.po:msgid "   sudo readprofile -M 20\n"
po-man/de.po:msgstr "   sudo readprofile -M 20\n"
po-man/de.po:"If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) "
po-man/de.po:"*sudo*(8) (ohne die Option *-g*) ist, versuchen Sie Folgendes:"
po-man/es.po:msgid "   sudo readprofile -M 20\n"
po-man/es.po:msgstr "   sudo readprofile -M 20\n"
po-man/es.po:"If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) "
po-man/util-linux-man.pot:msgid "   sudo readprofile -M 20\n"
po-man/util-linux-man.pot:"If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) "
po-man/cs.po:msgid "   sudo readprofile -M 20\n"
po-man/cs.po:msgid "If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) (without the *-g* option), try something like:"
po-man/fr.po:msgid "   sudo readprofile -M 20\n"
po-man/fr.po:msgstr "   sudo readprofile -M 20\n"
po-man/fr.po:"If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) "
po-man/sr.po:msgid "   sudo readprofile -M 20\n"
po-man/sr.po:msgstr "   sudo readprofile -M 20\n"
po-man/sr.po:"If you're looking for behavior similar to *su*(1)/*runuser*(1), or *sudo*(8) "
po-man/sr.po:"Ако тражите понашање слично са *su*(1)/*runuser*(1), или *sudo*(8) (без "
```

To be honest, I don't even understand why `runuser` is used in the first place, I thought `su` was the de-facto standard for user-switching.
There is no reason to call sudo at all (even if it was internally used) as the user for the `preStart` is root already (`PermissionsStartOnly=true`).

SuperSandro2000:
I wasn't fully paying attention when doing this PR. NixOS is doing sudo https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/continuous-integration/hydra/default.nix#L321-L325 while the hydra flake is using runuser. So sudo is not used indeed.

dasJ:
No, because in postgres the `root` user doesn't have full privileges while the `postgres` user does:
```
janne@helsinki-hydra-builder01:~ $ sudo psql
Password for user root:
[1]    203942 interrupt  sudo psql
janne@helsinki-hydra-builder01:~ $ sudo -u postgres psql
psql (14.4)
Type "help" for help.

postgres=#
```

We could use `su` in both places or `runuser` in both places but `sudo` does not make sense imo.
Given [this](https://superuser.com/questions/1062576/difference-between-su-c-and-runuser-l-c) I don't care which way we choose as long at's consistent.
One of the nice parts of `runuser` is that it's not a setuid binary and doesn't need an entry in `/run/wrappers` but it's not as widely used as su afaict

Mic92:
utillinux provides runuser: `path = [ pkgs.utillinux ];`

PR #1227, opened 2022-06-18
Add gitea push hook

SuperSandro2000:
The gitea webhook payload is slightly different and not compatible with the github endpoint. This PR adds a new api endpoint `/api/push-gitea` which understands the gitea webhook to trigger jobsets evaluation.

### DIFF
--- a/src/lib/Hydra/Controller/API.pm
+++ b/src/lib/Hydra/Controller/API.pm
@@ -292,6 +292,7 @@ sub push_gitea : Chained('api') PathPart('push-gitea') Args(0) {
 
     my $in = $c->request->{data};
     my $url = $in->{repository}->{clone_url} or die;
+    $url =~ s/.git$//;
     print STDERR "got push from Gitea repository $url\n";
 
     triggerJobset($self, $c, $_, 0) foreach $c->model('DB::Jobsets')->search(
--- a/src/lib/Hydra/Controller/API.pm
+++ b/src/lib/Hydra/Controller/API.pm
@@ -285,6 +285,22 @@ sub push_github : Chained('api') PathPart('push-github') Args(0) {
     $c->response->body("");
 }
 
+sub push_gitea : Chained('api') PathPart('push-gitea') Args(0) {
+    my ($self, $c) = @_;
+
+    $c->{stash}->{json}->{jobsetsTriggered} = [];
+
+    my $in = $c->request->{data};
+    my $url = $in->{repository}->{clone_url} or die;
+    print STDERR "got push from Gitea repository $url\n";
+
+    triggerJobset($self, $c, $_, 0) foreach $c->model('DB::Jobsets')->search(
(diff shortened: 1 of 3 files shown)


src/lib/Hydra/Controller/API.pm
```diff
@@ -285,6 +285,22 @@ sub push_github : Chained('api') PathPart('push-github') Args(0) {
     $c->response->body("");
 }
 
+sub push_gitea : Chained('api') PathPart('push-gitea') Args(0) {
+    my ($self, $c) = @_;
+
+    $c->{stash}->{json}->{jobsetsTriggered} = [];
+
+    my $in = $c->request->{data};
+    my $url = $in->{repository}->{clone_url} or die;
```
DarkKirb, line 294:
```suggestion
    my $url = $in->{repository}->{clone_url} or die;
    $url =~ s/.git$//;
```

DarkKirb (approved):
This PR works for me, but it would be nice if `push_gitea` removed the .git suffix from the repo URL, because I haven’t included them in my gitea flake inputs

src/lib/Hydra/Controller/API.pm
SuperSandro2000: Would that still match if the flake input had ``.git``?

DarkKirb: i would think so, since the flake/url matching is done with LIKE and wildcards before and after the URL

vtrf:
Hello, is this a work in progress? I wanted to use this feature on my Gitea instance and if needed, test it.

SuperSandro2000:
This PR is done and deployed in production.

CyborgPotato:
Can confirm that this PR works in my deployment as well :+1:

RTUnreal:
is there a update when this will be upstreamed? Or will this PR be a patch user need to always apply?

icewind1991:
I had to modify it to strip the `https://` prefix from the url to make it work with `git+ssh` job sources but works fine otherwise

SuperSandro2000:
@icewind1991 can you make a suggestion with that change? Then i'll add it.

@Ericson2314 could you review/merge this?

src/lib/Hydra/Controller/API.pm
```diff
@@ -285,6 +285,23 @@ sub push_github : Chained('api') PathPart('push-github') Args(0) {
...
+    $url =~ s/.git$//;
```
icewind1991, line 295:
```suggestion
    $url =~ s/.git$//;
    $url =~ s/^https?:\/\///;
```

Disclaimer: this is probably the 10th or so line of perl I've ever written :see_no_evil:

Ericson2314: We can follow-up for this part.

gador: @icewind1991 how did you apply your change? Do you use a custom fork of hydra?

PR #1232, opened 2022-07-01
Enable gitea status in flakes

ghost:
Curently the GiteaStatus plugin only works for legacy inputs, as it attempts to read repository metadata from the `jobsetevalinput`s (which I don’t think can be set for flakes).

This PR adds support for many gitea instances by checkng whether the domain is hosting gitea. This is done by querying the `/api/swagger` endpoint, as that is a small static page that contains “Gitea” even when the instance has been renamed.

[Gitea instances that are hosted in a subfolder](https://docs.gitea.io/en-us/reverse-proxies/#nginx-with-a-sub-path) are not supported.

### DIFF
--- a/src/lib/Hydra/Plugin/GiteaStatus.pm
+++ b/src/lib/Hydra/Plugin/GiteaStatus.pm
@@ -29,6 +29,53 @@ sub toGiteaState {
     }
 }
 
+sub url_from_jobsetevalinputs {
+    my ($eval) = @_;
+    my $giteastatusInput = $eval->jobsetevalinputs->find({ name => "gitea_status_repo" });
+    return undef unless defined $giteastatusInput && defined $giteastatusInput->value;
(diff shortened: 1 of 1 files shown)


Ma27:
Even though the UI prohibits it, it's perfectly possible to specify both a flake-ref and jobset inputs.
I changed this in https://github.com/NixOS/hydra/pull/922. With that patch you can do both (and I actually use it for precisely your use-case: building flakes in Hydra and pushing the status to gitea, see for instance https://hydra.ist.nicht-so.sexy/eval/795#tabs-inputs).

jiriks74:
I ran into this as well. Took me some time to get hydra running and for it to build my system but now I've hit a roadblock.

My systems are defined in a flake but I can't propagate the build status if the jobset is a flake and I don't know if it's possible to load it from a flake. Would be helpful to move this along.

PR #1234, opened 2022-07-08
Add a squiggly line to the Hydra link on hover

dasJ:
The effect is the same as the one on links in mail bodys on
https://lists.apache.org/

### DIFF
--- a/src/root/layout.tt
+++ b/src/root/layout.tt
@@ -93,7 +93,7 @@
       <footer class="navbar">
         <hr />
         <small>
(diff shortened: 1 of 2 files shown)


samueldr:
Why?

It's cute, but I really see no reason this should be done. Good thing it's only on that one link in the footer, I don't understand how the people at apache are legit using this in actual content.

dasJ:
We had a discussion about that a couple of months ago and I ended up dropping the idea because I couldn't manage to implement this properly.
This is basically a follow-up to that, now that I have a better implementation.

And yes, "It's cute" is pretty much the only reason for this. It does not add anything apart from me (and hopefully others) smiling sometimes when I/they see this silly feature.

PR #1237, opened 2022-07-12
Update to Nix 2.11.0 and Nixpkgs 22.05

edolstra:
There was some breakage due to the new auto-chroot feature in Nix 2.10, which doesn't handle correctly the case where `$NIX_STORE_DIR` etc. are set.


Ericson2314:
BTW what commit/PR is this new feature on the Nix side? Just want to stay up to date with these things for doc purposes etc.

cole-h:
I believe that would be https://github.com/NixOS/nix/pull/6714.

Ericson2314:
Thanks!

NickCao (commented):
Does this fix https://github.com/NixOS/nix/issues/6981

PR #1243, opened 2022-08-11
Run the JS code to make build trees collapsible at the right time

K900:
Turns out the code has been there all along, it just didn't play nice with lazy tab loading.

Disclaimer: my JS is bad and my Perl is even worse, but this _should_ work.


K900:
image

K900:
A few more changes: 
image

PR #1244, opened 2022-08-17
Drop the HipChat plugin

dasJ:
https://en.wikipedia.org/wiki/HipChat says:
> Following this, HipChat and Stride customers were migrated to the
> Slack group collaboration platform in a transition that was completed by
> February 2019.


ajs124:
Should also remove the documentation from `doc/manual/src/plugins/README.md`

PR #1250, opened 2022-09-29
hydra: fix localhost detection when protocol prefix are used

Mic92:
cc @dasJ


Ericson2314:
Note I don't think Hydra yet works with `ssh-ng://`. But the basic idea seems good.

Mic92:
> Note I don't think Hydra yet works with `ssh-ng://`. But the basic idea seems good.

This should not make a difference in this case because it won't use ssh-ng if it detects localhost.

PR #1256, opened 2022-10-26
exit with error if database connectivity lost

cransom:
There's currently no automatic recovery for disconnected databases in the evaluator. This means if the database is ever temporarily unavailable, hydra-evaluator will sit and spin with no work accomplished.

If this condition is caught, the daemon will exit and systemd will be responsible for resuming the service.

### DIFF
--- a/src/hydra-evaluator/hydra-evaluator.cc
+++ b/src/hydra-evaluator/hydra-evaluator.cc
@@ -366,6 +366,9 @@ struct Evaluator
                     printInfo("received jobset event");
                 }
 
+            } catch (pqxx::broken_connection & e) {
+                printError("Database connection broken: %s", e.what());
+                std::_Exit(1);
             } catch (std::exception & e) {
                 printError("exception in database monitor thread: %s", e.what());
                 sleep(30);
@@ -473,6 +476,9 @@ struct Evaluator
         while (true) {
             try {
                 loop();
+            } catch (pqxx::broken_connection & e) {
+                printError("Database connection broken: %s", e.what());
+                std::_Exit(1);
             } catch (std::exception & e) {
                 printError("exception in main loop: %s", e.what());
                 sleep(30);


cransom:
tested locally by running postgres and hydra-evaluator via foreman and then stopping postgres. 
```
1ogre:~/git/branches/hydra/master% foreman start hydra-evaluator
16:07:56 hydra-evaluator.1    | started with pid 417284
16:07:56 hydra-evaluator.1    | Connection to localhost (127.0.0.1) 63333 port [tcp/*] succeeded!
16:07:58 hydra-evaluator.1    | received jobset event
16:07:58 hydra-evaluator.1    | exception in database monitor thread: Lost connection to the database server.


16:08:28 hydra-evaluator.1    | exception in database monitor thread: Lost connection to the database server.
# here, the db was restarted but evaluator would never pick up on that.
^C16:08:53 system               | SIGINT received, starting shutdown
16:08:54 system               | sending SIGTERM to all processes
16:08:54 hydra-evaluator.1    | exited with code 1
2022-10-26 16:08:54-0400 [master|✚1…2]
1ogre:~/git/branches/hydra/master% foreman start hydra-evaluator
16:08:59 hydra-evaluator.1    | started with pid 420495
16:08:59 hydra-evaluator.1    | Connection to localhost (127.0.0.1) 63333 port [tcp/*] succeeded!
16:09:09 hydra-evaluator.1    | received jobset event
16:09:13 hydra-evaluator.1    | received jobset event
16:09:13 hydra-evaluator.1    | Database connection broken: Lost connection to the database server.
16:09:13 hydra-evaluator.1    | exited with code 1
16:09:13 system               | sending SIGTERM to all processes
```

src/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -366,6 +366,9 @@ struct Evaluator
                     printInfo("received jobset event");
                 }
 
+            } catch (pqxx::broken_connection & e) {
+                printError("Database connection broken: %s", e.what());
+                std::_Exit(1);
```
Mindavi, line 371: Isn't there `std::exit` or just `exit`?

cransom: Those calls exist, yes. I was following other bits in the code where exceptions happen and it is called this way. I don't have a good reason to deviate from that, but I'm happy to change if there is.

Mindavi: I don't know, the behavior is different apparently, so stick to the general pattern I'd say.

dasJ:
@grahamc can you take a look? I cannot judge the impact on the upstream Hydra

grahamc:
Thanks!

PR #1257, opened 2022-10-31
hydra-eval-jobs: don't use restrict-eval for Flakes

lheckemann:
Pure evaluation should cover all the uses for restrict-eval, and restrict-eval currently breaks accessing Git inputs on Flakes.

[1]: https://github.com/NixOS/nix/issues/7098


lheckemann:
Draft, because:
- I'm not 100% confident that there's nothing that works in pure-but-not-restricted eval but shouldn't work in the context of Hydra
- This results in Flake evals being able to access the local filesystem and arbitrary URLs if pure evaluation mode is disabled, which is a significant change from before

lheckemann:
I think it's OK to do this. People disabling pure evaluation presumably have a good reason to do so.

lheckemann:
In that case, I guess fixing https://github.com/NixOS/nix/issues/7098 in Nix is the right approach to solving this problem.

PR #1258, opened 2022-11-02
`renderInputDiff`: Increase git hash length 6 -> 8

nh2:
Nixpkgs has so many commits that length 6 is often ambiguous, making the use of the shown values with git difficult.

### DIFF
--- a/src/root/common.tt
+++ b/src/root/common.tt
@@ -374,7 +374,7 @@ BLOCK renderInputDiff; %]
               [% ELSIF bi1.uri == bi2.uri && bi1.revision != bi2.revision %]
                 [% IF bi1.type == "git" %]
                   <tr><td>
-                    <b>[% bi1.name %]</b></td><td><tt>[% INCLUDE renderDiffUri contents=(bi1.revision.substr(0, 6) _ ' to ' _ bi2.revision.substr(0, 6)) %]</tt>
+                    <b>[% bi1.name %]</b></td><td><tt>[% INCLUDE renderDiffUri contents=(bi1.revision.substr(0, 8) _ ' to ' _ bi2.revision.substr(0, 8)) %]</tt>
                   </td></tr>
                 [% ELSE %]
                   <tr><td>


SuperSandro2000:
In the long run it would probably be a good idea to use git to determine how long those hashes should be.

ajs124:
Yeah, iirc the database has a shortrev field for this, but the input plugins don't populate it.

edolstra:
How about 7 characters? That's what Nix's `shortRev` flake attribute uses.

dasJ:
I think the best way of doing it would be to use [whatever git uses](https://stackoverflow.com/a/21015031) which scales dynamically with repo size (but is never less than 7).
The scaling algorithm is already relevant for nixpkgs which uses 11 characters in my local checkout.

nh2:
There are currently

```
$ git rev-list --all | wc -l
426593
```

many commits in the repo and

```
$ echo $(( $(git rev-list --all | wc -l) - $(git rev-list --all | cut -c-6 | sort -u | wc -l) ))       
5250
```

commits that will be unclickable on Hydra. Respectively:

```
cut -c-6: 5250
cut -c-7:  327
cut -c-8:   24
cut -c-9:    2
cut -c-10:   0
```

> I think the best way of doing it would be to use [whatever git uses](https://stackoverflow.com/a/21015031) which scales dynamically with repo size

Wouldn't that require the Hydra render page to perform git operations, which might be slow?

---

Why is it important that the "changes" link use as-short-as-possible commits anyway?

Might as well show 12 chars and be done with it for a while: [`1b4722674c31 to b3a8f7ed267e`](https://hydra.nixos.org/api/scmdiff?type=git&branch=&uri=https%3A%2F%2Fgithub.com%2FNixOS%2Fnixpkgs.git&rev2=b3a8f7ed267e0a7ed100eb7d716c9137ff120fe3&rev1=1b4722674c315de0e191d0d79790b4eac51570a1)

vcunat:
I think the numbers are worse, because (most?) git operations won't proceed smoothly even in case of collision with non-commit objects (certainly trees).  While the number of revisions is under 0.5M, I have 3.7M objects (`git count-objects -v`).

Oh and I think we should merge (and deploy) this or something simple soon, and we can optionally think of some more ideal solution later.  I see no reason for this to remain open for months.

nh2:
> `cut -c-8:   24`
>
> Might as well show 12 chars and be done with it for a while

PR to bump the length to 12: https://github.com/NixOS/hydra/pull/1396

We shouldn't waste time on such papercuts.

PR #1265, opened 2022-12-07
Use nix with updated nixpkgs, and fix for removal of internal JSON functionality

lheckemann:
Draft, pending https://github.com/NixOS/nix/pull/7367 and referring to Nix's master branch or a release containing the changes

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -280,7 +280,7 @@
                 machine.wait_for_job("hydra-server")
                 machine.wait_for_job("hydra-evaluator")
                 machine.wait_for_job("hydra-queue-runner")
-                machine.wait_for_open_port("3000")
+                machine.wait_for_open_port(3000)
                 machine.succeed("curl --fail http://localhost:3000/")
               '';
           });
@@ -317,7 +317,7 @@
 
               # Wait until InfluxDB can receive web requests
               machine.wait_for_job("influxdb")
-              machine.wait_for_open_port("8086")
+              machine.wait_for_open_port(8086)
 
               # Create an InfluxDB database where hydra will write to
               machine.succeed(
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "A Nix-based continuous build system";
 
   inputs.nixpkgs.follows = "nix/nixpkgs";
-  inputs.nix.url = "github:NixOS/nix/2.11.0";
+  inputs.nix.url = "github:lheckemann/nix/nixpkgs-22.11";
```
edolstra, line 5: Can you point this to 2.12.0?

Ericson2314: I think it needs to be a branch with https://github.com/NixOS/nix/pull/7367, so `master` unless we backport that to the release branch?

Ericson2314: (IMO it would be nice to have a branch of hydra that always uses Nix master, along with at least one branch that uses a Nix release.)

Ma27 (changes requested):
The JSON changes break e.g. https://github.com/NixOS/nixos-org-configurations/blob/master/delft/prometheus/hydra-queue-runner-reexporter.py (which I'm also using on my instance) because `machines`/`machineTypes`/`jobsets` are always `null`.

Fixed in https://github.com/Ma27/hydra/commit/a7561ef7511c01c1a2db865a92caafec2f089891, feel free to cherry-pick this onto your branch :)

Ma27:
Implemented support for Nix 2.13 on my fork: https://github.com/Ma27/hydra/commit/de16336ca98d73ef0c629ce08baa903601c3526e

Note: this depends on https://github.com/NixOS/nix/pull/7705 since perl bindings for hydra-eval-jobset would break on runtime otherwise.

Issue #1267, opened 2022-12-29
Incorrect MIME type for .css and .js files breaks rendering of HTML reports


knedlsepp:
**Describe the bug**
HTML reports are no longer rendered correctly since https://github.com/NixOS/hydra/pull/1240 was merged, as the `Content-Type` of `.css` and `.js` files changed from `text/css` or `application/javascript` to  `text/plain`.

Firefox reports: 

```
The stylesheet was not loaded because its MIME type, "text/plain" is not "text/css"
```

**To Reproduce**

Visit https://hydra.nixos.org/build/203389362 and click [Manual](https://hydra.nixos.org/build/203389362/download/1/manual/). Or try visiting one of the following:

- https://nixos.org/nixops/manual
- https://nixos.org/hydra/manual

**Screenshots**
![Screenshot_20221229_215348](https://user-images.githubusercontent.com/3287933/210010369-7bb8be5a-3726-443a-8b71-1da35222f4da.png)

**Hydra Server:**

- OS and version: e.g. NixOS 22.11.20221229.ad35830
- Version of Hydra: hydra-unstable 2022-10-22
- Version of Nix Hydra is built against: nix-2.11.1
- Version of the Nix daemon: nix-2.11.1

peacememories:
It seems not only the `css` and `js` MIME-types are broken. When trying to view the page on an iPhone, it is instead downloaded, because the content type header looks like this:

```
content-type: cannot  open  `/nix/store/mp2x9s34n3m2lyd7hk9x15l1sry5qkda-hydra-manual-0.1.20221021.312cb42/share/doc/hydra/index.html'  (No such file or directory); charset=cannot open  `/nix/store/mp2x9s34n3m2lyd7hk9x15l1sry5qkda-hydra-manual-0.1.20221021.312cb42/share/doc/hydra/index.html'  (No such file or directory)
```

My desktop browser ignores this, but mobile Safari refuses to open the page and instead opens the download dialog.

PR #1268, opened 2022-12-29
Fix MIME types when serving .js and .css to fix rendering of HTML reports

knedlsepp:
To correctly render HTML reports we make sure to return the following MIME types instead of "text/plain"

- *.css: "text/css"
- *.js: "application/javascript"

Fixes: #1267

### DIFF
--- a/src/lib/Hydra/Controller/Build.pm
+++ b/src/lib/Hydra/Controller/Build.pm
@@ -238,9 +238,17 @@ sub serveFile {
                                                       "store", "cat", "--store", getStoreUri(), "$path"]) };
 
         # Detect MIME type.
-        state $magic = File::LibMagic->new(follow_symlinks => 1);
-        my $info = $magic->info_from_filename($path);
-        my $type = $info->{mime_with_encoding};
+        my $type = "text/plain";
+        if ($path =~ /.*\.(\S{1,})$/xms) {
+            my $ext = $1;
+            my $mimeTypes = MIME::Types->new(only_complete => 1);
+            my $t = $mimeTypes->mimeTypeOf($ext);
+            $type = ref $t ? $t->type : $t if $t;
+        } else {
+            state $magic = File::LibMagic->new(follow_symlinks => 1);
+            my $info = $magic->info_from_filename($path);
+            $type = $info->{mime_with_encoding};
+        }
         $c->response->content_type($type);
         $c->forward('Hydra::View::Plain');
     }


drupol:
While I agree that the proposed fix might be fixing the issue, I think it would be better to investigate why there is a permission issue on the server.

To reproduce it:

Anyone can reproduce the issue by doing:

```
❯ curl -I https://hydra.nixos.org/build/196107287/download/1/hydra/fonts/fonts.css
HTTP/2 200 
strict-transport-security: max-age=15552000
vary: Accept
content-type: cannot open `/nix/store/mp2x9s34n3m2lyd7hk9x15l1sry5qkda-hydra-manual-0.1.20221021.312cb42/share/doc/hydra/fonts/fonts.css' (No such file or directory); charset=cannot open `/nix/store/mp2x9s34n3m2lyd7hk9x15l1sry5qkda-hydra-manual-0.1.20221021.312cb42/share/doc/hydra/fonts/fonts.css' (No such file or directory)
last-modified: Sat, 22 Oct 2022 06:13:00 GMT
content-length: 3620
date: Mon, 10 Jul 2023 15:39:41 GMT
set-cookie: hydra_session=27212460deb4407adab3872f6adb1645739e1576; path=/; expires=Mon, 17-Jul-2023 15:39:41 GMT; HttpOnly; SameSite=Lax
server: Apache/2.4.54 (Unix) OpenSSL/1.1.1s

❯ curl -I https://hydra.nixos.org/build/196107287/download/1/hydra/clipboard.min.js
HTTP/2 200 
strict-transport-security: max-age=15552000
vary: Accept
content-type: cannot open `/nix/store/mp2x9s34n3m2lyd7hk9x15l1sry5qkda-hydra-manual-0.1.20221021.312cb42/share/doc/hydra/clipboard.min.js' (No such file or directory); charset=cannot open `/nix/store/mp2x9s34n3m2lyd7hk9x15l1sry5qkda-hydra-manual-0.1.20221021.312cb42/share/doc/hydra/clipboard.min.js' (No such file or directory)
last-modified: Sat, 22 Oct 2022 06:13:00 GMT
content-length: 10754
date: Mon, 10 Jul 2023 15:41:15 GMT
set-cookie: hydra_session=0052635d3699320dc95669657ee2b6ed0304b596; path=/; expires=Mon, 17-Jul-2023 15:41:15 GMT; HttpOnly; SameSite=Lax
server: Apache/2.4.54 (Unix) OpenSSL/1.1.1s
```

knedlsepp:
> I think it would be better to investigate why there is a permission issue on the server.

The issue was introduced in https://github.com/NixOS/hydra/pull/1240
When I debugged this locally the issue was that the content type changed to "text/plain" instead of "text/css". In this case we get a "(No such file or directory)". I'm guessing there are probably two issues with #1240 only one of which I identified. I don't know exactly how the official hydra.nixos.org is set up, but to my understanding the files are really stored on S3. Potentially the perl logic to determine the MIME type based on the content assumes that the file is available in the local nix store: https://github.com/NixOS/hydra/pull/1240/files#diff-752f4108f73ba8852c4532ebb985cdbd4601fbdb8971f746669c29157688ddf9R241-R242

It could be that this issue would still persist with the change I'm proposing here (which is a partial revert of #1240). It would at least be limited to files without an extension, but then again the question is if #1240 should be reverted altogether?

Ma27 (approved):
Encountered the same issue on my personal Hydra, this PR fixes it :+1:

PR #1277, opened 2023-02-16
systemd: hydra-queue-runner: wait for network-online

Mindavi:
This prevents eval errors when a machine is just started and the network isn't yet online. I'm running hydra on a laptop and the network takes a bit of time to come online (WLAN), so it's nice if the evaluator starts only when the network actually goes online.

Otherwise an error like this can happen on the first eval(s):

```
error fetching latest change from git repo at `https://github.com/nixos/nixpkgs.git':
fatal: unable to access 'https://github.com/nixos/nixpkgs.git/': Could not resolve host: github.com
```

See also https://github.com/NixOS/nixpkgs/pull/216451

### DIFF
--- a/hydra-module.nix
+++ b/hydra-module.nix
@@ -340,7 +340,7 @@ in
     systemd.services.hydra-queue-runner =
       { wantedBy = [ "multi-user.target" ];
         requires = [ "hydra-init.service" ];
-        after = [ "hydra-init.service" "network.target" ];
+        after = [ "hydra-init.service" "network-online.target" ];
         path = [ cfg.package pkgs.nettools pkgs.openssh pkgs.bzip2 config.nix.package ];
         restartTriggers = [ hydraConf ];
         environment = env // {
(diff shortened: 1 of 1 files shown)


hydra-module.nix
```diff
@@ -340,7 +340,7 @@ in
     systemd.services.hydra-queue-runner =
       { wantedBy = [ "multi-user.target" ];
         requires = [ "hydra-init.service" ];
-        after = [ "hydra-init.service" "network.target" ];
+        after = [ "hydra-init.service" "network-online.target" ];
```
SuperSandro2000, line 343:
```suggestion
        after = [ "hydra-init.service" "network.target" "network-online.target" ];
```

lheckemann (approved):

SuperSandro2000:
Also this might not fully fix this issue as network-online means the interface is online, not that actual internet connectivity is available but it will probably improve things.

Mindavi:
For me it seems to work but yeah, hopefully it at least improves.

PR #1279, opened 2023-03-06
Drop unused IndexBuildOutputsOnPath index

cole-h:
Also it's larger than the actual table it's indexing lol.

    -[ RECORD 30 ]----------+-----------------------------------------
    table_name              | buildoutputs
    index_name              | indexbuildoutputsonpath
    index_scans_count       | 0
    index_size              | 31 GB
    table_reads_index_count | 2128699937
    table_reads_seq_count   | 0
    table_reads_count       | 2128699937
    table_writes_count      | 22442976
    table_size              | 28 GB

### DIFF
--- a/src/sql/upgrade-83.sql
+++ b/src/sql/upgrade-83.sql
@@ -1,0 +1,3 @@
+-- This index was introduced in a migration but was never recorded in
+-- hydra.sql (the source of truth), which is why `if exists` is required.
+drop index if exists IndexBuildOutputsOnPath;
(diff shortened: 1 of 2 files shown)


src/sql/upgrade-83.sql
```diff
@@ -0,0 +1 @@
+drop index if exists IndexBuildOutputsOnPath;
```
grahamc, line 1: Note that this was introduced in a migration and doesn't seem to have been properly recorded in hydra.sql.

cole-h: I added a comment to that effect in this migration.

PR #1281, opened 2023-03-14
Use new Google for Web signin

rbvermaa:
Use new Google for Web signin, the old way will be deprecated Mar 31st 2023


grahamc:
thanks!

PR #1291, opened 2023-06-23
Update Nix and Nixpkgs

edolstra:


Ma27 (approved):
Heh, I have a change lying around that looks super-similar to that one, I just didn't get to file a PR yet ;-)

Anyways, this is close enough to what I've hacked and my personal Hydra is using that for a few days now, hence :+1: on the changes.

PR #1294, opened 2023-07-20
Document redirects in Hydra API

arianvp:
This documents useful redirects that Hydra exposes


dasJ:
With the other merges this has conflicts now :(

PR #1298, opened 2023-08-20
Implement generic OIDC-based authentication 

lheckemann:


src/lib/Hydra/Controller/User.pm
```diff
@@ -226,6 +272,29 @@ sub github_redirect :Path('/github-redirect') Args(0) {
     $c->res->redirect("https://github.com/login/oauth/authorize?client_id=$client_id&scope=user:email");
 }
 
+sub oidc_redirect :Path('/oidc-redirect') Args(0) {
+
+    my ($self, $c) = @_;
+
+    my $after = "/" . $c->req->params->{after};
+
+    $c->res->cookies->{'after_oidc'} = {
```
dasJ, line 281: I saw this in GitHub as well and I really don't know why we would do this. We already have a session so why not do:
```
$c->session->{oidc_after} = $after;
```
You already do this for the state (which is what GitHub is not doing right now afaik :eyes:)

src/lib/Hydra/Controller/User.pm
```diff
@@ -226,6 +272,29 @@ sub github_redirect :Path('/github-redirect') Args(0) {
...
+        name => 'after_oidc',
+        value => $after,
+    };
+
+    my $oidc_client = get_oidc_client($c);
+    my $state = UUID4::Tiny::create_uuid_string();
+    $c->session->{oidc_state} = $state;
+    my $redirect_url = $oidc_client->uri_to_redirect(
+        redirect_uri => $c->uri_for('/oidc-login'),
+        scope        => q{openid},
```
dasJ, line 291: Would be nice to have this configurable

src/lib/Hydra/Controller/User.pm
```diff
@@ -92,14 +104,16 @@ sub doLDAPLogin {
...
 
 sub doEmailLogin {
-    my ($self, $c, $type, $email, $fullName) = @_;
+    my ($self, $c, $type, $email, $fullName, $username) = @_;
 
     die "No email address provided.\n" unless defined $email;
 
     # Be paranoid about the email address format, since we do use it
     # in URLs.
     die "Illegal email address.\n" unless $email =~ /^[a-zA-Z0-9\.\-\_]+@[a-zA-Z0-9\.\-\_]+$/;
```
dasJ, line 113: Unrelated: this is the absolute worst because it excludes plus-emails

src/lib/Hydra/Controller/User.pm
```diff
@@ -143,6 +157,38 @@ sub doEmailLogin {
...
+        $c->config->{oidc_userinfo_uri},
+        Authorization => sprintf(q{Bearer %s}, $token_response->access_token)
+    );
+
+    my $claims = decode_json($res->decoded_content) or error($c, q{Could not decode claims.}, 401);
+
+    error($c, "Email address must be verified.", 401) unless $claims->{email_verified};
+
+    doEmailLogin($self, $c, "oidc", $claims->{email}, $claims->{name} // undef, sprintf("oidc:$claims->{sub}"));
+
```
dasJ, line 188: I think it'd be nice to have some role mapping as well because otherwise everyone has no permissions. We use:
```patch
+    my $role_mapping = $c->config->{oidc}->{role_mapping} or ();
+    # Give out roles
+    $c->user->userroles->delete;
+    push(@{$user_data->{groups}}, ("default"));
+    foreach my $group (@{$user_data->{groups}}) {
+        if (defined($role_mapping->{$group})) {
+            my $roles = $role_mapping->{$group};
+            if (ref($roles) eq 'ARRAY') {
+                for my $mapped_role (@{$roles}) {
+                    $c->user->userroles->create({ role => $mapped_role });
+                }
+            } else {
+                $c->user->userroles->create({ role => $roles });
+            }
+        }
+    }
```

src/lib/Hydra/Controller/User.pm
```diff
@@ -143,6 +157,38 @@ sub doEmailLogin {
 }
 
 
+sub oidc_login :Path('/oidc-login') Args(0) {
+    my ($self, $c) = @_;
+
+    error($c, "Logging in via OIDC is not enabled.", 404) unless $c->config->{enable_oidc_login};
```
dasJ, line 163: Why not scope these? Something like `$c>config->{oidc}->{enable}`? Or even better `$c->config->{oidc}->{client_id}` which can be undef or defined and enable/disable oidc auth that way

src/lib/Hydra/Controller/User.pm
```diff
@@ -143,6 +157,38 @@ sub doEmailLogin {
...
+        Authorization => sprintf(q{Bearer %s}, $token_response->access_token)
+    );
+
+    my $claims = decode_json($res->decoded_content) or error($c, q{Could not decode claims.}, 401);
+
+    error($c, "Email address must be verified.", 401) unless $claims->{email_verified};
+
+    doEmailLogin($self, $c, "oidc", $claims->{email}, $claims->{name} // undef, sprintf("oidc:$claims->{sub}"));
+
+    $c->res->redirect($c->uri_for($c->res->cookies->{'after_oidc'}));
```
dasJ, line 189: Same issue as with the GitHub auth: Doesn't work. If you use the cookie (and not the session which I recommended in another comment), you need `$c->req->cookies`, not `res`

src/root/topbar.tt
```diff
@@ -143,6 +143,10 @@
         <a class="dropdown-item" href="/github-redirect?after=[% c.req.path %]">Sign in with GitHub</a>
         <div class="dropdown-divider"></div>
       [% END %]
+      [% IF c.config.enable_oidc_login %]
+        <a class="dropdown-item" href="/oidc-redirect?after=[% c.req.path %]">Sign in with OIDC</a>
```
dasJ, line 147: Would also be nice to be able to change the name of the provider so it can be consistent in an organization.

dasJ (changes requested):
We actually have a patch for our idam that is less generic than yours (which is why it's not upstreamed). The comments probably reflect that ;)
I do like your approach which using proper libraries a lot! Also, how testable is this?

src/lib/Hydra/Controller/User.pm
lheckemann: Yes, this is just taken directly from the other auth methods. We did discuss that it should also probably be a single session variable regardless of the auth method. I'd prefer to fix this later though.

src/lib/Hydra/Controller/User.pm
lheckemann: I'm not sure that's desirable: we care about the `name`, `email` and `sub` fields which are standard in openid, and I would very much like to avoid going further than that and requiring mapping logic.

src/lib/Hydra/Controller/User.pm
lheckemann: Yes.

src/lib/Hydra/Controller/User.pm
lheckemann: We discussed this at length (this was what took longest by far), and came to the conclusion that this is a real can of worms, because OIDC is about _authentication_ and _identity_ but not _authorisation_. For now, our recommended approach is to manage permissions manually.

src/lib/Hydra/Controller/User.pm
lheckemann: Ergh, you're right. I think it _did_ work for some reason though.

src/lib/Hydra/Controller/User.pm
```diff
@@ -143,6 +157,38 @@ sub doEmailLogin {
...
+    my $code = $c->request->param("code");
+    my $token_response = $oidc_client->get_access_token(
+        code => $code,
+        redirect_uri => $c->uri_for('/oidc-login'),
+    ) or error($c, $oidc_client->errstr);
+
+    $c->session->{access_token} = $token_response->access_token;
+    $c->session->{expires_at} = time() + $token_response->expires_in;
+    $c->session->{refresh_token} = $token_response->refresh_token;
+
```
arianvp, line 177: We really ought to check the `id_token` and its signature here and make sure that the `id_token` is bound to the audience (client_id)..  And then save the `sub` and compare it to the sub received in userinfo.

Otherwise we are vulnerable against token substitution attacks as the access token is not audience-bound.

> NOTE: Due to the possibility of token substitution attacks, the UserInfo Response is not guaranteed to be about the End-User identified by the sub (subject) element of the ID Token. The sub Claim in the UserInfo Response MUST be verified to exactly match the sub Claim in the ID Token; if they do not match, the UserInfo Response values MUST NOT be used. 

https://openid.net/specs/openid-connect-basic-1_0.html#UserInfo

We're not reusing the `access_token` long-term but still. we should use the `id_token` as the source of truth about the user identity. Not the `userinfo` response

stepbrobd:
apparently i don't know how oidc role mapping work :-0
@ners resolved this for me at nixcon (nice)

example with kanidm:

```nix
services.hydra.extraConfig = ''
        enable_hydra_login = 0
        enable_oidc_login = 1
        oidc_client_id = "hydra"
        oidc_scope = "openid email profile groups"
        oidc_auth_uri = "https://kanidm.example.com/ui/oauth2"
        oidc_token_uri = "https://kanidm.example.com/oauth2/token"
        oidc_userinfo_uri = "https://kanidm.example.com/oauth2/openid/hydra/userinfo"
        include ${config.sops.secrets.hydra.path} # oidc_client_secret = ...

        <oidc_role_mapping>
          <my hydra admin role>@<kanidm tenant domain> = admin
          <my hydra admin role>@<kanidm tenant domain> = bump-to-front
          <my hydra user role>@<kanidm tenant domain> = cancel-build
          <my hydra user role>@<kanidm tenant domain> = eval-jobset
          <my hydra user role>@<kanidm tenant domain> = create-projects
          <my hydra user role>@<kanidm tenant domain> = restart-jobs
        </oidc_role_mapping>
'';
```

the only drawback of the current implementation is that we are missing PKCE support, otherwise patch lgtm (i can add docs if needed)

stepbrobd (approved):

PR #1300, opened 2023-08-30
Document all_builds (/eval/{eval-id}/builds) endpoint

sternenseemann:
The endpoint is implemented here:

https://github.com/NixOS/hydra/blob/00d30874da759eb0f44f446415b2469920ff41b5/src/lib/Hydra/Controller/JobsetEval.pm#L188-L196

### DIFF
--- a/hydra-api.yaml
+++ b/hydra-api.yaml
@@ -533,13 +533,13 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
 
-  /eval/{build-id}:
+  /eval/{eval-id}:
     get:
-      summary: Retrieves evaluations identified by build id
+      summary: Retrieves evaluations identified by eval id
       parameters:
-        - name: build-id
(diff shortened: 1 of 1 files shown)


Mindavi (approved):
Tried this out on my local instance and on hydra.nixos.org, and indeed this endpoint exists! Thanks for documenting it, it's a very useful endpoint and prevents having to scrape html or do hundreds of thousands of requests (if you want to check all builds in a nixpkgs eval...)

sternenseemann:
I think it has a lot of potential, especially for e.g. nix-review-tools which [currently scrapes the jobset overview HTML with `?full=1` passed](https://github.com/nix-community/nix-review-tools/blob/master/lib/hydra/eval.rb). However, the endpoint is very slow at the moment which can hopefully be helped somehow, though.

Ericson2314 (approved):

PR #1303, opened 2023-10-08
align `system` retrieval with `nix-build`

DavHau:
Read the system from the instantiated derivation instead of getting it from the eval time drv attrs.

This is better in terms of correctness as it derives the system through a derivations string context, instead of relying on builder implementations to set the `system` attribute correctly.

This derivation for example did not build with hydra before while it builds fine with the nix repl:

```
:b { name="hello"; type="derivation"; drvPath = hello.drvPath; }
```

### DIFF
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -177,15 +177,14 @@ static void worker(
 
                 DrvInfo::Outputs outputs = drv->queryOutputs();
 
-                if (drv->querySystem() == "unknown")
-                    throw EvalError("derivation must have a 'system' attribute");
+                auto drvContent = state.store->readDerivation(drv->requireDrvPath());
 
                 auto drvPath = state.store->printStorePath(drv->requireDrvPath());
 
                 nlohmann::json job;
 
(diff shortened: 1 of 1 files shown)


roberth (approved):
Besides removing a risk, this simplifies the language-level interface.

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -177,15 +177,14 @@ static void worker(
 
                 DrvInfo::Outputs outputs = drv->queryOutputs();
 
-                if (drv->querySystem() == "unknown")
-                    throw EvalError("derivation must have a 'system' attribute");
+                auto drvContent = state.store->readDerivation(drv->requireDrvPath());
```
grahamc, line 180: I'm concerned about how much IO this is going to introduce for a project like Nixpkgs.

roberth: I would expect them to still be in the page cache, but we could cache parsed derivations on the Nix side to save some context switches and a little bit of computation.

Ericson2314: Also do we even need to return the system here? I have some doubts that we in fact do.

PR #1308, opened 2023-11-21
support nix 2.18

chayleaf:


chayleaf:
it seems to work fine in my deployment

PR #1312, opened 2023-11-30
Cleanup deps

Ericson2314:
- `nativeBuildInputs` vs `buildInputs`

- narrow down `with`s for clarity

- use `autoreconfHook` not `bootstrap` script

These sorts of changes have also been done in the Nix repo.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -61,10 +61,11 @@
 
         };
 
-        hydra = with final; let
-          perlDeps = buildEnv {
+        hydra = let
+          inherit (final) lib stdenv;
+          perlDeps = final.buildEnv {
             name = "hydra-perl-deps";
-            paths = with perlPackages; lib.closePropagation
+            paths = with final.perlPackages; lib.closePropagation
               [
                 AuthenSASL
                 CatalystActionREST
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -141,15 +142,20 @@
 
           src = self;
 
-          buildInputs =
-            [
+          nativeBuildInputs =
+            with final.buildPackages; [
```
SuperSandro2000, line 146:
```suggestion
          nativeBuildInputs = with final.buildPackages; [
```

flake.nix
```diff
@@ -141,15 +142,20 @@
...
               libtool
-              unzip
               nukeReferences
               pkg-config
+              mdbook
+            ];
+
+          buildInputs =
+            with final; [
+              unzip
```
SuperSandro2000, line 158:
```suggestion
          buildInputs = with final; [
            unzip
```

flake.nix
Ericson2314: I don't like the existing style either, but if I did that I would be better off moving way more things around?

SuperSandro2000: hmmm but that would also be a much bigger change and I didn't see the other lines in the diff 😉

PR #1314, opened 2023-11-30
Upgrade to Nix 2.19

Ericson2314:
Depends on https://github.com/NixOS/hydra/pull/1308, which we should definitely merge first (and not skip a Nix release).

### DIFF
--- a/src/hydra-queue-runner/nar-extractor.cc
+++ b/src/hydra-queue-runner/nar-extractor.cc
@@ -24,13 +24,13 @@ struct Extractor : ParseSink
 
     void createDirectory(const Path & path) override
     {
-        members.insert_or_assign(prefix + path, NarMemberData { .type = FSAccessor::Type::tDirectory });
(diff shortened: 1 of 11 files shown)


src/hydra-queue-runner/nar-extractor.cc
```diff
@@ -66,8 +66,14 @@ struct Extractor : ParseSink
...
     {
-        members.insert_or_assign(prefix + path, NarMemberData { .type = FSAccessor::Type::tSymlink });
+        members.insert_or_assign(prefix + path, NarMemberData { .type = SourceAccessor::Type::tSymlink });
     }
+
+    void isExecutable() override
+    { }
+
+    void closeRegularFile() override
+    { }
```
Ericson2314, line 76: These when from being such empty functions in the interface to `= 0` virtual pure ones, so putting the old do-nothing implementations here.

PR #1316, opened 2023-12-04
Prepare for CA derivation support with lower impact changes

Ericson2314:
This is just C++ changes without any Perl / Frontend / SQL Schema changes.

The idea is that it should be possible to redeploy Hydra with these chnages with (a) no schema migration and also (b) no regressions. We should be able to much more safely deploy these to a staging server and then production `hydra.nixos.org`.

Extracted from #875

### DIFF
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -243,11 +243,16 @@ static void worker(
                 }
 
                 nlohmann::json out;
-                for (auto & [outputName, optOutputPath] : outputs)
-                    if (optOutputPath)
+                for (auto & [outputName, optOutputPath] : outputs) {
+                    if (optOutputPath) {
                         out[outputName] = state.store->printStorePath(*optOutputPath);
-                    else
+                    } else {
+                        // See the `queryOutputs` call above; we should
+                        // not encounter missing output paths otherwise.
+                        assert(experimentalFeatureSettings.isEnabled(Xp::CaDerivations));
                         out[outputName] = "";
+                    }
+                }
                 job["outputs"] = std::move(out);
                 reply["job"] = std::move(job);
             }
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -178,7 +178,11 @@ static void worker(
 
             if (auto drv = getDerivation(state, *v, false)) {
 
-                DrvInfo::Outputs outputs = drv->queryOutputs();
+                // CA derivations do not have static output paths, so we
+                // have to defensively not query output paths in case we
+                // encounter one.
+                DrvInfo::Outputs outputs = drv->queryOutputs(
+                    !experimentalFeatureSettings.isEnabled(Xp::CaDerivations));
 
                 if (drv->querySystem() == "unknown")
                     throw EvalError("derivation must have a 'system' attribute");
@@ -239,12 +243,12 @@ static void worker(
                 }
 
                 nlohmann::json out;
-                for (auto & j : outputs)
-                    // FIXME: handle CA/impure builds.
-                    if (j.second)
-                        out[j.first] = state.store->printStorePath(*j.second);
+                for (auto & [outputName, optOutputPath] : outputs)
+                    if (optOutputPath)
+                        out[outputName] = state.store->printStorePath(*optOutputPath);
+                    else
+                        out[outputName] = "";
                 job["outputs"] = std::move(out);
-
                 reply["job"] = std::move(job);
             }
 
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -250,6 +250,10 @@ static void worker(
                         // See the `queryOutputs` call above; we should
                         // not encounter missing output paths otherwise.
                         assert(experimentalFeatureSettings.isEnabled(Xp::CaDerivations));
+                        // TODO it would be better to set `null` than an
+                        // empty string here, to force the consumer of
+                        // this JSON to more explicitly handle this
+                        // case.
                         out[outputName] = "";
                     }
                 }
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -113,7 +113,7 @@ static void copyClosureTo(
     Machine::Connection & conn,
     Store & destStore,
     const StorePathSet & paths,
-    bool useSubstitutes = false)
+    SubstituteFlag useSubstitutes = NoSubstitute)
 {
     StorePathSet closure;
     destStore.computeFSClosure(paths, closure);
@@ -290,7 +290,7 @@ static BasicDerivation sendInputs(
             destStore.computeFSClosure(basicDrv.inputSrcs, closure);
             copyPaths(destStore, localStore, closure, NoRepair, NoCheckSigs, NoSubstitute);
         } else {
-            copyClosureTo(conn, destStore, basicDrv.inputSrcs, true);
+            copyClosureTo(conn, destStore, basicDrv.inputSrcs, Substitute);
         }
 
         auto now2 = std::chrono::steady_clock::now();
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -310,9 +310,6 @@ static BuildResult performBuild(
     counter & nrStepsBuilding
 )
 {
-
-    BuildResult result;
-
     conn.to << ServeProto::Command::BuildDerivation << localStore.printStorePath(drvPath);
     writeDerivation(conn.to, localStore, drv);
     conn.to << options.maxSilentTime << options.buildTimeout;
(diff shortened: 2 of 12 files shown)


src/hydra-queue-runner/build-remote.cc
```diff
@@ -182,6 +182,41 @@ static StorePaths reverseTopoSortPaths(const std::map<StorePath, ValidPathInfo>
...
+ * to the builder.
+ *
+ * If we can afford it, resolve it, so that the newly generated derivation still
+ * has some sensible output paths.
+ */
+BasicDerivation inlineInputDerivations(Store & store, Derivation & drv, const StorePath & drvPath)
+{
+    BasicDerivation ret;
+    auto outputHashes = staticOutputHashes(store, drv);
+    if (!drv.type().hasKnownOutputPaths()) {
```
Ericson2314, line 195: This condition is always `false` without ca derivations.

The else case is the same as the code which the single caller of this function replaces.

src/hydra-queue-runner/build-remote.cc
```diff
@@ -323,8 +347,33 @@ static BuildResult performBuild(
...
+        for (auto && [output, realisation] : builtOutputs)
+            result.builtOutputs.insert_or_assign(
+                std::move(output.outputName),
+                std::move(realisation));
+    } else {
+        // If the remote is too old to handle CA derivations, we can’t get this
+        // far anyways
+        assert(drv.type().hasKnownOutputPaths());
+        DerivationOutputsAndOptPaths drvOutputs = drv.outputsAndOptPaths(localStore);
+        auto outputHashes = staticOutputHashes(localStore, drv);
```
Ericson2314, line 365: These output hashes are not really used in the input-addressing case, but they may be expensive to calculate because the underlying `hashDerivationModulo` loads the derivation closure (but cached).

src/hydra-queue-runner/build-remote.cc
```diff
@@ -624,6 +673,23 @@ void State::buildRemote(ref<Store> destStore,
             result.overhead += std::chrono::duration_cast<std::chrono::milliseconds>(now2 - now1).count();
         }
 
+        /* Register the outputs of the newly built drv */
+        if (experimentalFeatureSettings.isEnabled(Xp::CaDerivations)) {
```
Ericson2314, line 677: Skipped in non-CA case

src/hydra-queue-runner/build-remote.cc
```diff
@@ -395,7 +445,8 @@ static void copyPathFromRemote(
           extractNarData(tee, localStore.printStorePath(info.path), narMembers);
       });
 
-      destStore.addToStore(info, *source2, NoRepair, NoCheckSigs);
+      store->addToStore(info, *source2, NoRepair, NoCheckSigs);
```
Ericson2314, line 448: re-adding to local store should be a no-op when it already has it: It is able skip bey looking at the path in `info`.

src/hydra-queue-runner/build-result.cc
```diff
@@ -107,13 +107,12 @@ BuildOutput getBuildOutput(
...
-        for (auto & [name, output] : drv.outputs) {
+        for (auto & [name, output] : derivationOutputs) {
             BuildProduct product;
-            auto outPath = output.path(*store, drv.name, name);
-            product.path = store->printStorePath(*outPath);
+            product.path = store->printStorePath(output);
             product.type = "nix-build";
             product.subtype = name == "out" ? "" : name;
-            product.name = outPath->name();
+            product.name = output.name();
```
Ericson2314, line 115: This whole file is the same logic as before. In the CA case more work has to be done to get the paths so that was moved to the caller. In the input-addressed case it is easy and moving the tiny bit of work back and forth should not be noticeable.

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -352,11 +352,23 @@ void State::finishBuildStep(pqxx::work & txn, const RemoteResult & result,
...
+        // Update the corresponding `BuildStepOutputs` row to add the output path
+        auto res = txn.exec_params1("select drvPath from BuildSteps where build = $1 and stepnr = $2", buildId, stepNr);
+        assert(res.size());
+        StorePath drvPath = localStore->parseStorePath(res[0].as<std::string>());
+        // If we've finished building, all the paths should be known
+        for (auto & [name, output] : localStore->queryDerivationOutputMap(drvPath))
+            txn.exec_params0
+                ("update BuildStepOutputs set path = $4 where build = $1 and stepnr = $2 and name = $3",
+                  buildId, stepNr, name, localStore->printStorePath(output));
+    }
```
Ericson2314, line 366: This is an idemopotent / no changed output extra step in the input-adressed case.

src/hydra-queue-runner/hydra-queue-runner.cc
```diff
@@ -457,6 +469,15 @@ void State::markSucceededBuild(pqxx::work & txn, Build::ptr build,
...
 
+    for (auto & [outputName, outputPath] : res.outputs) {
+        txn.exec_params0
+            ("update BuildOutputs set path = $3 where build = $1 and name = $2",
+             build->id,
+             outputName,
+             localStore->printStorePath(outputPath)
+            );
+    }
+
```
Ericson2314, line 480: This is an idempotent / no changed output extra step in the input-addressed case.

src/hydra-queue-runner/queue-monitor.cc
```diff
@@ -192,15 +192,14 @@ bool State::getQueuedBuilds(Connection & conn,
...
-                            auto res = txn.exec_params
-                                ("select max(s.build) from BuildSteps s join BuildStepOutputs o on s.build = o.build where path = $1 and startTime != 0 and stopTime != 0 and status = 1",
-                                 localStore->printStorePath(*i.second.second));
-                            if (!res[0][0].is_null()) {
-                                propagatedFrom = res[0][0].as<BuildID>();
-                                break;
-                            }
+                    for (auto & i : localStore->queryPartialDerivationOutputMap(ex.step->drvPath)) {
+                          auto res = txn.exec_params
+                              ("select max(s.build) from BuildSteps s join BuildStepOutputs o on s.build = o.build where drvPath = $1 and name = $2 and startTime != 0 and stopTime != 0 and status = 1",
```
Ericson2314, line 197: The switch from querying via `path` to querying via `(drvPath, name)` should be fine.

src/hydra-queue-runner/queue-monitor.cc
```diff
@@ -481,71 +478,84 @@ Step::ptr State::createStep(ref<Store> destStore,
...
+            if (!destStore->isValidPath(*maybeOutputPath.second)) {
+                valid = false;
+                missing.insert({{outputHash, outputName}, maybeOutputPath.second});
+            }
+        } else {
+            experimentalFeatureSettings.require(Xp::CaDerivations);
+            if (!destStore->queryRealisation(DrvOutput{outputHash, outputName})) {
+                valid = false;
+                missing.insert({{outputHash, outputName}, std::nullopt});
+            }
```
Ericson2314, line 496: This is dead code with ca-derivations disabled

src/hydra-queue-runner/queue-monitor.cc
```diff
@@ -481,71 +478,84 @@ Step::ptr State::createStep(ref<Store> destStore,
...
        substitutes. */
     if (!missing.empty()) {
 
         size_t avail = 0;
-        for (auto & i : missing) {
-            auto path = i.second.path(*localStore, step->drv->name, i.first);
-            if (/* localStore != destStore && */ localStore->isValidPath(*path))
+        for (auto & [i, maybePath] : missing) {
+            if ((maybePath && localStore->isValidPath(*maybePath)))
+                avail++;
```
Ericson2314, line 507: Same

src/hydra-queue-runner/queue-monitor.cc
```diff
@@ -481,71 +478,84 @@ Step::ptr State::createStep(ref<Store> destStore,
...
+            else if (experimentalFeatureSettings.isEnabled(Xp::CaDerivations) && localStore->queryRealisation(i)) {
+                maybePath = localStore->queryRealisation(i)->outPath;
                 avail++;
```
Ericson2314, line 510: Dead code without CA derivations

src/hydra-queue-runner/queue-monitor.cc
```diff
@@ -481,71 +478,84 @@ Step::ptr State::createStep(ref<Store> destStore,
...
 
         if (missing.size() == avail) {
             valid = true;
-            for (auto & i : missing) {
-                auto path = i.second.path(*localStore, step->drv->name, i.first);
-
-                try {
-                    time_t startTime = time(0);
+            for (auto & [i, path] : missing) {
+                if (path) {
```
Ericson2314, line 522: Next chunk of diff goes away with whitepaper enabled in the non-CA case we always have a `path` so extra `if` doesn't make a difference

thufschmitt:
@Ericson2314 : Thanks a lot for that. That looks like a good plan.

I'm a bit stretched out, but I'll try to block enough time to review this and follow-ups (I'm the one who committed this in the first place after all :p )

src/hydra-eval-jobs/hydra-eval-jobs.cc
```diff
@@ -239,12 +243,12 @@ static void worker(
...
                 nlohmann::json out;
-                for (auto & j : outputs)
-                    // FIXME: handle CA/impure builds.
-                    if (j.second)
-                        out[j.first] = state.store->printStorePath(*j.second);
+                for (auto & [outputName, optOutputPath] : outputs)
+                    if (optOutputPath)
+                        out[outputName] = state.store->printStorePath(*optOutputPath);
+                    else
+                        out[outputName] = "";
```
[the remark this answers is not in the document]
Ericson2314: Oh yes, good call!

Ericson2314:
@thufschmitt Thank you! Yeah I'd like to think this plan has enough fail-safes that is is OK if you don't review it, but you definitely have the most context having figured out what needs to happen.

I think I'll add some review comments where I'd like to write a code comment explaining *why* something, and if you could just provide the missing tidbit (if you remember it after all these years!) I think that would be plenty.

src/hydra-queue-runner/build-remote.cc
Ericson2314: @thufschmitt do you know why we need to add in both stores now? Does it have to do with realizations?

src/hydra-eval-jobs/hydra-eval-jobs.cc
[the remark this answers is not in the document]
Ericson2314: Done

src/hydra-queue-runner/build-remote.cc
thufschmitt: I think it's an easy (and terribly inefficient :grimacing:) way of making sure that `localStore.queryRealisations` will succeed (which we IIRC we need later to get back some metadata about the path to put it in the db).

To be honest, we shouldn't do that but instead carry the needed metadata in memory until the point where we need it (but that can come later once we're confident that this is at least correct)

Ericson2314: OK thanks! I think that was my only question!

Ericson2314:
After discussing with @tomberek  https://github.com/NixOS/nixos-org-configurations/pull/316/files this PR is now deployed to https://hydra.ngi0.nixos.org/. I'll make a CA derivations tracking issue for these things.

src/hydra-queue-runner/build-remote.cc
Ericson2314: Actually, since I want this PR to reach prod before the rest, I think it is better to kick this change out of here and back into the main PR.

Skipping this will be no problem for input-adressed derivations, and the point of this PR is, after all, to bridge the gap without causing regressions, not add new functionality.

Ericson2314: I think it has to be the ideal one to not mess up prod hydra, so I'll aim for that.

Ericson2314:
I think I want to fix the realisations issue first, now that I think I know how to.

src/hydra-queue-runner/build-remote.cc
Ericson2314: Branch is gone!

src/hydra-queue-runner/queue-monitor.cc
Ericson2314: No it's not fine, those columns don't exist without the migration!

Ericson2314: Reverted

Ericson2314:
As discussed with @delroth, now that `hydra.nixos.org` is caught up with Nix 2.19, the next step is merging this!

PR #1318, opened 2023-12-07
Use factored-out `BuildResult` serializer

Ericson2314:
For the record, here is the Nix 2.19 version:
https://github.com/NixOS/nix/blob/2.19-maintenance/src/libstore/serve-protocol.cc, which is what we would initially use.

It is a more complete version of what Hydra has today except for one thing: it always unconditionally sets the start/stop times.

I think that is correct at the other end seems to unconditionally measure them, but just to be extra careful, I reproduced the old behavior of falling back on Hydra's own measurements if `startTime` is 0.

The only difference is that the fallback `stopTime` is now measured from after the entire `BuildResult` is transferred over the wire, but I think that should be negligible if it is measurable at all. (And remember, this is fallback case I already suspect is dead code.)

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -286,9 +286,6 @@ static BuildResult performBuild(
     counter & nrStepsBuilding
 )
 {
-
-    BuildResult result;
-
     conn.to << ServeProto::Command::BuildDerivation << localStore.printStorePath(drvPath);
     writeDerivation(conn.to, localStore, drv);
     conn.to << options.maxSilentTime << options.buildTimeout;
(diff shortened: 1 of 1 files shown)


src/hydra-queue-runner/build-remote.cc
```diff
@@ -286,9 +286,6 @@ static BuildResult performBuild(
     counter & nrStepsBuilding
 )
 {
-
-    BuildResult result;
```
Ericson2314, line 290: Just moving this down, it is for the response so it doesn't need to be in scope for the request.

src/hydra-queue-runner/build-remote.cc
```diff
@@ -301,30 +298,25 @@ static BuildResult performBuild(
...
-    result.stopTime = time(0);
-
-
-    result.errorMsg = readString(conn.from);
-    if (GET_PROTOCOL_MINOR(conn.remoteVersion) >= 3) {
-        result.timesBuilt = readInt(conn.from);
-        result.isNonDeterministic = readInt(conn.from);
-        auto start = readInt(conn.from);
-        auto stop = readInt(conn.from);
-        if (start && start) {
```
Ericson2314, line 319: I think this was supposed to be `start && stop` :)

src/hydra-queue-runner/build-remote.cc
```diff
@@ -301,30 +298,25 @@ static BuildResult performBuild(
...
-            result.startTime = start;
-            result.stopTime = stop;
-        }
+        result = ServeProto::Serialise<BuildResult>::read(localStore, conn);
     }
-    if (GET_PROTOCOL_MINOR(conn.remoteVersion) >= 6) {
-        ServeProto::Serialise<DrvOutputs>::read(localStore, conn);
+    stopTime = time(0);
+
+    if (!result.startTime) {
```
Ericson2314, line 312: Since it was `start && start` before, i.e. `start`, now the converse condition is `!start`.

src/hydra-queue-runner/build-remote.cc
Ericson2314: FWIW this was added in 9989e6c0f45a3b2723206af32acdf76bb1994dd2

tomberek (approved):

PR #1320, opened 2023-12-08
Let tests themselves intentionally leak temp dir

Ericson2314:
By default Yath will clean up temporary files, so the result is the same. But `--keep-dirs` can be passed to `yath test` telling Yath to *not* clean them up instead. This is very useful for debugging.

### DIFF
--- a/t/lib/HydraTestContext.pm
+++ b/t/lib/HydraTestContext.pm
@@ -39,7 +39,9 @@ use Hydra::Helper::Exec;
 sub new {
     my ($class, %opts) = @_;
 
-    my $dir = File::Temp->newdir();
+    # Cleanup will be managed by yath. By the default it ill be cleaned
+    # up, but can be kept to aid in debugging test failures.
+    my $dir = File::Temp->newdir(CLEANUP => 0);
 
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
When building in a nix derivation er seine News to worry to clean up. It's building outside of that a thing? And if so, do we need to worry about quickly flooding people's disks?

Ericson2314:
@SuperSandro2000 the title is a bit misleading --- the tests *themselves* no longer clean up, but the test harness does (unless one opts out), so even in `nix-shell` by default one will not leak temp stuff.

SuperSandro2000:
Noise then this sounds like a no brainer, LGTM

t/lib/HydraTestContext.pm
```diff
@@ -39,7 +39,9 @@ use Hydra::Helper::Exec;
 sub new {
     my ($class, %opts) = @_;
 
-    my $dir = File::Temp->newdir();
+    # Cleanup will be managed by yath. By the default it ill be cleaned
```
cole-h, line 42: nit:
```suggestion
    # Cleanup will be managed by yath. By the default it will be cleaned
```

tomberek (approved):
module small nits

PR #1321, opened 2023-12-09
Mere `master` into `nix-next`

Ericson2314:
Since `nix-next` is like the `staging` branch, could we reuse the automation to do this automatically?


Ericson2314:
Oh CI won't test the merge so it's not useful here. Built locally instead and it passed.

PR #1324, opened 2023-12-10
Use `ServeProto::Serialise<ServeProto::BuildOptions>`

Ericson2314:
The interesting part is the preparatory commit:

> Do not attempt to speak a newer version of the protocol
>
> Both sides need to agree on a version (with `std::min`) for anything to work. Somehow... we've never done this.
>
> With this comment, the next commit succeeds. Without this commit, the next commit fails. This is because the next commit exposes serializers which do different things for proto version 2.7, and we're currently requesting 2.6.
>
> Opened https://github.com/NixOS/nix/issues/9584 to track this issue


Ericson2314:
https://github.com/NixOS/nix/issues/9584 is now idea-approved, which is the one question I had for this. The rest is entirely rote using serializers and changing no behavior, so I will merge this.

PR #1328, opened 2023-12-18
Document the `store_uri` parameter by way of example

JackKelly-Bellroy:
It took me far too long to find out how to do this properly, so I'm writing it down for posterity. If you want to start a configuration reference page in the docs, I'm happy to do some of that in this PR.


Ericson2314:
Thank you! That was a glaring omission.

PR #1329, opened 2023-12-24
Clean up `std::optional` dereferencing in the queue runner

Ericson2314:
Instead of doing this partial operation a number of times, assert (with a comment, get a reference to the thing inside, and use that just once. (This refactor was done twice, "just once" for each time.)


Ericson2314:
This will also shrink the diff for https://github.com/NixOS/hydra/pull/1316, which is my main motivation.

PR #1335, opened 2024-01-11
api: fix /eval/X/builds requiring 5*len(builds)+1 SQL queries

delroth:
Prefetch related tables that get fetched as part of serialization to JSON. Drop the `jobsetevals` field since making it efficient is trickier (due to not using DBIx native relationships but instead a many_to_many through a custom table). There is likely a way we could keep it and make it efficient, but I'm not fluent enough in DBIx to know.

-----

More context in https://github.com/NixOS/nixos-org-configurations/issues/270 . This used to """""work""""" (painfully inefficiently) but stopped working once the hydra.nixos.org got moved to a separate machine, inducing higher latency costs for each of the hundreds of thousands of SQL queries that need to run to generate the JSON output for this API endpoint.


delroth:
Tested on hydra.nixos.org, fetching /builds for haskell-updates eval ID 1798805.

Without change: times out after 15min.
With change: 2min23

Still too slow / too much data for a nixpkgs/trunk eval unfortunately.

Ericson2314:
The `jobsetevals` part I am not sure how to judge. The #1093 folllow-up part definitely seems good. The `prefetch` part is probably fine too but I don't know this DBX thing :)

Ma27:
> The prefetch part is probably fine too but I don't know this DBX thing :)

Tried this out on my personal instance with query log being activated and that works fine :+1:

mweinelt:
Breaks the nixos-channel-scripts which depends on `jobsetevals` field.

https://github.com/NixOS/nixos-channel-scripts/blob/master/mirror-nixos-branch.pl#L103

PR #1339, opened 2024-01-22
Use Nix's `SSHMaster`

Ericson2314:
Not sure why this doesn't work yet.

~~Depends on https://github.com/NixOS/nix/pull/9833~~

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -21,12 +21,6 @@
 
 using namespace nix;
 
-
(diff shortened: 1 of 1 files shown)


mweinelt:
This introduces a regression for machine definitions that have an SSH hostkey specified, because it gets base64 decoded in hydra, then passed to SSHMaster in nix which wants to base64 decode it again.

Fix applied for h.n.o is https://github.com/NixOS/hydra/commit/031aaa065bf001099deea946a3ff6bb87766fb7a.

PR #1346, opened 2024-01-25
Clean up the flake/build in a a few ways

Ericson2314:
Nix has done corresponding things already. The motivation is the same.

- Split out `package.nix`, which should closely match what Nixpkgs contains.
- Use `fileset` library to filter sources to avoid rebuilds
- Clean up deps (`strictDeps`, `check*Inputs`)
- Create hydra job for build without tests (make sure we have our check vs unconditional deps right)
- Move NixOS modules out of `flake.nix`, have them be referenced by `flake.nix`
- `.gitignore` `result-*` too


package.nix
```diff
@@ -0,0 +1,277 @@
...
+        DBDPg
+        DBDSQLite
+        DigestSHA1
+        EmailMIME
+        EmailSender
+        FileLibMagic
+        FileSlurper
+        FileWhich
+        # Not Perl
+        nix.perl-bindings
```
SuperSandro2000, line 96:
```suggestion
        nix.perl-bindings
        # Not Perl
```

package.nix
```diff
@@ -0,0 +1,277 @@
...
+    ]);
+  };
+
+  strictDeps = true;
+
+  nativeBuildInputs = [
+    makeWrapper
+    autoreconfHook
+    automake
+    libtool
```
SuperSandro2000, line 166:
```suggestion
```
already in autoreconfHook

SuperSandro2000:
> Split out `package.nix`, which should closely match what Nixpkgs contains.

We should try to sync them after this PR.

PR #1348, opened 2024-01-26
More CA derivations prep

Ericson2314:
Again, with care not to change the schema in any way.


Ericson2314:
These changes will have no effect when CA derivations are disabled.

PR #1349, opened 2024-01-26
Allow building content-addressed derivations with hydra, minimally

Ericson2314:
This verison has a worse UI, but also chnages the schema less: One non-null constraint is removed, but no new columns are added.


Mindavi:
What exactly is meant by worse UI?

Ericson2314:
@Mindavi The other one used the new `contentAddressed` column to show more info. This one, you cannot tell the difference after an output is built and the path column is filled in.

Ericson2314:
This PR is now basically just the Schema change. I added a migration and tested it with https://hydra.ngi0.nixos.org/ and the migration worked just fine. I think we are all set!

PR #1352, opened 2024-01-26
Build: use `nix path-info` to test for build output availability

Ma27:
cc @dasj @Ericson2314 

Hacky workaround for https://github.com/NixOS/nix/issues/9859, but after I managed to fix the issue on my end, I decided I could still submit it against upstream.

------

When using the Perl bindings, this usually takes the wrong store, i.e. `auto` even if `store_uri` is a flat-file binary cache.

See also https://github.com/NixOS/nix/issues/9859

Using fork/exec here on purpose to make sure that the output of `nix path-info` isn't spammed into stderr. This could probably done with `system("foo &>/dev/null")`, but I didn't want to worry about shell injection issues then. And finally, I can't seem to silence STDERR for `system` with a list as argument. Not a Perl dev though, so not sure if there's a better way.

### DIFF
--- a/src/lib/Hydra/Controller/Build.pm
+++ b/src/lib/Hydra/Controller/Build.pm
@@ -71,6 +71,17 @@ sub findBuildStepByDrvPath {
 
 sub build :Chained('buildChain') :PathPart('') :Args(0) :ActionClass('REST') { }
 
+sub isValidPathNixCLI {
+    my $path = shift;
+    my $pid = fork();
+    if (!$pid) {
+        open STDERR, ">", "/dev/null";
+        exec "nix", "path-info", $path, "--store", getStoreUri();
+    }
+    waitpid $pid, 0;
+    return $? == 0;
+}
+
 sub build_GET {
     my ($self, $c) = @_;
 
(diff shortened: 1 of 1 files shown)


Ericson2314:
I don't quite understand the things you are saying in that issue, but I soon as well just add the missing functionality. We have a `nix-next` branch so we can try it out right away.

lheckemann:
Does this even work when using the nixos module? `hydra-server.service` doesn't seem to have nix on its `PATH`. I also wonder if this is an appropriate approach given the other ways in which the nix command could fail (`experimental-features = nix-command` missing?) and with no distinction drawn between that and the path not being valid.

Ma27:
It definitely works since I just realized I forgot to silence stdout and now I get paths printed to hydra-server's journal whenever I open a build view. Not sure why yet :upside_down_face: 

> could fail (experimental-features = nix-command missing?)

Fair. Should probably use a nix-* equivalent then :shrug:

Ericson2314:
I am looking at the perl binding stuff.

Ericson2314:
I made https://github.com/NixOS/nix/pull/9863 to generalize the Perl bindings so we can fix this the right way. It would be nice to replace this with PR against `nix-next` using that branch.

I think we can basically do a big `sed` to update Hydra: it's fine to call `openStore()` again and again (without an argument, that is) because there is caching.

Ma27:
Not the right solution to the problem. Closing.

Issue #1353, opened 2024-01-27
Git submodules not checked out in flake project, even with '?submodules=1' in flake uri.


RichardWarfield:
**Describe the bug**
Git submodules not checked out in flake project, even with '?submodules=1' in flake uri.

**To Reproduce**
A reproducing example is `proj2` [here](https://github.com/RichardWarfield/proj2) and has has [proj1](https://github.com/RichardWarfield/proj2) as a git submodule.  

Create a jobset with flake URI of `git+ssh://git@github.com/RichardWarfield/proj2?submodules=1`

**Expected behavior**
The job checks out the git repo with the required submodule and builds the project successfully

**Log**
```
Running phase: unpackPhase
unpacking source archive /nix/store/cadzicaa78vyvbdy1f7r4aqv1h4fiwnf-ijhnhy852xcxdb1qd8166gs4x3l9a4r3-source
source root is ijhnhy852xcxdb1qd8166gs4x3l9a4r3-source
Running phase: patchPhase
Running phase: updateAutotoolsGnuConfigScriptsPhase
Running phase: configurePhase
no configure script, doing nothing
Running phase: buildPhase
cc1: fatal error: ./proj1/src/hello_world.c: No such file or directory
compilation terminated.
```
**Hydra Server:**

- OS and version: [e.g. NixOS 24.05pre576254.ae5c332cbb58]
- Version of Hydra: 

```
[nix-shell:/etc/nixos]# hydra-server --version
/nix/store/pyh4fx1ak03nl96rawdka35a8y30v3pq-hydra-2023-12-24/bin/.hydra-server-wrapped
(Getopt::Long::GetOptions version 2.54; Perl version 5.38.2)
```

- Version of Nix Hydra is built against: 2.19.3
- Version of the Nix daemon

PR #1356, opened 2024-02-03
Workaround for fetching git submodules for flakes

RichardWarfield:
Workaround/fix for [1353](https://github.com/NixOS/hydra/issues/1353)

If a flake URI contains 'submodules=1' in the query part then any submodules should be fetched/initialized.

At present, hydra-eval-jobset uses the 'url' field of the flake metadata as the flake URI.  This URI is missing the 'submodules=1' query, which is present elsewhere in the metadata.

This PR appends "&submodules=1" if the flake metadata indicates submodules are used (specifically, if the resolved.submodules field is true).

### DIFF
--- a/src/script/hydra-eval-jobset
+++ b/src/script/hydra-eval-jobset
@@ -645,10 +645,16 @@ sub checkJobsetWrapped {
     if (defined $flakeRef) {
         (my $res, my $json, my $stderr) = captureStdoutStderr(
             600, "nix", "flake", "metadata", "--refresh", "--json", "--", $flakeRef);
+
         die "'nix flake metadata' returned " . ($res & 127 ? "signal $res" : "exit code " . ($res >> 8))
             . ":\n" . ($stderr ? decode("utf-8", $stderr) : "(no output)\n")
             if $res;
-        $flakeRef = decode_json($json)->{'url'};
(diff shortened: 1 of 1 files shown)


src/script/hydra-eval-jobset
```diff
@@ -645,10 +645,16 @@ sub checkJobsetWrapped {
...
             600, "nix", "flake", "metadata", "--refresh", "--json", "--", $flakeRef);
+
         die "'nix flake metadata' returned " . ($res & 127 ? "signal $res" : "exit code " . ($res >> 8))
             . ":\n" . ($stderr ? decode("utf-8", $stderr) : "(no output)\n")
             if $res;
-        $flakeRef = decode_json($json)->{'url'};
+        my $decoded = decode_json($json);
+        my $url = $decoded->{'url'};   # Doesn't have the &submodules=1, so add it back if needed
+        if ($decoded->{'resolved'}->{'submodules'}) {
+            $url = $url . "&submodules=1";
```
Ma27, line 655: Is it safe that each `$url` has a query param already (i.e. `?foo=bar`), otherwise the URL (`git+file://foo&submodules=1` e.g.) is probably invalid.

Ma27 (commented):
I think a testcase for that would be nice.

PR #1359, opened 2024-02-07
Update perl bindings for nix#9863

Ma27:
For https://github.com/NixOS/nix/pull/9863

cc @Ericson2314 @dasJ

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -42,11 +42,11 @@
         "nixpkgs-regression": "nixpkgs-regression"
       },
       "locked": {
-        "lastModified": 1706629374,
-        "narHash": "sha256-KyAiLGxJ39fSY0cuq8EWAZQ4vaDdqAItSRP4+vjYvq8=",
+        "lastModified": 1707750141,
+        "narHash": "sha256-9qSzGQs/Rjf2i3UQjyaZUznZzYDHkLYro/1FTT1easg=",
         "owner": "NixOS",
         "repo": "nix",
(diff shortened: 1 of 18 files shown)


Ericson2314:
OK the underlying Nix PR is merged!

@Ma27 can you spit this up so the 1st (or 1st and part of 4th) commit is in this PR, and the possibly-behavior-changing middle parts are in a second PR?

I just want to be very cautious :)

Ma27:
> I just want to be very cautious and have a nice rollback ready in case something goes wrong :).

Sure, this PR only contains the `MACHINE_LOCAL_STORE` changes. A second PR will be opened in a few minutes.

> (or 1st and part of 4th)

Not necessary IMHO: w/o the BINARY_CACHE_STORE it's not necessary to read from `hydra.conf` and thus the include test doesn't break. Also this seemed to painful to split and push up in the history, so I figured it's OK that way.

flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "A Nix-based continuous build system";
 
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.05-small";
-  inputs.nix.url = "github:NixOS/nix";
+  inputs.nix.url = "github:obsidiansystems/nix/perl-open-other-store";
```
Ericson2314, line 5: This is no longer needed, because I merged the underlying Nix PR!

Ma27:
Done! Required a few changes to the cpp code though.

Ericson2314 (approved):
Looking good!

PR #1361, opened 2024-02-12
flake: fix gitea integration test

Ma27:
This is an integration test that confirms that jobset definitions from git repositories are correctly built and status updates pushed to the gitea instance. The following things needed to be fixed:

* We're still on 23.05 where gitea is marked as insecure. Not going to update nixpkgs right now, but going for the quick fix.
* Since gitea 1.19 tokens have scopes that describe what's possible. Not specifying the scope in the DB appears to imply that no permissions are granted.
* Apparently we have three status updates now (for three status hooks, queued/started/finished). No idea why that was broken before, but the behavior still looks correct.

cc @Ericson2314 is there anything keeping us from updating nixpkgs btw?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -180,13 +180,9 @@
                 root=d7f16a3412e01a43a414535b16007c6931d3a9c7
                 </gitea_authorization>
               '';
+              nixpkgs.config.permittedInsecurePackages = [ "gitea-1.19.4" ];
               nix = {
-                distributedBuilds = true;
-                buildMachines = [{
-                  hostName = "localhost";
-                  systems = [ system ];
-                }];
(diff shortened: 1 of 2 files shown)


ajs124:
Still fails for me with `AssertionError: Expected exactly three status updates for latest commit (queued, started, finished)!
`.

Ma27:
This is where I'd like to have a :melting_face: as reaction, GitHub!
Thanks, managed to reproduce it after several rounds of `nix build --rebuild`. Going to fix that.

Ma27:
@ajs124 fixed the test by removing a redundant status update from the plugin (rationale is in the commit message).

PR #1364, opened 2024-02-23
urlencode drv names when fetching logs

K900:
Otherwise names with special characters like + break things.


K900:
Turns out the issue is also on the cache end :(

K900:
Oops wrong button

K900:
@delroth fixed the other part.

tomberek (approved):

PR #1368, opened 2024-03-07
Use `submit` event in login form

Ma27:
It's a pet peeve from me when logging into my personal Hydra that I always have to press the button rather than hitting Return after entering my password.

Reason for that is that the form doesn't have a "submit" button, so far it was always listened to the "click" event. Submit does that and you can hit Return alternatively.

cc @dasj @Ericson2314


samueldr (approved):
I have not tested, but this is basic UX design with forms and the most proper fix assuming it has to happen via JS, rather than just POSTing the form.

mweinelt:
Deployed on hydra.nixos.org. Finally my password manager is working. Thanks!

mweinelt (approved):

PR #1370, opened 2024-03-15
hydra-queue-runner: drop broken connections from pool

### LINKED ISSUE #1336: hydra-queue-runner doesn't reconnect after a postgresql server restart, staying stuck forever
**Describe the bug**

After we restarted the PostgreSQL database for hydra.nixos.org, the queue runner just got stuck logging it lost its connection and never reconnected on its own. I had to manually restart it.

**Expected behavior**
The queue runner properly and gracefully recovers from a temporary PostgreSQL connection failure.

**Hydra Server:**

Please fill out this data as well as you can, but don't worry if you can't -- just do your best.

- Version of Hydra: 00d30874da759eb0f44f446415b2469920ff41b5
- Version of Nix Hydra is built against: from flake

**Additional context**

Logged forever every 10s:

```
Jan 12 22:14:13 rhea hydra-queue-runner[4016980]: main thread: Lost connection to the database server.
Jan 12 22:14:13 rhea hydra-queue-runner[4016980]: queue monitor: Lost connection to the database server.
```

Ma27:
Closes #1336

When restarting postgresql, the connections are still reused in `hydra-queue-runner` causing errors like this

    main thread: Lost connection to the database server.
    queue monitor: Lost connection to the database server.

and no more builds being processed.

`hydra-evaluator` doesn't have that issue since it crashes right away. We could let it retry indefinitely as well (see below), but I don't want to change too much.

If the DB is still unreachable 10s later, the process will stop with a non-zero exit code because of a missing DB connection. This however isn't such a big deal because it will be immediately restarted afterwards. With the current configuration, Hydra will never give up, but restart (and retry) infinitely. To me that seems reasonable, i.e. to retry DB connections on a long-running process. If this doesn't work out, the monitoring should fire anyways because the queue fills up, but I'm open to discuss that.

Please note that this isn't reproducible with the DB and the queue runner on the same machine when using `services.hydra-dev`, because of the `Requires=` dependency `hydra-queue-runner.service` -> `hydra-init.service` -> `postgresql.service` that causes the queue runner to be restarted on `systemctl restart postgresql`.

Internally, Hydra uses Nix's pool data structure: it basically has N slots (here DB connections) and whenever a new one is requested, an idle slot is provided or a new one is created (when N slots are active, it'll be waited until one slot is free). The issue in the code here is however that whenever an error is encountered, the slot is released, however the same broken connection will be reused the next time. By using `Pool::Handle::markBad`, Nix will drop a broken slot. This is now being done when `pqxx::broken_connection` was caught.

cc @NixOS/infra

### DIFF
--- a/src/hydra-queue-runner/hydra-queue-runner.cc
+++ b/src/hydra-queue-runner/hydra-queue-runner.cc
@@ -926,10 +926,17 @@ void State::run(BuildID buildOne)
     while (true) {
         try {
             auto conn(dbPool.get());
-            receiver dumpStatus_(*conn, "dump_status");
-            while (true) {
-                conn->await_notification();
-                dumpStatus(*conn);
+            try {
+                receiver dumpStatus_(*conn, "dump_status");
+                while (true) {
+                    conn->await_notification();
+                    dumpStatus(*conn);
+                }
+            } catch (pqxx::broken_connection & connEx) {
+                printMsg(lvlError, "main thread: %s", connEx.what());
+                printMsg(lvlError, "main thread: Reconnecting in 10s");
+                conn.markBad();
+                sleep(10);
             }
         } catch (std::exception & e) {
             printMsg(lvlError, "main thread: %s", e.what());
--- a/src/hydra-queue-runner/queue-monitor.cc
+++ b/src/hydra-queue-runner/queue-monitor.cc
@@ -10,8 +10,14 @@ using namespace nix;
 void State::queueMonitor()
 {
     while (true) {
+        auto conn(dbPool.get());
         try {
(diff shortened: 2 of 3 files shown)


mweinelt:
Deployed on hydra.nixos.org. We'll see on the next database reboot how well this will work.

Ericson2314:
Thanks @Ma27. This makes sense to me.

zowoq:
@Ericson2314 @Mic92 

https://git.lix.systems/lix-project/hydra/commit/44b9a7b95d23e7a8587cb963f00382046707f2db is a follow up for the changes in this but it doesn't apply cleanly.

PR #1371, opened 2024-03-16
foreman/start-hydra: enable query log in postgres

Ma27:
That way it's possible to observe which queries are fired against the DB from DBIx. This is particularly useful to find out if DBIx correctly prefetches everything it needs or if too much / too few things are loaded, see e.g. #1335.


delroth:
Another way (what I've been doing) is DBIC_TRACE=1 in the environment variables when starting the Perl server. Maybe better?

Ma27:
Oh yes it is, TIL!

PR #1372, opened 2024-03-16
Running builds view: show build step names

Ma27:
When using Hydra to build machine configurations, you'll often see "nixosConfigurations.foo" five times, i.e. for each build step being run. This isn't very helpful I think because in such a case, a single build step can also be compiling the Linux kernel.

This change also fetches the `drvpath` and `type` from the `buildsteps` relation. We're already joining it, so this doesn't make much difference (confirmed via query logging that this doesn't cause extra SQL queries).

Unfortunately build steps don't have a human readable name, so I'm deriving it from the drvpath by stripping away the hash (assuming that it'll never contain a `-` and that `/nix/store/` is used as prefix). I decided against using the Nix bindings for that to avoid too much overhead due to store operations for each build step.

### DIFF
--- a/src/root/common.tt
+++ b/src/root/common.tt
@@ -91,6 +91,15 @@ BLOCK renderDuration;
   duration % 60 %]s[%
 END;
 
+BLOCK renderDrvInfo;
+  drvname = step.drvpath
+    .substr(11) # strip `/nix/store/`
+    .split('-').slice(1).join("-") # strip hash part
+    .substr(0, -4); # strip `.drv`
+  IF step.type == 0; action = "Build"; ELSE; action = "Substitution"; END;
+  IF drvname; %]<em> ([% action %] of [% drvname %])</em>[% END;
+END;
+
 
 BLOCK renderBuildListHeader %]
   <table class="table table-striped table-condensed clickable-rows">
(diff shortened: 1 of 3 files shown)


mweinelt:
Annoyingly that page already takes >20s to load, but whether that is related to this change, or whether it was previously *this* slow (it was slow), is probably something you can tell.

https://hydra.nixos.org/status

mweinelt:
Also seems redundant with the `Package / release name` field.

image

Ma27:
> Also seems redundant with the Package / release name field.

We could skip that part in the UI if step name equals package name. It's only really interesting if you have something like a VM test or nixos configuration with lots of steps before.

Ma27:
I didn't get to deploy bd380f6 yet, but locally it seems as if it'd resolve the issue about showing too much information.

mweinelt:
Deployed on hydra.nixos.org.

src/root/common.tt
```diff
@@ -91,6 +91,17 @@ BLOCK renderDuration;
   duration % 60 %]s[%
 END;
 
+BLOCK renderDrvInfo;
+  drvname = step.drvpath
+    .substr(11) # strip `/nix/store/`
```
Ericson2314, line 96: `$Nix::Config::storeDir` I think this is a thing, can you use it?

Ericson2314:
@mweinelt 

> but whether that is related to this change, or whether it was previously this slow (it was slow), is probably something you can tell.

So you are fine with the performance of this, now that it deployed?

mweinelt:
No, https://hydra.nixos.org/status is really slow compared to https://hydra.nixos.org/machines.

zowoq:
These commits were merged in https://github.com/NixOS/hydra/pull/1462.

PR #1374, opened 2024-04-03
ca-derivations: fix rendering issue

Mindavi:
See commit messages for some more background and analysis.

I set up my hydra server with the recent content-addressed derivations support, and ran into some /build/\<id\> pages that didn't want to render correctly (not only ca-derivations, other pages also suffered from this). I think I found the issue and have worked around/resolved it.

Also adds a test that verifies that the page doesn't load properly before the patch is applied. In the next commit this bug is resolved.

```
[  FAIL  ]  job  1  + The 'build' page for build 'caDependingOnFailingCA' should load properly
```

### DIFF
--- a/src/lib/Hydra/Helper/Nix.pm
+++ b/src/lib/Hydra/Helper/Nix.pm
@@ -187,6 +187,10 @@ sub findLog {
 
     return undef if scalar @outPaths == 0;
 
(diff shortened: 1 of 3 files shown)


Ericson2314 (approved):
Very nice PR!

I think is fine to put that check there.

PR #1380, opened 2024-05-03
Factor out NixOS tests, and clean up

Ericson2314:
Due to newer nixpkgs, there were a number of things that could be cleaned up in the process.


Ericson2314:
Note that this *changes* the NixOS modules, such that they don't automatically drag in the Nixpkgs overlays and this hydra in particular. This allows a more targeted setting `services.hydra.package = ....` to use this Hydra while not impacting the rest of the system (if one wants to generally use a different nixpkgs / not use those overlays).

PR #1219, opened 2022-06-12
CompressLog: Add zstd compression

dasJ:

PR #1400, opened 2024-08-06
CompressLog: Add zstd compression

SuperSandro2000:
Basically #1219 rebased on master

- https://github.com/NixOS/nixpkgs/pull/333212

### DIFF
--- a/nixos-modules/hydra.nix
+++ b/nixos-modules/hydra.nix
@@ -459,10 +459,15 @@ in
     # logs automatically after a step finishes, but this doesn't work
     # if the queue runner is stopped prematurely.
     systemd.services.hydra-compress-logs =
-      { path = [ pkgs.bzip2 ];
+      { path = [ pkgs.bzip2 pkgs.zstd ];
         script =
           ''
-            find ${baseDir}/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            set -eou pipefail
+            compression=$(sed -nr 's/compress_build_logs_compression = ()/\1/p' ${baseDir}/hydra.conf)
+            if [[ $compression == zstd ]]; then
+              compression="zstd --rm"
+            fi
+            find ${baseDir}/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r $compression --force --quiet
           '';
         startAt = "Sun 01:45";
       };
--- a/nixos-modules/hydra.nix
+++ b/nixos-modules/hydra.nix
@@ -408,6 +408,7 @@ in
         requires = [ "hydra-init.service" ];
         after = [ "hydra-init.service" ];
         restartTriggers = [ hydraConf ];
+        path = [ pkgs.zstd ];
         environment = env // {
           PGPASSFILE = "${baseDir}/pgpass-queue-runner"; # grrr
           HYDRA_DBI = "${env.HYDRA_DBI};application_name=hydra-notify";
--- a/src/lib/Hydra/Plugin/CompressLog.pm
+++ b/src/lib/Hydra/Plugin/CompressLog.pm
@@ -9,11 +9,24 @@ use Hydra::Helper::CatalystUtils;
 sub stepFinished {
     my ($self, $step, $logPath) = @_;
 
-    my $doCompress = $self->{config}->{'compress_build_logs'} // "1";
+    my $doCompress = $self->{config}->{'compress_build_logs'} // '1';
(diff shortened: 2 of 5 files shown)


src/lib/Hydra/Plugin/CompressLog.pm
```diff
@@ -9,11 +9,24 @@ use Hydra::Helper::CatalystUtils;
...
+    }
+
+    if ($silent ne '1') {
+        print STDERR "compressing '$logPath' with $compression...\n";
+    }
+
+    if ($compression eq 'bzip2') {
+        system('bzip2', '--force', $logPath);
+    } elsif ($compression eq 'zstd') {
+        system('zstd', '-T0', $logPath);
```
dasJ, line 27: Something that still annoys me with the old patch is that zstd is not quiet. Could you add this or do you think this adds any benefit?

Mindavi:
See my comments @ nixpkgs.

src/lib/Hydra/Plugin/CompressLog.pm
SuperSandro2000: I also disabled it in https://github.com/NixOS/nixpkgs/pull/332752, we don't need three lines of useless progress bars.

src/lib/Hydra/Plugin/CompressLog.pm
```diff
@@ -9,11 +9,24 @@ use Hydra::Helper::CatalystUtils;
...
+    }
+
+    if ($silent ne '1') {
+        print STDERR "compressing '$logPath' with $compression...\n";
+    }
+
+    if ($compression eq 'bzip2') {
+        system('bzip2', '--force', $logPath);
+    } elsif ($compression eq 'zstd') {
+        system('zstd', '--quiet', -T0', $logPath);
```
Mindavi, line 27:
```suggestion
        system('zstd', '--quiet', '-T0', $logPath);
```

Mindavi:
Looks good to me, but haven't tested. Would really love it if a small test is added, but maybe that's a big ask. Would definitely improve my confidence, at least of the case where CompressLog is called. The fallback is probably a bit more annoying to test.

src/lib/Hydra/Plugin/CompressLog.pm
```diff
@@ -9,11 +9,24 @@ use Hydra::Helper::CatalystUtils;
...
+    }
+
+    if ($silent ne '1') {
+        print STDERR "compressing '$logPath' with $compression...\n";
+    }
+
+    if ($compression eq 'bzip2') {
+        system('bzip2', '--force', $logPath);
+    } elsif ($compression eq 'zstd') {
+        system('zstd', '--quiet', '-T0', $logPath);
```
SuperSandro2000, line 27:
```suggestion
        system('zstd', '--rm', '--quiet', '-T0', $logPath);
```

if we want to match bzip2, we need to rm the files.

SuperSandro2000:
> Would really love it if a small test is added, but maybe that's a big ask.

I have tested this on my hydra and from what I can tell, it seems to work minus the bit that the original logs where not removed. I don't have time to write a test especially with the current test infrastructure.

SuperSandro2000:
I did some more testing and the logs where not displayed in the web ui. That is now also fixed. I think I have now tested this change end to end and hopefully found all the misses along the way.

Mindavi (approved):
Diff now looks good to me, unfortunately I cannot test it this week. Maybe next week.

Out of curiosity, what happens if you fill in something else than bz2 or zstd? It'll just fail somewhere in the compression routines?

nixos-modules/hydra.nix
```diff
@@ -458,10 +459,15 @@ in
...
     systemd.services.hydra-compress-logs =
-      { path = [ pkgs.bzip2 ];
+      { path = [ pkgs.bzip2 pkgs.zstd ];
         script =
           ''
-            find ${baseDir}/build-logs -type f -name "*.drv" -mtime +3 -size +0c | xargs -r bzip2 -v -f
+            set -eou pipefail
+            compression=$(sed -nr 's/compress_build_logs_compression = ()/\1/p' ${baseDir}/hydra.conf)
+            if [[ $compression == zstd ]]; then
+              compression=zstd --rm
```
SuperSandro2000, line 468:
```suggestion
              compression="zstd --rm"
```

SuperSandro2000:
> Out of curiosity, what happens if you fill in something else than bz2 or zstd? It'll just fail somewhere in the compression routines?

I hope so. 😅 I didn't test this explicitly.

dasJ (approved):

PR #1402, opened 2024-08-21
tests: use `like` for testing regexes

Mic92:
This gives us better diagnostics when the test fails.


Mindavi:
I applied this to my hydra instance and the build still passes, so this LGTM.

PR #1408, opened 2024-09-08
set a default for hydra-compress-logs service

zowoq:
follow up from 99ca560d583239b1bc34485f2f719989642fa697


Mic92:
Merged in https://github.com/NixOS/hydra/pull/1412

PR #1388, opened 2024-05-23
`nix-2.21` -> `hydra.nixos.org`

Ericson2314:
Bump to Nix 2.21

~~The tests failed in a local build, I am guessing perhaps due to #1359. @Ma27, would you mind taking a look?~~

Issue #1315, opened 2023-12-04
Deduplicate with `nix-eval-jobs`


Ericson2314:
https://github.com/nix-community/nix-eval-jobs was forked from Hydra, because it is nice to be able to evaluate exactly as Hydra does outside of Hydra. This is a virtuous goal in my view --- a core Nix philosophy is that "everything CI does one should be able to do locally", *without* running Hydra locally of course.

But unfortunately Hydra still has its own `hydra-eval-jobs`, which means that we have two sources of truth now --- `nix-eval-jobs` and `hydra-eval-jobs` --- and nothing stops them from diverging. This is not good, and puts the original goes (do what hydra does locally without Hydra) at risk! The risk is not hypothetical: https://github.com/NixOS/hydra/pull/875 changes `hydra-eval-jobs` to support CA derivations, and I wouldn't want `nix-eval-jobs` to not have this first (or, if it already does, add another implementation).

Here area a couple options on how to fix this:

1. Make `hydra` depend on `nix-eval-jobs`. It is just an out of tree executable it will know how to use.
2. Make the in-tree thing `nix-eval-jobs`, and make sure it is possible to build in isolation. Converting Hydra to use Meson (#1392), like `nix-eval-jobs` does, could possibly help with this.

mweinelt:
We are on nix-eval-jobs now, so I think this issue may be complete?

PR #1411, opened 2024-09-20
Nix 2.24 upgrade

Mic92:
Probably now needs cherry-picking from https://github.com/NixOS/hydra/pull/1388

### DIFF
--- a/src/hydra-eval-jobs/Makefile.am
+++ b/src/hydra-eval-jobs/Makefile.am
@@ -1,5 +1,5 @@
 bin_PROGRAMS = hydra-eval-jobs
 
 hydra_eval_jobs_SOURCES = hydra-eval-jobs.cc
-hydra_eval_jobs_LDADD = $(NIX_LIBS) -lnixcmd
+hydra_eval_jobs_LDADD = $(NIX_LIBS) -lnixcmd -lnixflake
 hydra_eval_jobs_CXXFLAGS = $(NIX_CFLAGS) -I ../libhydra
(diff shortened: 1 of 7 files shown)


src/hydra-eval-jobs/Makefile.am
```diff
@@ -1,5 +1,5 @@
 bin_PROGRAMS = hydra-eval-jobs
 
 hydra_eval_jobs_SOURCES = hydra-eval-jobs.cc
-hydra_eval_jobs_LDADD = $(NIX_LIBS) -lnixcmd
+hydra_eval_jobs_LDADD = $(NIX_LIBS) -lnixcmd -lnixflake
```
Mic92, line 4: I am now thinking about making hydra-eval-jobs it's own derivation so that we can split the nix update over multiple pull requests.

Ericson2314: That would be good for deduplicating with nix-eval-jobs too 

https://github.com/NixOS/hydra/issues/1315

Also hope we can convert hydra to Meson and then use subprojects to still support incremental builds of everything.

Ericson2314:
Now Hydra builds, but the tests fail :(

Ericson2314:
```
hydra-queue-runner: build-result.cc:118: BuildOutput getBuildOutput(nix::ref<nix::Store>, NarMemberDatas&, nix::OutputPathMap): Assertion `file != narMembers.end()' failed.
```
This seems to be the issue in the failing tests

Ericson2314 (approved):

PR #1418, opened 2024-10-19
Make the in-tree package the default package

mweinelt:
There is an overlay for the `hydra` name, but `hydra_unstable` was used, which can refer to the nixpkgs package and lead to and outdated hydra version and requires configuring the correct package attribute downstream.


SuperSandro2000 (approved):

Mindavi:
The name in nixpkgs has also been updated to hydra a few weeks ago.

Mindavi (approved):

PR #1419, opened 2024-10-19
make nixos module hydra from this repository by default

Mic92:
When people reach out to the git repository they probably want to use hydra from the same source.
This also removes the need for an overlay with simpler and more performant direct use of the nixpkgs passed in. Before it was re-importing nixpkgs.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -9,14 +9,6 @@
     let
       systems = [ "x86_64-linux" "aarch64-linux" ];
       forEachSystem = nixpkgs.lib.genAttrs systems;
-
-      overlayList = [ self.overlays.default nix.overlays.default ];
-
-      pkgsBySystem = forEachSystem (system: import nixpkgs {
-        inherit system;
-        overlays = overlayList;
-      });
-
     in
     rec {
 
@@ -30,7 +22,6 @@
       };
 
       hydraJobs = {
-
         build = forEachSystem (system: packages.${system}.hydra);
 
         buildNoTests = forEachSystem (system:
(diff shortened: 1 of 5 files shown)


Mic92:
cc @Mindavi for review

Mindavi (approved):
Seems like a good improvement and I don't see any obvious mistakes. I'm personally not using the flake though.

Ericson2314:
Note I was trying to make fewer things depend on the flake before / not potentially mix Nixpkgs downstream by default. But this is fine.

I realized one of the things I put in nixos-tests is not a NixOS test; that is moved out maybe that file doesn't need `nixpkgs.legacyPackages` anymore.

Mic92:
People that used the nixos module instead of the flake now have to explicitly have to set the hydra package. I believe this is less error-prone interface than assuming they imported the overlay correctly.

flake.nix
```diff
@@ -64,20 +67,24 @@
...
 
       packages = forEachSystem (system: {
-        hydra = pkgsBySystem.${system}.hydra;
-        default = pkgsBySystem.${system}.hydra;
+        hydra = nixpkgs.legacyPackages.${system}.callPackage ./package.nix {
+          inherit (nixpkgs.lib) fileset;
+          rawSrc = self;
+          nix = nix.packages.${system}.nix;
+          nix-perl-bindings = nix.hydraJobs.perlBindings.${system};
+        };
```
Ericson2314, line 75: The duplicating of the `overlay` also tripped me up, forgot to add `nix-eval-jobs` in both places.

PR #1421, opened 2024-11-13
Use `nix-eval-jobs` and delete `hydra-eval-jobs`

Ericson2314:
`nix-eval-jobs` was forked from `hydra-eval-jobs` originally, for others to use. That's great! But maintaining two copies is no good. But we use that.

Original author is @delroth, for Lix's Hydra, thank you!

Draft because some tests fail. I am not sure why. Maybe there are fixes on the Lix branch that can be cherry-picked to fix? Not sure!

PR #1422, opened 2024-11-13
autotools -> meson 

Ericson2314:
Original author is @delroth, for Lix's Hydra, thank you!

Depends on #1421


Mindavi:
```
hydra> doc/manual/meson.build:29: WARNING: Project does not target a minimum version but uses feature deprecated since '0.60.0': install_subdir with empty directory. It worked by accident and is buggy. Use install_emptydir instead.
hydra> Build targets in project: 3
hydra> WARNING: Deprecated features used:
hydra>  * 0.60.0: {'install_subdir with empty directory'}
```

Mindavi:
Trying this out now (so including nix-eval-jobs) using this PR by zowoq (thanks to them!). But having some issues with the CSS (I think?), see https://github.com/NixOS/nixpkgs/pull/356391#issuecomment-2481629537

Ericson2314:
@Mindavi I had trouble finding that exact message, but I think I fixed the issue?

Ericson2314:
OK matching what I wrote in https://github.com/NixOS/hydra/pull/1421#issuecomment-2481584696, this one seems easier so doing it first. It now doesn't depend on that, and should be ready to go.

Mindavi:
Some things I noticed while running the phases manually (no action needed, but just wanted to note it down somewhere):

```
[11/13] Compiling C++ object src/hydra-queue-runner/hydra-queue-runner.p/queue-monitor.cc.o
../src/hydra-queue-runner/queue-monitor.cc: In member function ‘Jobset::ptr State::createJobset(pqxx::work&, const std::string&, const std::string&, JobsetID)’:
../src/hydra-queue-runner/queue-monitor.cc:623:18: warning: ‘bool pqxx::row::empty() const’ is deprecated: Row slicing is going away. [-Wdeprecated-declarations]
  623 |     if (res.empty()) throw Error("missing jobset - can't happen");
      |         ~~~~~~~~~^~
In file included from /nix/store/knjg1dsi6abrqb4hh3aid715i9jbfzp2-libpqxx-7.7.5/include/pqxx/transaction_base.hxx:37,
                 from /nix/store/knjg1dsi6abrqb4hh3aid715i9jbfzp2-libpqxx-7.7.5/include/pqxx/dbtransaction.hxx:20,
                 from /nix/store/knjg1dsi6abrqb4hh3aid715i9jbfzp2-libpqxx-7.7.5/include/pqxx/blob.hxx:34,
                 from /nix/store/knjg1dsi6abrqb4hh3aid715i9jbfzp2-libpqxx-7.7.5/include/pqxx/pqxx:6,
                 from ../src/libhydra/db.hh:3,
                 from ../src/hydra-queue-runner/state.hh:15,
                 from ../src/hydra-queue-runner/queue-monitor.cc:1:
/nix/store/knjg1dsi6abrqb4hh3aid715i9jbfzp2-libpqxx-7.7.5/include/pqxx/row.hxx:184:3: note: declared here
  184 |   empty() const noexcept;
      |   ^~~~~
```

Performance of the build in general seems a lot better (due to configure not happening anymore, but also other things happened that took a lot of time and I didn't understand).

It is a bit unfortunate that meson is now showing output of yath (running the tests) directly, but not a really big deal to me. I guess the silent timeout for h.n.o. is long enough to not cause issues there. But still a bit weird to look at without direct output.

From the outputs there are some diffs:

- bin/
  - missing: hydra-dev-server
- lib/hydra/root/
  - missing; 158 files (mostly javascript, css files)
- lib/hydra/sql/
  - missing: update-dbix-harness.sh
- lib/hydra/ttf/
  - missing: README.txt (doesn't seem important)
- share/
  - missing 55 files (html, woff2, javascript, css), the whole share/doc/hydra folder is missing

Feels like some things don't get installed as they should be? Or are these files all unnecessary?

Mindavi:
Ah, for the documentation I guess I was looking at the wrong output, oops...

Ericson2314:
Thanks @Mindavi! OK I'll try to get that in sync.

Mindavi:
FYI, I think the people at Lix removed some vendored Javascript files. That may be the cause of them not being handled/included in the meson build files.

Ericson2314:
OK @Mindavi, think the outputs are not missing anything now!

src/sql/meson.build
```diff
@@ -0,0 +1,90 @@
+sql_files = files(
```
dasJ, line 1: It'd be nice if we could glob this but it looks like that's not how it's supposed to be done in meson land

dasJ (approved):
I diffed through the builds and found these files missing:
```diff
@@ -565,70 +565,5 @@
 /libexec/hydra/ttf/StayPuft.ttf
 /nix-support
 /share
-/share/doc
-/share/doc/hydra
-/share/doc/hydra/404.html
-/share/doc/hydra/about.html
-/share/doc/hydra/api.html
-/share/doc/hydra/ayu-highlight.css
-/share/doc/hydra/book.js
-/share/doc/hydra/clipboard.min.js
-/share/doc/hydra/configuration.html
-/share/doc/hydra/css
-/share/doc/hydra/css/chrome.css
-/share/doc/hydra/css/general.css
-/share/doc/hydra/css/print.css
-/share/doc/hydra/css/variables.css
-/share/doc/hydra/elasticlunr.min.js
-/share/doc/hydra/favicon.png
-/share/doc/hydra/favicon.svg
-/share/doc/hydra/FontAwesome
-/share/doc/hydra/FontAwesome/css
-/share/doc/hydra/FontAwesome/css/font-awesome.css
-/share/doc/hydra/FontAwesome/fonts
-/share/doc/hydra/FontAwesome/fonts/FontAwesome.ttf
-/share/doc/hydra/FontAwesome/fonts/fontawesome-webfont.eot
-/share/doc/hydra/FontAwesome/fonts/fontawesome-webfont.svg
-/share/doc/hydra/FontAwesome/fonts/fontawesome-webfont.ttf
-/share/doc/hydra/FontAwesome/fonts/fontawesome-webfont.woff
-/share/doc/hydra/FontAwesome/fonts/fontawesome-webfont.woff2
-/share/doc/hydra/fonts
-/share/doc/hydra/fonts/fonts.css
-/share/doc/hydra/fonts/OPEN-SANS-LICENSE.txt
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-300italic.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-300.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-600italic.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-600.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-700italic.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-700.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-800italic.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-800.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-italic.woff2
-/share/doc/hydra/fonts/open-sans-v17-all-charsets-regular.woff2
-/share/doc/hydra/fonts/SOURCE-CODE-PRO-LICENSE.txt
-/share/doc/hydra/fonts/source-code-pro-v11-all-charsets-500.woff2
-/share/doc/hydra/hacking.html
-/share/doc/hydra/highlight.css
-/share/doc/hydra/highlight.js
-/share/doc/hydra/index.html
-/share/doc/hydra/installation.html
-/share/doc/hydra/introduction.html
-/share/doc/hydra/jobs.html
-/share/doc/hydra/mark.min.js
-/share/doc/hydra/monitoring
-/share/doc/hydra/monitoring/index.html
-/share/doc/hydra/.nojekyll
-/share/doc/hydra/notifications.html
-/share/doc/hydra/plugins
-/share/doc/hydra/plugins/declarative-projects.html
-/share/doc/hydra/plugins/index.html
-/share/doc/hydra/plugins/RunCommand.html
-/share/doc/hydra/print.html
-/share/doc/hydra/projects.html
-/share/doc/hydra/searcher.js
-/share/doc/hydra/searchindex.js
-/share/doc/hydra/searchindex.json
-/share/doc/hydra/tomorrow-night.css
-/share/doc/hydra/webhooks.html
 /share/nix
 /share/nix/hydra-module.nix
```

which is to be expected since we now put the manual in a separate output.

The wrappers are also different (which is to be expected since $out changed). Nothing else about $out is different after the change.

src/sql/meson.build
Ericson2314: Someday Meson will have a Nix backend using dynamic derivations, and then this will work :)

Mindavi:
Woo! 🥳

PR #1425, opened 2024-11-28
Allow globbing for named constituents

Ma27:
__Note:__ I'm well-aware of the ongoing effort to switch to `nix-eval-jobs`. However, when this was implemented, I decided to stick with the Hydra version from nixpkgs 24.05 because we wanted to roll it out internally.
I'm filing the current state to see if there's any interest in accepting these patches. If that's the case, I'm happy to port the relevant changes to `nix-eval-jobs` and add some more documentation on it.

---

This basically allows you to do `constituents = [ "nixos.*" ];` to select all jobs starting with `nixos.` as constituents. Additionally, it fixes a bug in Hydra where transitive constituent jobs may be discarded silently if declared by string instead of by derivation.

The commit messages go pretty much into detail what happened here, so I suggest to read those.

cc @ctheune I hope it's OK I also included your patch to log memory usage per job, I figured it's also useful even though it probably needs to be ported to nix-eval-jobs.
cc @Ericson2314 @Mic92 for input.

### DIFF
--- a/src/hydra-eval-jobs/hydra-eval-jobs.cc
+++ b/src/hydra-eval-jobs/hydra-eval-jobs.cc
@@ -27,6 +27,8 @@
 #include <sys/wait.h>
 #include <sys/resource.h>
 
+#include <fnmatch.h>
+
 #include <nlohmann/json.hpp>
 
 void check_pid_status_nonblocking(pid_t check_pid)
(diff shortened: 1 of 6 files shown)


Ericson2314:
@Ma27 this sounds like a good plan to me. New features that also work / can be tested in nix-eval-jobs alone are much less costly.

If you can take over the nix-eval-jobs for hydra PR in addition to porting this feature to nix-eval-jobs, that would be even better, and I would personally appreciate getting that off my plate too. :)

Only thing to keep in mind is that I would like to do the nix-eval-jobs switchover prior to upgrading hydra to Nix 2.25, so please do the work off the 2.24 branch in the nix-eval-jobs repo.

Ericson2314 (approved):

PR #1431, opened 2024-12-18
Add a plugin to poll Gitea pull requests

fx-chun:
Based off the existing `GithubPulls.pm` and `GitlabPulls.pm` plugins.

Also adds an integration test for the new `giteapulls` input type to the existing `gitea` test.

### DIFF
--- a/nixos-tests.nix
+++ b/nixos-tests.nix
@@ -145,10 +145,18 @@ in
             git -C /tmp/repo add .
             git config --global user.email test@localhost
             git config --global user.name test
+
+            # Create initial commit
             git -C /tmp/repo commit -m 'Initial import'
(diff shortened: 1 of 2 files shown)


ahuston-0:
Hi!

Just adding in my own experience here, I just set this branch up on my existing flake (which I migrated from github to my self-hosted gitea instance) and it worked pretty seamlessly. Not sure what would be required to get this merged in, but this would be awesome to have.

image

ahuston-0:
Hi! 

I opened a pr against fx-chun's fork (https://github.com/fx-chun/hydra/pull/1), I updated the plugin docs for GiteaPulls and added a GiteaRefs plugin based on the GithubRefs plugin.

Evidence that its working as expected:
image
image

PR #1444, opened 2025-02-14
Use `LegacySSHStore`

Ericson2314:
In https://github.com/NixOS/nix/pull/10748 it is extended with everything we need.

---------

We are no longer explicit about reusing the same connection. However:

1. Nix's `pool` mechanism will not close valid connections; it always keeps them around in case they are of use later

2. Max connections defaults to 1, so we won't open up multiple connections.

Between these two things, we are, in fact, guaranteed to reuse the same connection from one command/query to the next.

### DIFF
--- a/src/hydra-queue-runner/build-remote.cc
+++ b/src/hydra-queue-runner/build-remote.cc
@@ -9,13 +9,10 @@
 #include "path.hh"
 #include "legacy-ssh-store.hh"
 #include "serve-protocol.hh"
-#include "serve-protocol-impl.hh"
 #include "state.hh"
 #include "current-process.hh"
 #include "processes.hh"
 #include "util.hh"
-#include "serve-protocol.hh"
-#include "serve-protocol-impl.hh"
 #include "ssh.hh"
 #include "finally.hh"
 #include "url.hh"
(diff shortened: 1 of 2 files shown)


src/hydra-queue-runner/build-remote.cc
```diff
@@ -39,38 +36,6 @@ bool ::Machine::isLocalhost() const
...
-    });
-
-    // XXX: determine the actual max value we can use from /proc.
-
-    // FIXME: Should this be upstreamed into `startCommand` in Nix?
-
-    int pipesize = 1024 * 1024;
-
-    fcntl(ret->in.get(), F_SETPIPE_SZ, &pipesize);
-    fcntl(ret->out.get(), F_SETPIPE_SZ, &pipesize);
```
Ericson2314, line 68: In https://github.com/NixOS/nix/pull/10765 @vcunat determind this code was wrong. However, if the address is not too large it might make the pipe larger? Something should be done upstream.

src/hydra-queue-runner/build-remote.cc
```diff
@@ -39,38 +36,6 @@ bool ::Machine::isLocalhost() const
...
-    } else {
-        auto remoteStore = machine->storeUri.params.find("remote-store");
-        if (remoteStore != machine->storeUri.params.end()) {
-            command.push_back("--store");
-            command.push_back(shellEscape(remoteStore->second));
-        }
-    }
-
-    auto ret = master.startCommand(std::move(command), {
-        "-a", "-oBatchMode=yes", "-oConnectTimeout=60", "-oTCPKeepAlive=yes"
```
Ericson2314, line 58: These are moved below

src/hydra-queue-runner/build-remote.cc
```diff
@@ -39,38 +36,6 @@ bool ::Machine::isLocalhost() const
 
 namespace nix::build_remote {
 
-static std::unique_ptr<SSHMaster::Connection> openConnection(
-    ::Machine::ptr machine, SSHMaster & master)
-{
-    Strings command = {"nix-store", "--serve", "--write"};
-    if (machine->isLocalhost()) {
-        command.push_back("--builders");
-        command.push_back("");
```
Ericson2314, line 48: This is done below

src/hydra-queue-runner/build-remote.cc
```diff
@@ -39,38 +36,6 @@ bool ::Machine::isLocalhost() const
...
-    } else {
-        auto remoteStore = machine->storeUri.params.find("remote-store");
-        if (remoteStore != machine->storeUri.params.end()) {
-            command.push_back("--store");
-            command.push_back(shellEscape(remoteStore->second));
-        }
```
Ericson2314, line 54: This is handled by Nix.

src/hydra-queue-runner/build-remote.cc
Ericson2314: This is now done below too.

See 

```c++
remoteStore->connPipeSize = 1024 * 1024;
```

Setting a variable doesn't have any side-effects. We do that before any connection is established, however, and then that setting will be applied.

PR #1456, opened 2025-03-29
web: replace 'errormsg' with 'errormsg IS NULL' in most cases

Mic92:
This is implement in an extremely hacky way due to poor DBIx feature
support. Ideally, what we'd need is a way to tell DBIx to ignore the
errormsg column unless explicitly requested, and to automatically add a
computed 'errormsg IS NULL' column in others. Since it does not support
that, this commit instead hacks some support via method overrides while
taking care to not break anything obvious.

From the hydra.nixos.org branch

### DIFF
--- a/package.nix
+++ b/package.nix
@@ -89,6 +89,7 @@ let
         DateTime
         DBDPg
         DBDSQLite
+        DBIxClassHelpers
         DigestSHA1
         EmailMIME
         EmailSender
--- a/src/lib/Hydra/Controller/Jobset.pm
+++ b/src/lib/Hydra/Controller/Jobset.pm
@@ -371,6 +371,12 @@ sub errors_GET {
 
     $c->stash->{template} = 'eval-error.tt';
 
+    my $jobsetName = $c->stash->{params}->{name};
+    $c->stash->{jobset} = $c->stash->{project}->jobsets->find(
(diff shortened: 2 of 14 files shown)


mweinelt:
It would be nice to get these finally landed on master so we can stop carrying them in a separate branch tho.

zowoq:
> It would be nice to get these finally landed on master so we can stop carrying them in a separate branch tho.

I asked about this last month: https://github.com/nix-community/infra/pull/1349#discussion_r1955500624

Ericson2314:
Odd the constituents job test started failing

Mic92:
https://github.com/NixOS/hydra/commit/882c1526f4ca87b7519025f84c6ac81e17bbb465

mweinelt:
Which I think was fixed by 2c8202fd4b299d628ffaf49d68b0e02deda63fc3

Mic92:
I guess than I can start pulling these commits.

Ericson2314:
After #1469 lands, this will just contain one commit. That commit needs to be reworked to not conflict with @Ma27's constitution job globbing changes (that new test fails).

Ericson2314:
I think I have a fix!

Ericson2314 (approved):

PR #1469, opened 2025-04-07
queue-runner: release machine reservation while copying outputs

Ericson2314:
This allows for better builder usage when the queue runner is busy. To avoid running into uncontrollable imbalances between builder/queue runner, we only release the machine reservation after the local throttler has found a slot to start copying the outputs for that build.

As opposed to asserting uniqueness to understand resource utilization, we just switch to using `std::unique_ptr`.


Ericson2314:
CC @Mic92 I combined @delroth's original and null pointer fix, and then reworked using `std::unique_ptr` to remove asserts (now statically known) and double indirections.

PR #1478, opened 2025-04-15
jobset-eval: fix actions not showing up sometimes for new jobs

zowoq:


src/lib/Hydra/Helper/BuildDiff.pm
```diff
@@ -80,8 +85,15 @@ sub buildDiff {
         } else {
             push @{$ret->{new}}, $build if !$found;
         }
-        if (defined $build->buildstatus && $build->buildstatus != 0) {
-            push @{$ret->{failed}}, $build;
+
+        if ($build->finished != 0 && $build->buildstatus != 0) {
+            if ($aborted) {
+                ++$ret->{totalAborted};
```
Ericson2314, line 91: Woah, is that really the right precedence?

Would

```suggestion
                ++($ret->{totalAborted});
```
work for those of us that don't really know Perl?

Ericson2314: Ah OK, it's a cherry-pick, and more info is in the commit message again.

PR #1479, opened 2025-04-20
queue runner: attempt at slightly smarter scheduling criteria

zowoq:
(cherry picked from commit https://git.lix.systems/lix-project/hydra/commit/b8d03adaf45105452bf1040deeaaccc8b8b22efb)

@mweinelt Was this commit ever tried on h.n.o?


mweinelt:
I don't think it was.

Mic92 (approved):
Looks good in principle. However someone would need to take the time to watch the graphs in NixOS infra after this patch was applied. @zowoq have you tried this out by chance?

zowoq:
Yes, used this on the community hydra for a while before switching back to upstream but that probably doesn't help much as the workloads are too different.

mweinelt:
I think its worth giving it a try on hno.

PR #1482, opened 2025-04-29
Expose nix package in hydra package

Mic92:
We need this in nixos infra to keep both in sync.

### DIFF
--- a/package.nix
+++ b/package.nix
@@ -277,5 +277,8 @@ stdenv.mkDerivation (finalAttrs: {
   dontStrip = true;
 
   meta.description = "Build of Hydra on ${stdenv.system}";
-  passthru = { inherit perlDeps; };
+  passthru = {
+    inherit perlDeps;
+    nix = nixComponents.nix-cli;
+  };
 })


nixos-modules/hydra.nix
```diff
@@ -1,3 +1,4 @@
+{ self, ... }:
```
Ericson2314, line 1: Please don't pass in `self`, because then this file becomes non-extensible to new platforms

Ericson2314 (dismissed):
The `passthru` is fine, but I don't like how the module changes just infect it with flake stuff

Ericson2314:
What was wrong with how it set the default before?

nixos-modules/hydra.nix
Mic92: I dropped this commit. Because than I don't have access to pkgs anymore and need a second nixos module with the _file hack again, which kind of defeats the point the change.

Mic92:
> What was wrong with how it set the default before?

Just a minor improvement to not require two nixos modules. But not important enough to be worth the discussion.

Ericson2314 (approved):

PR #1493, opened 2025-07-04
Replace nettools with hostname-debian

mweinelt:
As far as I understand we include nettools for its hostname executable used by the Sys-Hostname-Long perl package. But if we just need that then the hostname-debian package provides a simpler and better maintained version.


SuperSandro2000:
https://github.com/NixOS/nixpkgs/pull/422364

PR #1497, opened 2025-07-16
Show queue runner v2 status

dasJ:
This is guarded behind a setting and will overwrite everything that was learned from the machines file. Also drops `sshKeys` since that wasn't used anyway.


Mic92:
Project for context: https://github.com/helsinki-systems/hydra-queue-runner

PR #1498, opened 2025-07-31
machine-status: Render new queue runner details

dasJ:
Looks like this:
image


dasJ:
Updated styling to look like this:
image

PR #1504, opened 2025-08-05
Correctly apply the setting `allow_import_from_derivation = true`

ulucs:
`ne` evaluates before `//` in Perl; which caused the setting
```
allow_import_from_derivation = true
```
to actually evaluate into not allowing import-from-derivation; while `0` became a value that did allow IFD.

This PR adds parentheses to fix the evaluation order, so `true` allows IFD while `0` does not anymore.

### DIFF
--- a/src/script/hydra-eval-jobset
+++ b/src/script/hydra-eval-jobset
@@ -382,7 +382,7 @@ sub evalJobs {
     push @cmd, "--meta";
     push @cmd, "--constituents";
     push @cmd, "--force-recurse";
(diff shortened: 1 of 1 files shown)


SuperSandro2000 (approved):

SuperSandro2000:
Oh, we should probably change the default then to avoid breaking things.

Mic92:
Ah let's wait for https://discourse.nixos.org/t/pre-disclosure-announcement-security-advisory-for-hydra-on-august-12-2025/67615

PR #1507, opened 2025-08-10
jobset-eval: reduce compare options to active jobsets

mweinelt:
The list of jobsets is very high on hydra.nixos.org and the compare to dropdown listing goes over multiple full pages in the busy projects.

If we ignore jobsets that we disable this interface becomes more usable again.

This is untested.

image


Mic92:
Ah, wait the security announcement...

mweinelt:
Rebased

mweinelt:
image

PR #1514, opened 2025-08-28
hydra-eval-jobset: disable eval cache

Mic92:


Mic92:
cc @zowoq could you test?

Ericson2314 (approved):

zowoq:
Tested both but we aren't using send-stats or any of the plugins.

PR #1517, opened 2025-08-31
feat: hydra-queue-runner-v2

Conni2461:
- [ ] integrate into meson
- [ ] integrate into testsuite


tomberek:
What does this do?

src/meson.build
```diff
@@ -75,3 +74,60 @@ install_subdir('script',
...
+    '--workspace',
+    '--message-format=json-render-diagnostics'
+  ],
+  build_by_default: true,
+  install: false,
+  depends: [],
+)
+
+rust_binaries = {
+  'queue-runner': 'hydra-queue-runner-v2',
```
dasJ: I would personally drop the -v2 suffix. It would be something we will carry with us forever, even with the old queue runner long gone

PR #1526, opened 2025-09-17
feat(performance): Stream file downloads to reduce memory usage

shift:
This patch modifies the `serveFile` subroutine in `src/lib/Hydra/Controller/Build.pm` to stream file content directly from the nix store to the client.

Previously, the entire file was read into memory, leading to high memory consumption for large build artifacts. The new implementation uses a pipe from `nix store cat` and streams the data in chunks, significantly reducing the memory footprint.

### DIFF
--- a/t/Hydra/Controller/Build/download.t
+++ b/t/Hydra/Controller/Build/download.t
@@ -61,7 +61,7 @@ subtest "Test downloading build products (regression test for #1520)" => sub {
         # The key test: should not return 500 error with "Can't use string ("1") as a HASH ref"
         isnt($response->code, 500, "Download should not return 500 error (regression test for #1520)");
         is($response->code, 200, "Download should succeed with 200")
-            or diag("Response code: " . $response->code . ", Content: " . $response->content);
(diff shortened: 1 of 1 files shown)


src/lib/Hydra/Controller/Build.pm
```diff
@@ -256,7 +253,12 @@ sub serveFile {
             $type = $info->{mime_with_encoding};
         }
         $c->response->content_type($type);
-        $c->forward('Hydra::View::Plain');
+
+        $c->res->finalize;
+        my @cmd = ("nix", "--experimental-features", "nix-command", "store", "cat", "--store", getStoreUri(), "$path");
+        my $ok = run \@cmd, '>', $c->res->output;
```
Mic92, line 259: Does this actually work?

The unit test suite says otherwise:

```
| 0E0FFC9A-93E8-11F0-8F8 | Hydra/Controller/Build | Test downloading build |
| 8-B17F2BFB4156         | /download.t            |  products (regression  |
```

PR #1537, opened 2025-11-11
[wip] hydra-eval-jobs: use nix-eval-jobs's --select rather --expr

Mic92:
the current use of builtins.getFlake, mean we don't have pure evaluation


Ericson2314 (approved):

Mic92:
Forgot that I already opened a pr for this. Here is the new one. Basically the same idea: https://github.com/NixOS/hydra/pull/1747

PR #1540, opened 2025-11-25
package.nix: update postgresql to 17

mweinelt:
NixOS 25.11 does not ship with PostgreSQL 13 any more.


mweinelt:
Merging to unblock early 25.11 deployments.

PR #1548, opened 2026-01-04
feat: Use short revision from git

### LINKED ISSUE #1245: Invalid truncation of git hashes
**Describe the bug**
hydra truncates git revisions too short, in some places. e.g. `src/root/common.tt`.

**To Reproduce**
1. Go to https://hydra.nixos.org/build/185573231
2. Take the second hash under Changes: `bba87f`
3. Run any git command involving this hash against nixpkgs
4. Observe that this hash is **not unique**

**Expected behavior**
Produce valid hashes for a given repository.

dasJ:
We actually already ask git for the short revision, so we might as well persist it to DB and render it on the frontend. To persist it, this change includes a database migraton. For older jobsetevals, we fall back on truncating the revision on 12 characters.

This logic should allow nixpkgs to scale to an infinite number of commits without any more Hydra adjustments.

Closes #1245

### DIFF
--- a/src/lib/Hydra/Schema/Result/JobsetEvalInputs.pm
+++ b/src/lib/Hydra/Schema/Result/JobsetEvalInputs.pm
@@ -66,6 +66,11 @@ __PACKAGE__->table("jobsetevalinputs");
   data_type: 'text'
   is_nullable: 1
 
+=head2 shortRevLength
+
+  data_type: 'number'
+  is_nullable: 1
+
 =head2 value
 
   data_type: 'text'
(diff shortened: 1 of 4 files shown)


dasJ:
Damn.

Mic92:
see https://github.com/NixOS/hydra/pull/1558

dasJ:
I don't think so. For each eval we store the length of a short hash and **on rendering** the build page, we take the max value for the URL over all jobsetevalinputs. So when rendering a page for an old eval, we still use the length from a later jobseteval if that resulted in a longer hash

PR #1555, opened 2026-01-05
Add option to disable search

dasJ:


src/root/topbar.tt
```diff
@@ -121,9 +121,11 @@
 [% END %]
 
 [% IF showPrivate %]
-  <form class="form-inline" action="[% c.uri_for('/search') %]">
-    <input name="query" type="text" class="form-control" placeholder="Search" [% HTML.attributes(value => c.req.params.query) %]/>
-  </form>
+  [% UNLESS c.config.disable_search %]
```
mweinelt:
```suggestion
  [% UNLESS c.config.search_enable %]
```

Wondering if going forward we want more nixos module like behaviour, where the scope comes first, and we use positive toggles (which can default either way) rather than negative ones.

PR #1556, opened 2026-01-07
treewide: update split calls to make perlcritic happy

Mindavi:
In nixpkgs this started to fail the build due to an update of, I'm guessing, perlcritic.

Note that hydra currently doesn't build in nixpkgs due to other issues with libpqxx, but while working on that I ran into this. See ticket https://github.com/NixOS/nixpkgs/issues/476278 for background on that.

### DIFF
--- a/src/lib/Hydra/Controller/User.pm
+++ b/src/lib/Hydra/Controller/User.pm
@@ -106,11 +106,11 @@ sub doEmailLogin {
     my $allowed_domains = $c->config->{allowed_domains} // ($c->config->{persona_allowed_domains} // "");
     if ($allowed_domains ne "") {
         my $email_ok = 0;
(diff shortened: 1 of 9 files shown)


mweinelt:
split is not the scope, but the function application which is changed. the scope is treewide, so

> treewide: update split calls to make perlcritic happy

Mindavi:
https://github.com/Perl-Critic/Perl-Critic/blob/c437d557ec903c99c5114bac738a484dfedb1f9f/lib/Perl/Critic/Policy/BuiltinFunctions/ProhibitStringySplit.pm#L36

Seems like the spaces have special handling/behavior. Will update without changing the spaces and the commit message too and then I expect it to be good.

PR #1565, opened 2026-01-28
hydra-queue-runner: make drv available to remote pre-build-hook

GaetanLepage:
For our use-case, [we](https://nixos-cuda.org/) need the pre-build-hook to have access to the `.drv` file of the derivation being built.
See https://github.com/NixOS/nix/issues/9272 for additional context.

Attached is a naive solution that we have tested on our infrastructure.
Please, instruct us on how to implement this properly.

cc @SomeoneSerge @YorikSar


dasJ:
cc @Conni2461

PR #1567, opened 2026-01-29
Enable gitea status in flakes

SuperSandro2000:
Rebased PR of #1232

---

Curently the GiteaStatus plugin only works for legacy inputs, as it attempts to read repository metadata from the `jobsetevalinput`s (which I don’t think can be set for flakes).

This Commit adds support for many gitea instances by checkng whether the domain is hosting gitea. This is done by querying the /api/swagger endpoint, as that is a small static page that contains “Gitea” even when the instance has been renamed.

[Gitea instances that are hosted in a
subfolder](https://docs.gitea.io/en-us/reverse-proxies/#nginx-with-a-sub-path) are not supported.


SuperSandro2000:
> Can you also add URLs that ends with `.git` in the flake URI regex?

done

PR #1568, opened 2026-01-31
Another attempt at an OIDC integration

KJTsanaktsidis:
Building on the work of @lheckemann in https://github.com/NixOS/hydra/pull/1298 & in @ners in https://github.com/ners/hydra/tree/oidc, I took a stab at trying to shape that work up in a way that can get merged.

This implementation:

* Allows you to specify multiple OIDC providers in the config file, using a syntax like

```
<oidc>
  <provider keycloak>
    display_name = "Keycloak"
    discovery_url = "http://localhost:64446/realms/hydra-dev/.well-known/openid-configuration"
    client_id = "hydra-local"
    client_secret = "hydra-local-secret"
  </provider>
  <provider other_provider>
      ...
  </provider>
</oidc>
```

* Fetches most of the configuration from the `discovery_url` (although you can specify `issuer`, `authorization_endpoint`, `token_endpoint`, and `jwks_url` manually)
* Properly validates the returned ID token against the jwks keyset fetched from jwks_url (well, I hope it's properly, anyway!)
* Looks for a `hydra_roles` claim, and, if present, sets roles on the user based on that. I did think about implementing some kind of mapping mechanism, but Keycloak, at least, seems to make it pretty easy to beat the mapping into shape on that side, so perhaps there is no need.
* I made it set the username of OIDC users to "provier:$sub", which means I had to muck around with a few templates that kind of expect to display something that visibly identifies the user (like... an email address) and the $sub that keycloak, at least, uses, is just a UUID.

I banged my head trying to write some good tests for this, but _really_ what this needs to meaningfully test things are full-system integration tests. I have a working `KeycloakContext.pm` on my machine which starts a Keycloak instance in the test framework, but I'm not really sure how to write the kind of browser test that would be needed to go through the login flow. Any suggestions?

### DIFF
--- a/subprojects/hydra-tests/lib/KanidmContext.pm
+++ b/subprojects/hydra-tests/lib/KanidmContext.pm
@@ -73,7 +73,7 @@ sub generate_config {
     open(my $fh, '>', $self->{kanidm_config}) or die "Cannot open $self->{kanidm_config}: $!";
     print $fh <<EOF;
 version = "2"
-bindaddress = "[::1]:$self->{port}"
+bindaddress = "127.0.0.1:$self->{port}"
 db_path = "$self->{kanidm_dir}/kanidm.db"
 tls_chain = "$self->{kanidm_dir}/chain.pem"
 tls_key = "$self->{kanidm_dir}/key.pem"
@@ -93,7 +93,7 @@ sub wait_for_ready {
     my ($self) = @_;
 
     my $logfile = $self->{_logfile};
-    my $max_attempts = 60;
+    my $max_attempts = $ENV{KANIDM_STARTUP_TIMEOUT} // 60;
 
     for (my $i = 0; $i < $max_attempts; $i++) {
         eval { $self->assert_running(); };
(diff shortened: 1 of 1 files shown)


stepbrobd:
for context, my hydra instance with @ners patch use kanidm as well and it looks like this 

https://github.com/stepbrobd/dotfiles/blob/5f859debb62e710950780980047e279f4d3d778d/hosts/server/odake/hydra.nix#L53

KJTsanaktsidis:
Hm, OK, so the reason kanidm can't literally just put the inside-hydra role names in a custom claim is because the custom claim values have to match the regex `^[0-9a-zA-Z_]+$`: https://github.com/kanidm/kanidm/blob/cbd9c5135e08671d642284758ad977dac5391ac1/server/lib/src/value.rs#L1947

Most of the hydra roles (all except `admin`) have a `-` in them.

This limitation is probably a bit too strict on the kanidm side IMO - it's been discussed a few times https://github.com/kanidm/kanidm/issues/2641, https://github.com/kanidm/kanidm/issues/2882. I'll go add my two-cents worth into there. And in the meanwhile, I guess, make the hydra side accept e.g. `bump_to_front` as an alias for the `bump-to-front` role perhaps.

Mic92:
@KJTsanaktsidis sorry, just see this just now. I will put it on my list.

Mic92:
I pushed a stack of `fixup!` commits addressing issues found during review. Kept separate for easier review; once you're happy I'll autosquash.

   - 780d45f8 Open redirect: `?after=/evil.com` became `//evil.com` (protocol-relative). Now strips leading `/` and `\`.
   - b40f60c5 state/nonce comparison now uses `String::Compare::ConstantTime`.
   - 913409f9 `GET /logout` now requires a CSRF token (`sha256(sessionid + ":logout")`), embedded in the Sign out link.
   - f5ad8fa7 Token endpoint: reject HTTP errors even when JSON lacks `error` field; validate `aud` claim exists.
   - 2d69ea17 `setRoles`: wrap INSERT+DELETE in a transaction.
   - 56279811 OIDC discovery failure at startup now logs a warning instead of preventing Hydra from booting.
   - b43f6245 Moved test-only Perl modules (`TestWWWMechanize`, `HTTPCookieJar`, …) from hydra to hydra-tests.
   - Commit 6e25009b `config-fixup` needs manual squash into "Add mapping of OIDC claims to roles" (not a proper `fixup!`).
   - bc6aa4d2 RP-Initiated Logout: redirect to IdP's `end_session_endpoint` on sign out. Adapted from https://git.lix.systems/lix-project/hydra/pulls/73, credited to Maximilian Bosch.

PR #1581, opened 2026-03-18
nix-utils: remove `FfiRealisation` wrapper, return `Realisation` directly from query

amaanq:
Callers went through `FfiRealisation.as_rust()` to deserialize the JSON
into a `Realisation` every time. Inlining that into the trait method
removes the wrapper struct, the unsafe Send/Sync impls, and the extra
conversion step at each call site.


Ericson2314:
No longer needed --- FFI is gone.

PR #1620, opened 2026-03-30
Don't pass resolved_drv to the builder

artemist:
Whenever possible, only a resolved derivation should be passed around. Remove all the separation between "resolved" and "unresolved" derivation past `Job`, including in the builder.


Ericson2314:
Superceded by https://github.com/NixOS/hydra/pull/1629, I believe.

PR #1625, opened 2026-03-30
fix: Remove println that exposed secrets

Erethon:
These two `println`s look like debugging output that was left over. The second one is harmless in the test, but the first one can leak AWS secrets in logs if I understand it correctly.


Conni2461:
we should merge this, as is. we dont really need this. I forgot to remove this, after implementing this, mybad

PR #1629, opened 2026-03-30
queue-runner: resolve CA derivations after dependency outputs land

amaanq:
Instead of resolving at dispatch time and carrying two drv identities through the gRPC layer, resolve in `succeed_step` once outputs are in the DB. The resolved drv becomes a fresh `Step` that dispatch picks up normally; the builder only ever sees one drv. A new `resolvedFromBuild`/`resolvedFromStep` foreign key `BuildSteps` ties the resolved step back to the dependency that triggered it.

### DIFF
--- a/subprojects/crates/db/src/models.rs
+++ b/subprojects/crates/db/src/models.rs
@@ -19,7 +19,8 @@ pub enum BuildStatus {
     LogLimitExceeded = 10,
     NarSizeLimitExceeded = 11,
     NotDeterministic = 12,
-    Busy = 100, // not stored
+    Resolved = 13, // step was resolved to a CA derivation, see resolvedTo FK
+    Busy = 100,    // not stored
 }
 
 impl BuildStatus {
@@ -38,6 +39,7 @@ impl BuildStatus {
             10 => Some(Self::LogLimitExceeded),
             11 => Some(Self::NarSizeLimitExceeded),
             12 => Some(Self::NotDeterministic),
+            13 => Some(Self::Resolved),
             100 => Some(Self::Busy),
             _ => None,
         }
--- a/subprojects/crates/db/src/models.rs
+++ b/subprojects/crates/db/src/models.rs
@@ -22,6 +22,8 @@ pub enum BuildStatus {
     LogLimitExceeded = 10,
     NarSizeLimitExceeded = 11,
     NotDeterministic = 12,
+    /// step was resolved to a CA derivation
+    Resolved = 13,
     /// not stored
     Busy = 100,
 }
@@ -42,6 +44,7 @@ impl BuildStatus {
             10 => Some(Self::LogLimitExceeded),
             11 => Some(Self::NarSizeLimitExceeded),
             12 => Some(Self::NotDeterministic),
+            13 => Some(Self::Resolved),
             100 => Some(Self::Busy),
             _ => None,
         }
--- a/subprojects/hydra-queue-runner/src/state/mod.rs
+++ b/subprojects/hydra-queue-runner/src/state/mod.rs
@@ -10,6 +10,7 @@ mod step;
 mod step_info;
 mod uploader;
 
+use anyhow::Context as _;
 pub use atomic::AtomicDateTime;
 pub use build::{Build, BuildOutput, BuildResultState, BuildTimings, Builds, RemoteBuild};
 pub use jobset::{Jobset, JobsetID, Jobsets};
@@ -18,6 +19,7 @@ pub use queue::{BuildQueueStats, Queues};
 pub use step::{Step, Steps};
 pub use step_info::StepInfo;
 
+use std::collections::BTreeMap;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicI64, Ordering};
 use std::time::Instant;
@@ -46,6 +48,8 @@ enum CreateStepResult {
 
 enum RealiseStepResult {
     None,
+    /// Created a new resolved `BuildStep`
+    Resolved,
     Valid(Arc<Machine>),
     MaybeCancelled,
     CachedFailure,
@@ -275,7 +279,7 @@ impl State {
         let drv = step_info.step.get_drv_path();
         let mut build_options = nix_utils::BuildOptions::new(None);
 
-        let build_id = {
+        let build = {
             let mut dependents = HashSet::new();
             let mut steps = HashSet::new();
             step_info.step.get_dependents(&mut dependents, &mut steps);
@@ -308,14 +312,12 @@ impl State {
             build_options
                 .set_max_silent_time(biggest_max_silent_time.unwrap_or(build.max_silent_time));
             build_options.set_build_timeout(biggest_build_timeout.unwrap_or(build.timeout));
-            build.id
+            build.clone()
         };
 
-        let mut job = machine::Job::new(
-            build_id,
-            drv.to_owned(),
-            step_info.resolved_drv_path.clone(),
-        );
+        let build_id = build.id;
+
+        let mut job = machine::Job::new(build_id, drv.to_owned());
         job.result.set_start_time_now();
         if self.check_cached_failure(step_info.step.clone()).await {
             job.result.step_status = BuildStatus::CachedFailure;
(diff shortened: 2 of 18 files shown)


subprojects/hydra/sql/hydra.sql
```diff
@@ -293,9 +293,15 @@ create table BuildSteps (
     -- Whether this build step produced different results when repeated.
     isNonDeterministic boolean,
 
+    -- If this step was created by resolving a CA derivation, points to the
+    -- original unresolved step that triggered resolution.
+    resolvedFromBuild integer,
+    resolvedFromStep  integer,
```
Ericson2314: Has to be "to" direction not "from" direction, because many derivations will resolve to the same derivation

Ericson2314 (changes requested):
At the point where we `try_resolve` before, we do this:

- If result is `None` or "same derivation":
  
  just continue with this step, same as we always did

- If result is `Some(drv)` and that `drv` is not the same as the current drv:

  Step finished, status "resolved", fill in the DB pointer to the new step, and the new step now builds, this step is done.

-----

the current "try resolve reverse dependencies" is a possible optimisation --- incrementally resolve reverse dependencies, reminiscent of union find --- but that's a more complex thing we shouldn't worry about at this time.

artemist:
Rebased onto #1642

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -308,14 +311,12 @@ impl State {
             build_options
                 .set_max_silent_time(biggest_max_silent_time.unwrap_or(build.max_silent_time));
             build_options.set_build_timeout(biggest_build_timeout.unwrap_or(build.timeout));
-            build.id
+            build.clone()
```
Ericson2314, line 326: Could this be a move?

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -357,33 +358,175 @@ impl State {
...
+                        }
+                    )
+                });
+
+                if is_ca && needs_resolution && resolved.is_none() {
+                    return Err(anyhow::anyhow!(
+                        "Failed to resolve CAFloating derivation {drv}"
+                    ));
+                }
+
```
Ericson2314: We should probably do this stuff first, before we bother to compute the resolution?

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -357,33 +358,175 @@ impl State {
...
+                resolved
+            } else {
+                None
+            }
+        };
+
+        if let Some(resolved_path) = resolved {
```
Ericson2314: I think we can combine this with the previous branch? Or at least we could compute a "do we want to resolve" `bool`, and then have the actual resolving (which much succeed) and this stuff in the same branch.

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -357,33 +358,175 @@ impl State {
...
+            tracing::info!("resolved CA derivation {drv} -> {resolved_path}");
+
+            // Finish original step as "resolved".
```
Ericson2314: If possible, I would like to create the follow-up build step before we marked this resolved. This will bode well for the future when we start doing more of the state machine in the database.

Ideally, we will have a check constraint on the resolved status, saying it can only be used when the foreign key to the next step is set!

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -357,33 +358,175 @@ impl State {
...
+            step_info.step.set_finished(true);
+            let mut resolved_result = RemoteBuild::new();
+            resolved_result.step_status = BuildStatus::Resolved;
+            resolved_result.set_start_time_now();
+            resolved_result.set_stop_time_now();
+            resolved_result.log_file.clone_from(&job.result.log_file);
+            finish_build_step(
```
Ericson2314: Let's try to do everything in one transaction here. `finish_build_step` current creates its own transaction, so let's make take the transaction as an argument instead.

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -357,33 +358,175 @@ impl State {
...
+                    BuildStatus::Busy,
+                    None,
+                    None,
+                    vec![],
+                )
+                .await?;
+
+            tx.set_resolved_to(build_id, step_nr, nr).await?;
+            tx.commit().await?;
+
```
Ericson2314, line 506: TODO think about restart resiliency. I believe everything else after this point in this branch doesn't use the database. So that's good, we aren't logically splitting a transaction in two (assuming the previous stuff is done).

When is left to think about is ---- if the queue runner was restarted at this point, would the general "rebuild state machine from build database state" logic get us into the same step as what this code right here does?

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -357,33 +358,175 @@ impl State {
...
+            // Actually schedule the step
+            let resolved_step = match self
+                .create_step(
+                    build.clone(),
+                    resolved_path,
+                    Some(build.clone()),
+                    None,
+                    Arc::new(parking_lot::RwLock::new(HashSet::new())),
+                    Arc::new(parking_lot::RwLock::new(HashSet::new())),
+                    Arc::new(parking_lot::RwLock::new(HashSet::new())),
```
Ericson2314: We should have a big fat comment why it is correct to not give this function the mutable state it expects

subprojects/hydra-queue-runner/src/state/mod.rs
Ericson2314: We could make a variation of our CA tests where where the queue runner dies here and then restarts to test this

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -357,33 +358,175 @@ impl State {
...
+
+            // Copy rdeps so dependents don't try to start too early
+            for rdep in step_info.step.clone_rdeps() {
+                if let Some(rdep) = rdep.upgrade() {
+                    resolved_step.make_rdep(&rdep);
+                }
+            }
+
+            // If we're the root of a build then we need to become the new root,
+            // lest the Step get garbage collected.
```
Ericson2314: Comment should be clear this is about `Arc` keeping things alive, not store keeping `,drv` files alive.

We should also be clear that in the non-root case, the idea is that we registered some reverse des above that will keep us alive. Maybe we should assert that we did in fact go through that loop.

Ericson2314: We will need to figure out how to keep the `.drv` file alive for *all* resolved derivations, and the GC roots that `nix-eval-jobs` made will not do the job since these are new drvs.

subprojects/hydra/sql/hydra.sql
```diff
@@ -293,9 +293,14 @@ create table BuildSteps (
...
 
+    -- If this step's outputs triggered CA resolution of a dependent, points
+    -- to the resolved step that was created.
+    resolvedToStep   integer,
+
     primary key   (build, stepnr),
     foreign key   (build) references Builds(id) on delete cascade,
-    foreign key   (propagatedFrom) references Builds(id) on delete cascade
+    foreign key   (propagatedFrom) references Builds(id) on delete cascade,
+    foreign key   (build, resolvedToStep) references BuildSteps(build, stepnr) on delete cascade
```
Ericson2314: So long term, when the database distinguishes *derivations* from *build attempts*, this should actually point to a derivation, not another build attempt.

A build attempt samples the state of the database at that point to resolve, and creates a new (resolved) derivation. That derivation can be attempted to be built multiple times (We don't really care to privilage the first attempt with a foreign key). However, subsequent attempts to build the original derivation --- if we rebuild other upstream derivations and they are content-addressed and non-deterministic --- will result in a *different* resolved derivations, and that too can be attempted to build multiple times.

Memoizing realization in this way does give us something close a derived build trace, but the fact that our build trace is thus keyed off attempts not not derivations gives us some ability to represent (and not choke on) incoherence.

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: Not trivially, although it could be replaced with `dependents.take(build).unwrap()` or similar to avoid a clone

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: This is doable, though it will take out a lock on the step's derivation for longer than strictly necessary.

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: Comment should be done

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: The ordering change is done

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: Done

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: Unfortunately it looks like hydra-queue-runner is making duplicate `BuildStep` rows in the database on restart. I'm unsure if this also occurs for non-CA derivations, but it should be fixed. Restart currently doesn't fail, but it leaves the DB in a not very clean state.

artemist: Oh, that may be an issue with CA builds, since in `create_step` we assume that all CAFloating derivations are unfinished, instead of checking if the outputs exist.

It is still true that the queue runner does not attempt to load steps from the database on restart, it recalculates them.

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: Comment made

subprojects/crates/db/src/models.rs
```diff
@@ -17,7 +17,8 @@ pub enum BuildStatus {
     LogLimitExceeded = 10,
     NarSizeLimitExceeded = 11,
     NotDeterministic = 12,
-    Busy = 100, // not stored
+    Resolved = 13, // step was resolved to a CA derivation, see resolvedTo FK
```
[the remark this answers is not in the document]
Ericson2314: Nevermind, this was always a separate enum.

subprojects/hydra-queue-runner/src/state/mod.rs
artemist: That doesn't seem to be easily possible, since my `take` proposal doesn't work with the borrow checker.

Ericson2314: ah ok, way below

Ericson2314:
Unfortunately, once https://github.com/NixOS/hydra/pull/1667 is merged in, this PR will hang.

subprojects/hydra-queue-runner/src/state/mod.rs
Ericson2314: Now none of this stuff writes to the database, should be just fine

Ericson2314 (approved):
I realized we can temporarily do this without a DB changes, allowing it to go in right away.

(Without a DB migration, and without `hydra.nixos.org` enabling CA derivations, it is functionally a no-op there.)

PR #1630, opened 2026-03-31
treefmt: add nightly rustfmt via fenix with formatting config

amaanq:
### Problem

Rustfmt OOTB doesn't format a lot of stuff that'd be nice to have formatted. As an example, having comments automatically wrap would be nice.

### Solution

This commit adds a rustfmt formatting config with some nightly options enabled that greatly improve the formatter. Some of the highlights include wrapping comments & strings, organizing imports, and formatting imports to have 1 import per line (which really helps with diffs). Due to all these formatting goodies, the diff is quite massive but nothing functionally changed.


Conni2461:
i like that rust has a builtin formatter and i dont have to discuss which is the "correct" formatting, so i'd prefer if we not pull in the unstable rustfmt (we build with a stable rustc) with a bunch of custom config and just use the default config that rustfmt

Ericson2314:
Fair enough

PR #1642, opened 2026-04-04
Do not share store between builder and queue runner in unit tests

Ericson2314:


subprojects/crates/nix-utils/src/drv.rs
```diff
@@ -34,13 +18,14 @@ pub async fn query_drv(
...
-    if !fs_err::tokio::try_exists(&full_path).await? {
+    if !store.is_valid_path(drv).await {
         return Ok(None);
     }
 
-    let input = fs_err::tokio::read_to_string(&full_path).await?;
-    Ok(Some(parse_drv(store.get_store_dir(), drv, &input)?))
+    let json = store.read_derivation_json(drv).await?;
+    let drv = serde_json::from_str::<Derivation>(&json)
+        .map_err(|e| anyhow::anyhow!("failed to parse derivation JSON: {e}"))?;
```
artemist: This could be `.context("failed to parse derivation JSON")`, depending on what you want the output to look like.

PR #1648, opened 2026-04-07
hydra-tests: stream queue-runner/builder logs while polling

Ericson2314:
Pump the queue-runner `IPC::Run` harness alongside the builder and flush accumulated stderr on every poll iteration, so when a test hangs we actually see both processes' logs in real time instead of only getting them after the harness is killed.


artemist (approved):
Output is a bit weird, but that's yath's fault. It seems useful for debugging to me.

Ericson2314:
I will merge this because this is just test debugging code -- it doesn't really affect when the tests pass (unless we forgot to stream and pipes started filling up) and it *certainly* doesn't affect Hydra in production.

PR #1671, opened 2026-04-15
fix: re-check step outputs on re-encounter to unstick cached builds

angerman:
## Summary

Fixes an infinite re-load loop where builds whose outputs became available
between poll cycles (via concurrent builds, substitution, or external upload)
get stuck forever.

When a step is encountered that already exists (`is_new = false`), the code
previously just returned it as-is. If the step's outputs had become available
in the store since it was first created, it would be dispatched, the builder
would find nothing to do, and the cycle would repeat indefinitely.

## Fix

Add two checks when `is_new = false`:
1. If the step is already marked finished, return `None` (skip it)
2. Query the store for missing outputs — if all are present, mark the step
   finished and return `None`

## Origin

Originally developed and deployed in `zw3rk/hydra-queue-runner` (commit
`67ef4845`), adapted here to the upstream harmonia `StorePath` types.

## Test plan

- [ ] Verify that builds whose outputs appear in the store between poll
  cycles no longer get stuck in an infinite re-dispatch loop
- [ ] Verify normal build dispatch is unaffected

### DIFF
--- a/subprojects/hydra-queue-runner/src/state/mod.rs
+++ b/subprojects/hydra-queue-runner/src/state/mod.rs
@@ -1761,6 +1761,24 @@ impl State {
             self.steps
                 .create(&drv_path, referring_build.as_ref(), referring_step.as_ref());
         if !is_new {
+            // Re-check whether the step's outputs have appeared in the store
+            // since it was first created. This handles the case where outputs
+            // became available between poll cycles (e.g. built by a concurrent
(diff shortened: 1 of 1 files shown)


Ericson2314:
I talked to @angerman elsewhere, and it came up this is probably happening because of IFD — regularly scheduled builds do not have this race condition.

I'm fine merging this for now then. Long term we should replace this with interposing the daemon socket so IFD builds can also be distributed.

PR #1673, opened 2026-04-17
fix(darwin): align builder module CLI flag with binary

### LINKED ISSUE #1672: Builder module passes --tmp-avail-threshold but binary expects --build-dir-avail-threshold
## Summary

The darwin (and linux) builder NixOS/nix-darwin modules pass `--tmp-avail-threshold` to the `hydra-builder` binary, but the binary's clap CLI expects `--build-dir-avail-threshold`. This causes all builders to crash-loop on startup:

```
error: unexpected argument '--tmp-avail-threshold' found

  tip: a similar argument exists: '--store-avail-threshold'

Usage: hydra-builder --gateway-endpoint <GATEWAY_ENDPOINT> --ping-interval <PING_INTERVAL> --speed-factor <SPEED_FACTOR> --max-jobs <MAX_JOBS> --store-avail-threshold <STORE_AVAIL_THRESHOLD>
```

## Root cause

The mismatch has existed since the original `feat: hydra-queue-runner` commit (7d9f5364, 2025-08-31). Both files were introduced in that commit but were never aligned:

- `src/builder/src/config.rs` (now `subprojects/hydra-builder/src/config.rs`): defines the field as `build_dir_avail_threshold`, which clap maps to `--build-dir-avail-threshold`
- `nixos-modules/darwin-builder-module.nix` (now `darwin-modules/builder-module.nix`): defines the option as `tmpAvailThreshold` and passes `--tmp-avail-threshold`

The same mismatch exists in `nixos-modules/builder-module.nix`.

## Impact

All gRPC builders fail to start, leaving the queue runner with zero connected machines. All builds are aborted with "Unsupported system type".

## Fix

Either:
- Rename the Nix module option and CLI string to `--build-dir-avail-threshold` (preferred — matches what the binary actually checks: `/nix/var/nix/builds`)
- Or rename the Rust field to `tmp_avail_threshold` to match the module

angerman:
## Summary

- The darwin builder module (`darwin-modules/builder-module.nix`) passed `--tmp-avail-threshold` to the `hydra-builder` binary, but the binary expects `--build-dir-avail-threshold`
- The linux module (`nixos-modules/builder-module.nix`) already uses the correct option name `buildDirAvailThreshold` and flag `--build-dir-avail-threshold`
- This aligns the darwin module to match

The mismatch has existed since the original `feat: hydra-queue-runner` commit (7d9f5364). The linux module was fixed at some point, but the darwin module was never updated.

## Impact

All darwin gRPC builders crash-loop on startup, leaving the queue runner with zero connected darwin machines and all darwin builds aborted with "Unsupported system type".

Fixes #1672

## Test plan

- [x] Deploy darwin builders with updated module — verify `hydra-builder` starts and connects to queue runner
- [x] Confirm `--build-dir-avail-threshold` appears in process args instead of `--tmp-avail-threshold`

### DIFF
--- a/darwin-modules/builder-module.nix
+++ b/darwin-modules/builder-module.nix
@@ -39,8 +39,8 @@ in
         default = 4;
       };
 
-      tmpAvailThreshold = lib.mkOption {
-        description = "Threshold in percent for /tmp before jobs are no longer scheduled on the machine";
+      buildDirAvailThreshold = lib.mkOption {
+        description = "Threshold in percent for nix build dir before jobs are no longer scheduled on the machine";
         type = lib.types.float;
         default = 10.0;
       };
@@ -160,8 +160,8 @@ in
               cfg.speedFactor
               "--max-jobs"
               cfg.maxJobs
-              "--tmp-avail-threshold"
-              cfg.tmpAvailThreshold
+              "--build-dir-avail-threshold"
+              cfg.buildDirAvailThreshold
               "--store-avail-threshold"
               cfg.storeAvailThreshold
               "--load1-threshold"
--- a/darwin-modules/builder-module.nix
+++ b/darwin-modules/builder-module.nix
@@ -211,6 +211,10 @@ in
       environment = {
         RUST_BACKTRACE = "1";
         NIX_SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
+        # The builder execs `nix build` to realise derivations.  The linux
+        # module sets `path = [ config.nix.package ]` on the systemd service;
+        # launchd has no equivalent, so we set PATH explicitly.
+        PATH = "${config.nix.package}/bin:/usr/bin:/bin:/usr/sbin:/sbin";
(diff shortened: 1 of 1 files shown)


darwin-modules/builder-module.nix
```diff
@@ -211,6 +211,10 @@ in
       environment = {
         RUST_BACKTRACE = "1";
         NIX_SSL_CERT_FILE = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
+        # The builder execs `nix build` to realise derivations.  The linux
+        # module sets `path = [ config.nix.package ]` on the systemd service;
+        # launchd has no equivalent, so we set PATH explicitly.
+        PATH = "${config.nix.package}/bin:/usr/bin:/bin:/usr/sbin:/sbin";
```
Ericson2314, line 217: @mweinelt @Mic92 do you all have opinions on this part? I do not have much `nix-darwin` experience so I don't know what the right thing to do is.

The rest looks good to me, though.

PR #1680, opened 2026-04-24
Port non-trivial dynamic derivation test from Nix, fix builder upload closure

Ericson2314:
Port `tests/functional/dyn-drv/non-trivial.nix` from the Nix test suite as a Hydra integration test. This exercises recursive-nix: a single derivation dynamically creates a DAG of inner derivations (a-e) via `nix derivation add`, outputs the final `.drv`, and a wrapper depends on building that `.drv` through `builtins.outputOf`.

Porting this test uncovered a bug in `upload_nars_regular`: it only uploaded the direct build outputs without computing their closure first. When the output is a `.drv` file whose references include other dynamically-created store paths (from recursive-nix), the queue runner's import would fail because those referenced paths were never uploaded. `upload_nars_presigned` already handled this correctly by calling `query_requisites` before uploading — this commit brings the regular path in line.

Supporting changes:
- Add `nixBinDir` to test `config.nix.in` so job expressions can find the `nix` CLI for recursive-nix builds
- Generalize `replace_variable_in_file` to accept multiple substitutions in a single pass (needed because it reads from `.in` each time)
- Add `extra-system-features = recursive-nix` to the test's nix.conf so the builder advertises the feature for scheduling

### DIFF
--- a/subprojects/hydra-builder/src/state.rs
+++ b/subprojects/hydra-builder/src/state.rs
@@ -879,6 +879,14 @@ async fn upload_nars_regular(
     metrics: Arc<crate::metrics::Metrics>,
     nars: Vec<nix_utils::StorePath>,
 ) -> anyhow::Result<()> {
+    // Compute the full closure of output paths so that all referenced
+    // store paths (e.g. dynamically-created derivations from recursive-nix)
+    // are uploaded alongside the direct outputs.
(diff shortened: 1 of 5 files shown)


Ericson2314 (commented):
Most of this is just tests, so I am happy to self-merge.

PR #1682, opened 2026-04-24
web: navigate from unresolved to resolved CA build steps

amaanq:
>[!NOTE]
> This depends on #1629

Resolved CA steps do not own the log users are trying to inspect, so
leaving them as terminal pages makes the build view look stuck or failed
in the wrong place. The log page now follows the recorded resolution chain
to the step that actually performed the work, while preserving a pending
page for unresolved, running, or malformed chains.

The redirect banner is reconstructed from BuildSteps instead of session
flash so copied log URLs and refreshes keep the same context. The JSON
summary exposes the same durable resolution data for API clients that
need to follow CA build steps without scraping HTML.

### DIFF
--- a/subprojects/hydra-tests/Hydra/Controller/Build/resolved.t
+++ b/subprojects/hydra-tests/Hydra/Controller/Build/resolved.t
@@ -1,0 +1,324 @@
+use strict;
+use warnings;
+use Setup;
+use File::Basename ();
+use JSON::MaybeXS qw(decode_json);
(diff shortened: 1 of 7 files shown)


Ericson2314:
1. `nixlog` URL is for any type of step, not just build steps. Resolve steps should have pages too.
2. "resolve of" should be "resolution of"

Ericson2314 (approved):

PR #1678, opened 2026-04-21
queue-runner-module: fix tokenListPath type

mweinelt:
```
error: A definition for option `services.hydra-queue-runner-dev.settings.tokenListPath' is not of type `null or absolute path'. Definition values:
       - In `/nix/store/i7mgnwhv201ynilyf4v7nlbb3haxxm00-source/build/hydra-queue-runner.nix':
           [
             "/run/agenix/elated-minsky-queue-runner-token"
             "/run/agenix/goofy-hopcroft-queue-runner-token"
             "/run/agenix/hopeful-rivest-queue-runner-token"
             "/run/agenix/sleepy-brown-queue-runner-token"
           ...
```

Was eventually a list in the original repo and is now a nullable path again.

PR #1684, opened 2026-04-25
feat: support passing multiple token files to queue-runner

mweinelt:
Ports https://github.com/helsinki-systems/hydra-queue-runner/commit/5169e77b0aabede15fedf41ac41d1c55627ef2b8 and renames the option to match.

Follow-up to #1678 which built on the assumption that the full queue-runner landed in this repo, but apparently it did not.


mweinelt:
Tests succeed locally with limited number of cores. :melting_face:

PR #1685, opened 2026-04-25
Create a proper build step details page

Ericson2314:
Please review the commits in order, and see each commit message for details.

### DIFF
--- a/subprojects/hydra/lib/Hydra/Controller/Build.pm
+++ b/subprojects/hydra/lib/Hydra/Controller/Build.pm
@@ -6,6 +6,7 @@ use warnings;
 use base 'Hydra::Base::Controller::NixChannel';
 use Hydra::Helper::Nix;
 use Hydra::Helper::CatalystUtils;
+use Hydra::Helper::LogEndpoints;
 use File::Basename;
 use File::LibMagic;
 use File::stat;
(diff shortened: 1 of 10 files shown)


Ericson2314:
@mweinelt While the last commit adds a redirect from `nixlog` so users get no breaking change, I suspect a deployment config change might be needed to change a `nixlog/n/raw` -> S3 redirect to a  `step/n/log/raw` S3 redirect?

In other words, while there isn't a user-facing breaking change, I think there is a sysadmin-facing breaking change.

mweinelt:
I don't believe we currently do any redirects to S3 ourselves.

https://github.com/NixOS/infra/blob/main/build/hydra-proxy.nix#L68-L102

Ericson2314:
OK Thanks @mweinelt. I guess, we will see!

subprojects/hydra/lib/Hydra/Controller/Build.pm
```diff
@@ -133,17 +134,13 @@ sub constituents_GET {
...
-    notFound($c, "Build doesn't have a build step $stepnr.") if !defined $step;
-
-    $c->stash->{step} = $step;
-
-    my $drvPath = $step->drvpath;
-    my $log_uri = $c->uri_for($c->controller('Root')->action_for("log"), [WWW::Form::UrlEncoded::PP::url_encode(basename($drvPath))]);
-    showLog($c, $mode, $log_uri);
+    my @path = ('/build', $c->stash->{id}, 'step', $stepnr, 'log');
+    push @path, $mode if defined $mode;
+    $c->res->redirect($c->uri_for(@path), 301);
```
Conni2461: are we missing `$c->req->query_params` here? there could be `tail=` set and this would get lost

Conni2461 (approved):

PR #1690, opened 2026-04-28
Add migration testing infrastructure

Ericson2314:
Reorganize `sql/` into `sql/migrations/` and `sql/schemas/`, and add per-migration Nix checks that verify each `upgrade-N.sql` produces a schema matching the historical snapshot.

- Move `upgrade-*.sql` files into `sql/migrations/`
- Extract historical `hydra.sql` snapshots from git into `sql/schemas/` with `commit-N.txt` provenance files
- Add `check-migrations.nix` which generates one derivation per migration step, using `pg-schema-diff` for semantic schema comparison and `cpp` to preprocess legacy `#ifdef SQLITE`/`#ifdef POSTGRESQL` directives
- Add `schemas/verify.sh` to verify snapshots match their recorded commits
- Update `hydra-init`, `meson.build`, `package.nix`, and `flake.nix` for the new directory layout
- Fix two historical schema snapshots where `hydra.sql` was out of sync with its corresponding migration (versions 2 and 41)
- Work around PG 17 FK constraint naming change in `upgrade-67` test


Ericson2314:
An alternative to vendoring the old schemas is doing a self-flake fetch.

Ericson2314:
I will just merge this for now, and we can explore the alternative design later. (It's not like these newly-added files are bloating the repo, because they are being pulled from history --- it already has these blobs.)

PR #1691, opened 2026-04-29
fix(queue-runner): single threaded zstd compression in tonic

Conni2461:
Tonic uses single threaded zstd compression, see
https://github.com/hyperium/tonic/blob/368361b54c1f09bf652d87b11833860b016be872/tonic/src/codec/compression.rs#L203-L295. To resolve this issue we no longer use compression with tonic and just directly encode/decode the log stream as well as the streaming of nars using async-compression.

I am not 100% sure that this solves the poor single core performance but it might explain it. I'm currently testing this locally, by streaming nars in a second nix store

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1453,9 +1453,11 @@ name = "hydra-builder"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "async-compression",
  "async-stream",
  "backon",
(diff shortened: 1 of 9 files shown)


Ericson2314 (approved):
I just glanced at this, but idea sounds good. Hit merge whenever you want @Conni2461

PR #1692, opened 2026-04-30
`hydra-ad-hoc`: Hydra as one giant nix daemon (new, optional, experimental)

amaanq:
See each commit for details. Most of the prep work has already been merged. We just have:

1. Fix a bug where `BuildOutputs` didn't get filled in in some cases
2. Split out some code for live logs to be shared with `hydra-wss`
3. Implement the new feature itself.

The new service is experimental, optional, and off by default.


subprojects/hydra-tests/drv-daemon/adhoc.t
Ericson2314, line 1: This should not be called this, dynamic derivations and import-from-derivation are completely separate features.

We want one test for "ad hoc jobs" and one test for a regular job (like with the project/jobset stuff) but using import from derivation.

Ericson2314: Also this test isn't using content-addressing (on purpose, because IFD is not experimental)

subprojects/hydra-tests/content-addressed/dyn-drv-imperative.t
```diff
@@ -0,0 +1,83 @@
...
+use Hydra::Helper::Exec;
+
+# Exercises nix-store --realise through drv-daemon into an ad-hoc Hydra build.
+
+my $ctx = test_context(
+    nix_config => qq|
+    experimental-features = ca-derivations dynamic-derivations
+    |,
+);
+
```
Ericson2314:
```suggestion
```

subprojects/hydra-tests/lib/DrvDaemonContext.pm
```diff
@@ -0,0 +1,345 @@
...
+            NIX_STATE_DIR => $ctx->{builder}{nix_state_dir},
+            NIX_STORE_DIR => $ctx->{builder}{nix_store_dir},
+            RUST_LOG      => "hydra_builder=debug,info",
+        },
+    );
+
+    _wait_for($self->{ua}, "$self->{base_url}/status/machines", sub {
+        shift->decoded_content =~ /"hostname"/;
+    }) or die "Timed out waiting for builder to register\n";
+}
```
Ericson2314: Very sus! we already have code to start the builder, we should not need new such code.

subprojects/hydra-tests/lib/DrvDaemonContext.pm
```diff
@@ -0,0 +1,345 @@
...
+            LocalPort => $port,
+            Proto     => 'tcp',
+            ReuseAddr => 0,
+        );
+        if ($sock) {
+            close($sock);
+            return $port;
+        }
+    }
+}
```
Ericson2314: Should be shared with other htings

Ericson2314:
I think it would be best to be able to run the *entire test suite* (or at least the parts doing evaluation) with and without the new drv daemon.

That means that the new IFD and ad hoc jobs tests would require the new daemon, but every other test would run twice in both modes.

subprojects/hydra/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -278,7 +300,16 @@ struct Evaluator
...
+        // unix:// ignores NIX_STORE_DIR, so NIX_REMOTE must include the logical store dir.
+        auto drvDaemonSocket = config->getStrOption("drv_daemon_socket");
+        auto nixStoreDir = config->getStrOption("nix_store_dir", "/nix/store");
+
         jobset.pid = startProcess([&]() {
+            if (!drvDaemonSocket.empty()) {
+                auto remote = "unix://" + urlEncodePath(drvDaemonSocket)
+                    + "?store=" + urlEncodePath(nixStoreDir);
+                setenv("NIX_REMOTE", remote.c_str(), 1);
+            }
```
Ericson2314: I am not sure, but this feels like the wrong layer of abstraction. I forgot that `hydra-evaluator` was its own separate demon, not one of the "scripts". 
What that means is that:

1. If the setting stays in the config, it should be holistic --- do we want evaluations to use the drv daemon or not? That is the actual policy choice here.
2. If we instead mess around with the underlying mechanisms, we could just set `NIX_REMOTE` for the entire hydra-evaluator *from the outside*, not within it. I.e. it is the responsibility of however the daemons set up and run to wire things up correctly.

subprojects/hydra/hydra-evaluator/hydra-evaluator.cc
```diff
@@ -164,6 +166,26 @@ enum class EvaluationStyle
...
+            out.push_back(static_cast<char>(c));
+        } else {
+            char buf[4];
+            std::snprintf(buf, sizeof(buf), "%%%02X", c);
+            out.append(buf);
+        }
+    }
+    return out;
+}
+
```
Ericson2314: Per https://github.com/NixOS/hydra/pull/1692/commits/b5e42d44af9a60c05711c90b36c5f707cdfe9937#r3171126384 we might not even need this (because the environment variables are set externally and not in this C++, option 2).

If we got with option 1, I think we can replace this with a Nix function, not hand-roll it.

nixos-modules/drv-daemon-module.nix
```diff
@@ -0,0 +1,187 @@
...
+    };
+
+    storeDir = lib.mkOption {
+      type = lib.types.path;
+      default = "/nix/store";
+      description = ''
+        Logical store directory. Must match the store dir used by the
+        queue runner / builder; the value is also what the evaluator
+        appends as `?store=<dir>` when it points `NIX_REMOTE` at the
+        socket, since `unix://` connections ignore `NIX_STORE_DIR`.
```
Ericson2314: I don't think that is true, I don't think they ignore `NIX_STORE_DIR`

Ericson2314: But also, we much rather not rely on `NIX_STORE_DIR` and just use `NIX_REMOTE`, so maybe this fine.

nixos-modules/drv-daemon-module.nix
```diff
@@ -0,0 +1,187 @@
...
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        When true, wire hydra-evaluator to this daemon and order it after hydra-drv-daemon.service.
+
+        Note: routing only takes effect when
+        `allowImportFromDerivation` is also enabled — otherwise
+        `nix-eval-jobs` rejects every IFD before the daemon can see it.
+      '';
+    };
```
Ericson2314: I don't understand why this is a setting for `hydra-drv-daemon` itself. That seems wrong --- why does it care who is using it and for what purpose?

nixos-modules/drv-daemon-module.nix
```diff
@@ -0,0 +1,187 @@
...
+
+    allowImportFromDerivation = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        When true, enable IFD evaluation in hydra.conf. This lets untrusted evals run builders, so it stays opt-in.
+      '';
+    };
```
Ericson2314: Ditto

nixos-modules/drv-daemon-module.nix
```diff
@@ -0,0 +1,124 @@
...
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.callPackage ../subprojects/hydra-drv-daemon/package.nix { };
+    };
+
+    socketPath = lib.mkOption {
+      type = lib.types.path;
+      default = "/run/hydra-drv-daemon/socket";
+      description = ''
+        Socket path used by clients (e.g. hydra-evaluator, imperative `nix-store --realise`) to reach the daemon.
```
Ericson2314:
```suggestion
        Socket path used by clients to reach the daemon.
```

subprojects/hydra-manual/src/architecture.md
```diff
@@ -97,6 +108,9 @@ graph BT
 - `binary-cache`:
   reading and writing Nix binary cache artifacts (NARinfo, NAR files, signatures, presigned uploads)
 
+- `build-logs`:
+  following build-step logs as the queue runner writes them, and reading its step/build notifications; shared by `hydra-ws` and anything else that shows a log live
```
Ericson2314: This should mention `hydra-ad-hoc` in the final commit

Ericson2314 (approved):
I don't like adding a bunch of new code, but most of the stuff here is totally optional, and disabled by default. That makes me feel better about this --- `hydra.nixos.org` will not be affected by the new server. The prep commits that do effect it I have reviewed very carefully.

I have also been dog-fooding this a bit and will continue to do so.

Ultimately I think this use-case is very important for the future of Hydra, and so I am eager to start iterating on it in-tree.

PR #1694, opened 2026-05-01
Move Nix's perl bindings to this repo

Ericson2314:
The Perl bindings are only used by Hydra (to our knowledge) and so it will be easier to maintain them here than the Nix repo.

In particular, sometimes Hydra needs something new, or stops needing something, and if those changes now no longer require coordinating multiple repos / multiple branches (Hydra master uses Nix stable), that will be much easier.

Just like when it was in the Nix repo, the Perl bindings are wired up as both a Meson subproject (for dev builds) and a Nix component in `mkHydraComponents` (for release builds). `package.nix` was rewritten to use plain `stdenv.mkDerivation` instead of Nix's `mkMesonDerivation`, since the latter's `workDir`/`fileset` assumptions conflict with living inside a different repo. (We might want to think about this in the future, though.)

Versioning: the Perl bindings report the version of `nix-store` they are built against, not Hydra's own version. In Nix builds, `preConfigure` writes this into `.version`; in the dev shell, a `shellHook` does the same. The generated `.version` is gitignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,17 +1,39 @@
+# Editor backup files
 *~
+
+# Nix dev shell
 /.direnv/
+
+# Test artifacts
 .test_info.*
+
+# Downloaded frontend assets
 /subprojects/hydra/root/static/bootstrap
 /subprojects/hydra/root/static/fontawesome
 /subprojects/hydra/root/static/js/flot
+
+# Generated SQL files
 /subprojects/hydra/sql/hydra-postgresql.sql
 /subprojects/hydra/sql/hydra-sqlite.sql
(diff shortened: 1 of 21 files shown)


Ericson2314:
I forgot about `nix-serve` also using the Perl bindings....So I rewrote it in C++ https://github.com/edolstra/nix-serve/pull/72/files :)

Ericson2314:
https://github.com/edolstra/nix-serve/pull/72#issuecomment-4364225271 @edolstra agrees `nix-seve` in Perl form is being retired. We have until 26.11 to figure this out, so I am merging this now!

PR #1703, opened 2026-05-04
Add systemd socket activation for `hydra-server` (Starman)

Ericson2314:
Instead of forking Starman to add socket activation support, carry a `Net::Server::Systemd::PreFork` personality inside hydra's own Perl libs and monkey-patch `@Starman::Server::ISA` in `Hydra::Script::Server` when `LISTEN_FDS` is set. This avoids a forked dependency while the upstream PR (https://github.com/miyagawa/Starman/pull/156) is pending.

I am working on a forked foreman for optional local testing, but for now, the VM test tests that that the socket activation works correctly.

### DIFF
--- a/nixos-modules/web-app.nix
+++ b/nixos-modules/web-app.nix
@@ -254,8 +254,14 @@ in
 
     systemd.services.hydra-server = {
       wantedBy = [ "multi-user.target" ];
-      requires = [ "hydra-init.service" ];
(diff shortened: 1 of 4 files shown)


Ericson2314:
~~Leaving this draft until https://github.com/miyagawa/Starman/pull/156 gets some review.~~

PR #1704, opened 2026-05-04
Make the test suite's queue runner and builder go through a nix daemon

Ericson2314:
We were seeing the builder crash due to SQLite WAL issues via the FFI. While we investigate that, we can side-step the issue by making the FFI use a nix daemon rather than directly opening the SQLite database.

Each test that runs builds now starts a fresh `nix-daemon` process for both the queue runner (central store) and the builder, managed 1:1 with their respective processes in `QueueRunnerContext` and`QueueRunnerBuildOne`. The daemon opens the same `local?root=...&store=...` store URI via `NIX_REMOTE`, and clients connect through `unix://...?root=...&store=...` URIs. `NIX_STORE_DIR` is set for the daemon so that substituters see the correct store prefix (note: it is unclear why the `store=....` query param was not good enough for this). The daemon also gets `trusted-users = *` via `NIX_CONFIG` so it accepts unsigned paths from the builder, which is equivalent to how the local store worked (direct local store is always trusted).

The evaluator and other central services continue to open the store directly — only the queue runner and builder are routed through daemons.

Long term, we want Hydra to be pure Rust anyways, so this isn't so bad; the FFI is just a stop-gap, the daemon protocol can also be a stop-gap.

### DIFF
--- a/subprojects/hydra-tests/lib/HydraTestContext.pm
+++ b/subprojects/hydra-tests/lib/HydraTestContext.pm
@@ -53,11 +53,13 @@ sub new {
     # use sandboxed builds.
     my $builder = { root => "$dir/builder" };
     make_path($builder->{root});
+    $builder->{nix_conf_dir} = "$dir/nix/etc/nix";
     $builder->{nix_store_dir} = "$builder->{root}/nix/store";
     $builder->{nix_state_dir} = "$builder->{root}/nix/var/nix";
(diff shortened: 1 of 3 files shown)


Ericson2314:
It looks like this does not fix a recursive nix intermittent failure, but it does fix the other ones --- only `dyn-drv-non-trivial` is failing intermittently now.

PR #1706, opened 2026-05-04
Remove legacy build machine discovery and `SystemStatus`-based status…

Ericson2314:
… reporting

The Perl web controllers were falling back to reading `/etc/nix/machines` (the old Nix remote systems file) when `queue_runner_endpoint` wasn't configured. But this doesn't really make sense because Hydra won't actually fall back to using those machines — the Rust queue runner uses a completely different gRPC self-registration mechanism.

`getMachines()` now always queries the queue runner's `/status/machines` HTTP endpoint, and warns if `queue_runner_endpoint` is missing.

Additionally, the `SystemStatus` DB table was being used as an intermediary for status reporting: the queue runner would write a JSON dump there, and `hydra-queue-runner --status` would read it back. This is no longer needed since the queue runner has a REST API that serves the same data directly. All consumers (`queue-runner-status` page, `hydra-send-stats`) now query the HTTP endpoint instead.

Removed:

- Legacy `/etc/nix/machines` file parsing from `getMachines()`

- `SystemStatus` DB reads from `Root.pm` (the Perl controller)

- The `--status` CLI flag and `dump_status_loop` from the Rust queue runner (no more writing status to the database)

- `upsert_status`, `get_status`, `notify_dump_status`, `notify_status_dumped` from the DB crate

- The `POST /dump_status` HTTP endpoint

- The `buildMachinesFiles` NixOS option (with `mkRemovedOptionModule` deprecation message) and `NIX_REMOTE_SYSTEMS` env var

- The machines file parsing test and related test scaffolding

`hydra-send-stats` and the `queue-runner-status` page now query the queue runner's REST API directly. The `SystemStatus` table is marked for removal in a future migration. The `hydra-send-stats` test now exercises both the no-endpoint and with-endpoint code paths.

### DIFF
--- a/subprojects/hydra/sql/hydra.sql
+++ b/subprojects/hydra/sql/hydra.sql
@@ -628,6 +628,8 @@ create rule IdempotentInsert as on insert to FailedPaths
 
 
 
+-- TODO: Drop this table in a future migration. The queue runner no longer
+-- writes status here; all status is served via its REST API instead.
 create table SystemStatus (
     what text primary key not null,
     status json not null
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@ Running Hydra is currently only supported on NixOS. The [hydra module](https://g
     enable = true;
     hydraURL = "http://localhost:3000";
     notificationSender = "hydra@localhost";
-    buildMachinesFiles = [];
     useSubstitutes = true;
   };
 }
--- a/nixos-modules/web-app.nix
+++ b/nixos-modules/web-app.nix
@@ -24,7 +24,6 @@ let
   env = {
     NIX_REMOTE = "daemon";
     PGPASSFILE = "${baseDir}/pgpass";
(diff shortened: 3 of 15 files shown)


Ericson2314:
@mweinelt I think this is what you wanted, more or less?

mweinelt:
I think so.

subprojects/hydra/sql/hydra.sql
```diff
@@ -628,6 +628,8 @@ create rule IdempotentInsert as on insert to FailedPaths
 
 
 
+-- TODO: Drop this table in a future migration. The queue runner no longer
```
dasJ, line 631: Why not remove this directly? this is only for new databases and a migration that removes the table would be adequate

dasJ: A review by another person would have spotted this btw

mweinelt: As long as we are testing the new queue-runner and rolling back this would be problematic for h.n.o.

PR #1709, opened 2026-05-05
queue-runner: include actual error in `build_log` failure messages

Ericson2314:
The `build_log` gRPC handler discarded the `io::Error` with `|_|`, reporting only `"Failed to write log file."` regardless of whether the cause was a disk issue, a broken stream from a crashed builder, or something else. This made intermittent CI failures impossible to diagnose.

Include the error in the gRPC status message so the builder logs show the real cause (e.g. `"broken pipe"`, `"No space left on device"`).


Ericson2314:
I don't think this error happens much anymore in CI, but this is still a nice thing to do.

PR #1712, opened 2026-05-05
Write CA realisations to binary cache in pure Rust

Ericson2314:
Update harmonia to latest main, where `DrvOutput` uses a `StorePath` instead of a derivation hash. This lets us construct `Realisation` objects directly from data we already have (resolved drv path + output name + concrete output path) without calling `static_output_hashes` through the nix C++ FFI.

Remove the realisation query FFI (`query_raw_realisation`, `InternalRealisation`, `realisation.rs`, `realisation.cpp`) and `static_output_hashes` FFI entirely. The binary-cache crate's `copy_realisation` (which queried the local store via FFI) is replaced with `write_realisation` that accepts a pre-constructed `Realisation`.

In `succeed_step`, after a build step with CA floating outputs succeeds, write a `Realisation` for each output to all S3 binary caches. The realisations are signed with the cache's secret keys.

TODO: also write realisations to the local store's SQLite `Realisations` table (for non-S3 stores) once nix is updated to 2.35, which uses path-based `DrvOutput` matching the new harmonia types.


Ericson2314:
I need to bump Harmonia for other reasons so I'll merge this. But to effectively give @Conni2461 more time to review I will re-instate the no-S3 FFI code path on `nix-next`.

PR #1714, opened 2026-05-06
Update the "reproduce locally" script

HenrySwanson:
The two changes are:
- `build.job` appears to be a string, so `build.job.name` should just be `build.job`.
  - For me, `build.job.name` resolves to an empty string, and this causes the final `nix-build` to build no attributes (so it always succeeds)
- `git clone` doesn't seem to copy over orphaned commits, so if the branch moves and a commit we're interested in is abandoned, we have to add a call to `git fetch` to get that specific commit
  - This happens often when one does a squash or rebase


Ericson2314:
Thanks!

For posterity, looks like `job.name` was supposed to become `job` since 8adb433e3b48638f3faaecc61ae5cb63efefff6e

Ericson2314 (approved):

PR #1729, opened 2026-05-12
Use harmonia for NAR listings instead of Nix C++ FFI

Ericson2314:
Replace the `list_nar_deep` FFI call (which used Nix's C++
`getFSAccessor` and `listNarDeep`) with harmonia's pure-Rust
`parse_nar_listing`. The new `nar_listing()` helper streams NAR bytes
from `nar_from_path` through `harmonia-file-nar`'s parser, producing a
`FileTree<NarFileInfo>` that serializes to the same JSON format.

This removes the last use of `list_nar_deep` from `binary-cache`, so
the FFI function and its `BaseStore` trait method are deleted along
with the C++ implementation and the `list_nar` example.

Also adapts `parse_drv` to take `&[u8]` instead of `&str`, matching
upstream harmonia's `parse_derivation_aterm` signature change.

The previous commit's NixOS test ensures we get this right.


subprojects/crates/binary-cache/src/presigned.rs
```diff
@@ -206,8 +206,12 @@ impl PresignedUploadClient {
     ) -> Result<PresignedUploadResult, CacheError> {
         let start = std::time::Instant::now();
 
-        let ls = store.list_nar_deep(store_path).await?;
-        let stream = Box::new(std::io::Cursor::new(Bytes::from(ls)));
+        let listing = super::nar_listing(store, store_path).await?;
+        let ls_json = serde_json::json!({
+            "version": 1,
+            "root": listing,
```
Conni2461, line 212: should been a real type rather than an real struct with serialize inline `serde_json::json!` 

that would have been much cleaner

subprojects/crates/binary-cache/src/lib.rs
```diff
@@ -586,8 +586,13 @@ impl S3BinaryCacheClient {
         tracing::debug!("start copying path: {path}");
         let mut narinfo = self.path_to_narinfo(store, path).await?;
         if self.cfg.write_nar_listing {
-            let ls = store.list_nar_deep(&narinfo.path).await?;
-            self.upload_listing(&get_ls_path(&narinfo), ls).await?;
+            let listing = nar_listing(store, &narinfo.path).await?;
+            let ls_json = serde_json::json!({
+                "version": 1,
+                "root": listing,
+            });
```
Conni2461, line 593: another duplicated inline serde_json::json

subprojects/crates/binary-cache/src/lib.rs
```diff
@@ -949,3 +954,31 @@ impl debug_info::DebugInfoClient for S3BinaryCacheClient {
...
+    let closure = move |data: &[u8]| {
+        let data = Bytes::copy_from_slice(data);
+        tx.send(Ok(data)).is_ok()
+    };
+
+    tokio::task::spawn({
+        let path = path.clone();
+        let store = store.clone();
+        async move {
+            let _ = store.nar_from_path(&path, closure);
```
Conni2461, line 974: this still dumps the nar using c++ ffi, right? we use just use parse from harmonia and generate a listing from that? how does this avoid ffi? i dont follow your comments made in this pr and the comments in this change

Conni2461 (commented):
why was this merged within 2hours

PR #1737, opened 2026-05-13
Remove `OutputNameOnly` from proto, error on missing path info

Ericson2314:
A successful build must have path info for all its outputs — the
`OutputNameOnly` variant was defensive code that silently dropped
outputs when `query_path_info` failed, hiding bugs.

Replace the `Output` oneof (`OutputNameOnly` | `OutputWithPath`) with
a single `OutputPathInfo` message. The builder now queries path info
per-output immediately after building and errors if any are missing,
before uploads begin. The combined `BTreeMap<OutputName, (StorePath,
PathInfo)>` is passed to `new_success_build_result_info` to avoid
re-querying.

> [!NOTE]
> 
>  `OutputNameOnly` does seem like something that was added on purpose, though. So it is possible that there is a legitamate reason for a successful build to be missing outputs that I am completely not thinking of. In that case, we'll revert this commit, and add a test for that scenario

### DIFF
--- a/subprojects/crates/proto/build.rs
+++ b/subprojects/crates/proto/build.rs
@@ -6,6 +6,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let workspace_version = env::var("CARGO_PKG_VERSION")?;
 
+    println!("cargo:rerun-if-changed=../../proto/v1/streaming.proto");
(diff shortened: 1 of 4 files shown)


Ericson2314:
CC @Conni2461 heads up on this. I was not able to find anything in your original dedicated queue runner git history.

PR #1745, opened 2026-05-16
Remove `nix-utils` C++ FFI, use harmonia daemon protocol

Ericson2314:
Replace the `nix-utils` C++ FFI crate with harmonia's pure Rust daemon protocol client (`ConnectionPool` + `DaemonStore` trait) across all hydra crates. This eliminates the C++ FFI and is roughly 1500 fewer lines of code in the repo.

I had strongly encouraged @Conni2461 to use FFI, thinking we would be able to pare the C++ down to the just the new `DerivationBuilder` abstraction I had separated from the rest of the code base. The issue I had not realized is is that on macOS at least, we have to go through the daemon anyways (because no chroot stores) (and on all platforms we currently are choosing to go through the daemon), so the FFI work ended up just speaking to an out-of-process Nix anyways. In other words, my suggestion was a bad idea.

Also, we were opening the local store directly during testing, but that ended up hitting memory corruption (or similar) issues with the SQLite write-ahead log via the FFI, so in f8524d5bd0a5d17017b7645d7158c6df7832cec2 we switched to also using nix daemons during tests.

At that point, all usage — both in production and in testing — was going through the daemon, and so the FFI had just become an overcomplicated way to speak the daemon protocol.

**Note:** Building still uses `nix build --json` subprocess with build log streaming to the queue-runner preserved. Switching to `build_paths_with_results` via the daemon protocol will be done in a future commit. `realise.rs` containing this invocation is moved from the now-defunct `nix-utils` library to `hydra-builder`.

**Key changes:**

- `nix-utils` is deleted, since it was for FFI.

- Store operations (`is_valid_path`, `query_path_info`, `nar_from_path`, `add_to_store_nar`) go through `ConnectionPool` from `harmonia-store-remote`.

- Config queries (`system`, `extra-platforms`, etc.) replaced by `nix show-config --json` in new `nix_config.rs` module. I don't love shelling out, but it is fine for now. We can revisit this later.

- GC roots are now filesystem symlinks instead of FFI `add_root`.

- `store-transfer` crate's `import` and `export` functions converted from `nix_utils::BaseStore` FFI to `harmonia_store_remote::ConnectionPool` + `DaemonStore` trait. Import uses `nar_size`-based byte counting to split the decompressed NAR stream per path. Export is now async (was sync `spawn_blocking`).

- New `daemon-client-utils` crate with `NixDaemonStoreConfig`: parses `NIX_REMOTE` URI (socket path, store dir, chroot root, state dir) with `to_uri()` for round-tripping. All nix env vars (`NIX_REMOTE`, `NIX_STORE_DIR`, `NIX_STATE_DIR`, `NIX_DAEMON_SOCKET_PATH`) are read in this one place. Precedence rules match nix's `LocalFSStoreConfig` (root derives state and real store dirs, URI params override env vars). Also provides `query_closure` (returns `Vec<ValidPathInfo>` topologically sorted via `petgraph` so dependencies come before dependents — required by `add_to_store_nar` which validates references exist) and `compute_closure_size`.

- `binary-cache::copy_path` / `copy_paths` now accept `ValidPathInfo` instead of `StorePath`, avoiding redundant `query_path_info` calls since `query_closure` already collects the infos during its walk.

- Non-S3 dest stores use `nix copy` subprocess (`NixCopy` backend).

- `.drv` files read via filesystem with `real_store_dir` for chroot stores, parsed with `harmonia-store-aterm`.


Ericson2314:
@Conni2461 OK I made this no longer depend on the error handling reworking, which will take longer. It now just depends on https://github.com/NixOS/hydra/pull/1767 a trivial PR bumping harmonia.

I am not sure how to review something so big, but I think this is ready. The main motivation remains the follow-up work of only sending resolved derivations not build closures, which should help if not solve the scheduling/perf issues.

subprojects/crates/binary-cache/src/debug_info.rs
```diff
@@ -54,11 +51,11 @@ where
 }
 
 pub async fn get_debug_info_build_ids(
-    store: &nix_utils::LocalStore,
-    store_path: &StorePath,
+    real_store_dir: &std::path::Path,
```
Conni2461, line 54: should we use harmonia_store_path::StoreDir here?

subprojects/crates/binary-cache/src/lib.rs
```diff
@@ -50,24 +48,24 @@ pub use crate::presigned::{
...
     let narinfo = narinfo_simple(path, path_info, Compression::None);
-    let queried_references = store
-        .query_path_infos(&narinfo.info.info.references.iter().collect::<Vec<_>>())
-        .await?;
     for r in &narinfo.info.info.references {
-        if !queried_references.contains_key(r) {
+        if daemon_client_utils::query_path_info(store, r)
+            .await?
+            .is_none()
+        {
```
Conni2461, line 67: unsure if we should add a lru cache here

subprojects/crates/binary-cache/src/lib.rs
```diff
@@ -602,20 +584,44 @@ impl S3BinaryCacheClient {
...
-        };
 
         tokio::task::spawn({
             let path = narinfo.path.clone();
             let store = store.clone();
             async move {
-                let _ = store.nar_from_path(&path, closure);
+                use harmonia_protocol::types::DaemonStore;
+                use tokio::io::AsyncBufReadExt as _;
+                let result: Result<(), Box<dyn std::error::Error + Send + Sync>> = async {
```
Conni2461: think a real error type should be used here, now that we use thiserror everywhere in the crate

subprojects/crates/binary-cache/src/narinfo.rs
```diff
@@ -219,14 +217,14 @@ pub fn parse_narinfo(input: &str) -> Result<NarInfo, NarInfoError> {
                 references = val
                     .split_whitespace()
                     .filter(|s| !s.is_empty())
-                    .map(StorePath::from_base_path)
-                    .collect::<Result<_, _>>()?;
+                    .map(|s| s.parse().expect("valid store path"))
```
Conni2461: expect can now result in a panic, previously this would have been an error we've returned

subprojects/crates/binary-cache/src/narinfo.rs
```diff
@@ -219,14 +217,14 @@ pub fn parse_narinfo(input: &str) -> Result<NarInfo, NarInfoError> {
...
+                    .collect();
             }
             "Deriver" => {
                 deriver = if val.is_empty() {
                     None
                 } else {
-                    Some(StorePath::from_base_path(val)?)
+                    Some(val.parse().expect("valid store path"))
```
Conni2461: same here, expect used

subprojects/crates/binary-cache/src/presigned.rs
```diff
@@ -155,21 +152,41 @@ impl PresignedUploadClient {
...
+                        let data = Bytes::copy_from_slice(buf);
+                        let len = data.len();
+                        if tx.send(Ok(data)).is_err() {
+                            break;
+                        }
+                        nar_reader.consume(len);
+                    }
+                    Ok(())
+                }
+                .await;
```
Conni2461: String error and dupliacted with code in subprojects/crates/binary-cache/src/lib.rs, lets not do that inline and rather add a separate function for that

Conni2461 (changes requested):
part one

subprojects/crates/binary-cache/src/debug_info.rs
Ericson2314: Possibly?

I was intentionally only using `StoreDir` for logical store dirs, and `std::path::Path` for physical store dirs. Former is platform agnostic `std::path::Path` underneath the hood, and then `std::path::Path` is special on windows.

subprojects/crates/store-transfer/src/export.rs
```diff
@@ -42,28 +40,59 @@ pub fn export(
...
+                break;
+            }
+            use tokio::io::AsyncWriteExt as _;
+            bytes_written += buf.len() as u64;
+            encoder.write_all(buf).await?;
+            let len = buf.len();
+            nar_reader.consume(len);
         }
 
         assert_eq!(
```
Conni2461: hmm, that was already here and i know why its good to have that, but it would result in a panic in prod. i think a debug_assert would be better or a error return

subprojects/crates/store-transfer/src/export.rs
```diff
@@ -42,28 +40,59 @@ pub fn export(
...
-        return;
+        return Ok(());
     }
 
     // Stream all NARs as one continuous zstd-compressed stream.
-    let mut encoder =
-        zstd::stream::Encoder::new(NarChunkWriter(tx), 3).expect("failed to create zstd encoder");
+    // We use a duplex pipe: write compressed data to one end,
+    // spawn a task to read from the other and send as gRPC chunks.
+    let (compressed_writer, compressed_reader) = tokio::io::duplex(256 * 1024);
```
Conni2461: can we turn the magic into constant values?

subprojects/crates/store-transfer/src/export.rs
```diff
@@ -42,28 +40,59 @@ pub fn export(
...
+
+    let tx_clone = tx.clone();
+    let chunk_sender = tokio::spawn(async move {
+        use tokio::io::AsyncReadExt as _;
+        let mut reader = compressed_reader;
+        let mut buf = vec![0u8; 64 * 1024];
```
Conni2461: same, can we turn this magic value into a constant

Conni2461: we also use the same constants in import.rs

subprojects/hydra-queue-runner/src/io/stats.rs
```diff
@@ -232,7 +232,7 @@ pub struct StoreStats {
 
 impl StoreStats {
```
Conni2461, line 233: if we no longer have these stats we can also stop returning them rather than returning 0 for everything. and drop these prom metrics alltogether

subprojects/hydra-builder/src/state.rs
```diff
@@ -230,23 +240,38 @@ impl State {
...
+        let sysinfo = crate::system::SystemLoad::new(&self.config.build_dir, store_path)?;
 
         Ok(PingMessage {
             machine_id: self.id.to_string(),
             load1: sysinfo.load_avg_1,
             load5: sysinfo.load_avg_5,
             load15: sysinfo.load_avg_15,
             mem_usage: sysinfo.mem_usage,
-            pressure: sysinfo.pressure,
+            pressure: sysinfo.pressure.map(|p| PressureState {
```
Conni2461: unnecessary map

subprojects/hydra-builder/src/state.rs
```diff
@@ -829,41 +865,44 @@ async fn import_requisites<T: IntoIterator<Item = StorePath>>(
...
+    pool: harmonia_store_remote::ConnectionPool,
     metrics: Arc<crate::metrics::Metrics>,
     nars: Vec<StorePath>,
 ) -> anyhow::Result<()> {
-    let nars = store
-        .query_requisites(&nars.iter().collect::<Vec<_>>())
+    // Compute full closure by walking references via daemon protocol.
+    // query_closure returns ValidPathInfos in dependency order with
+    // path infos already populated, so we don't need to re-query.
+    let closure = binary_cache::query_closure(&pool, &nars)
```
Conni2461: we should not mix this, the actual import is in daemon_client_utils and down below we are pulling that function from there. maybe binary_cache should not reexport that function in the first place

subprojects/hydra-builder/src/state.rs
```diff
@@ -1118,3 +1183,11 @@ async fn new_success_build_result_info(
...
     })
 }
+
+/// Compute the total NAR size of a path's closure by walking references.
+async fn compute_closure_size(
+    pool: &harmonia_store_remote::ConnectionPool,
+    path: &StorePath,
+) -> u64 {
+    daemon_client_utils::compute_closure_size(pool, path).await
+}
```
Conni2461: we dont need that, its only used once and we can use `daemon_client_utils::compute_closure_size` directly

subprojects/crates/nix-support/src/lib.rs
```diff
@@ -532,7 +532,7 @@ mod tests {
         let fs = FilesystemOperations {
             real_store_dir: store_dir.to_path().to_owned(),
         };
-        let bp = Box::pin(parse_build_product(&store_dir, &fs, line)).await;
+        let bp = parse_build_product(&store_dir, &fs, line).await;
```
Conni2461:
```
error: large future with a size of 17016 bytes
   --> subprojects/crates/nix-support/src/lib.rs:535:18
    |
535 |         let bp = parse_build_product(&store_dir, &fs, line).await;
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider `Box::pin` on it: `Box::pin(parse_build_product(&store_dir, &fs, line))`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_futures
```

subprojects/crates/daemon-client-utils/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+///
+/// Use [`to_uri`](Self::to_uri) to reconstruct a `unix://` URI
+/// suitable for `nix copy --from` etc.
+pub struct NixDaemonStoreConfig {
+    /// Path to the daemon socket.
+    pub socket: String,
+    /// Logical store directory (e.g. `/nix/store`).
+    pub store_dir: StoreDir,
+    /// Chroot root directory, if any (e.g. `/foo`).
+    pub root: Option<String>,
```
Conni2461: can't this be a `Option<PathBuf>`

subprojects/crates/daemon-client-utils/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+    };
+
+    // Derive socket. URI path overrides everything; NIX_DAEMON_SOCKET_PATH
+    // is a fallback for when the URI has no path (e.g. `unix://?store=...`).
+    // Final default: stateDir / "daemon-socket/socket".
+    let socket = socket_from_uri.or(explicit_socket).unwrap_or_else(|| {
+        state_dir
+            .join("daemon-socket/socket")
+            .to_string_lossy()
+            .into_owned()
```
Conni2461: we recently got right of to_string_lossy usage and disallowed that in clippy as well, cant we handle the utf8 error here?

subprojects/crates/daemon-client-utils/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+    )
+}
+
+/// Factored out pure internal function for unit testing purposes.
+fn parse_nix_remote_from(
+    nix_remote: Option<&str>,
+    nix_store_dir: Option<&str>,
+    nix_state_dir: Option<&str>,
+    nix_daemon_socket_path: Option<&str>,
+) -> Result<NixDaemonStoreConfig, String> {
```
Conni2461, line 92: String Error

subprojects/crates/daemon-client-utils/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+    while let Some(p) = queue.pop() {
+        if !visited.insert(p.clone()) {
+            continue;
+        }
+        let mut guard = pool.acquire().await?;
+        let info = guard
+            .client()
+            .query_path_info(&p)
+            .await?
+            .ok_or_else(|| DaemonError::custom(format!("path '{p}' is not valid")))?;
```
Conni2461, line 175: this happens that often that i acutlly think right now that an abstraction on top with a cache would be helpful, especially because we implement query_closure in rust now

subprojects/crates/daemon-client-utils/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+            .collect::<Vec<StorePath>>()
+    };
+
+    Ok(sorted_paths
+        .into_iter()
+        .filter_map(|p| {
+            let info = infos.remove(&p)?;
+            Some(ValidPathInfo { path: p, info })
+        })
+        .collect())
```
Conni2461: cant we merge these two iter calls into one

subprojects/hydra-queue-runner/src/server/grpc.rs
```diff
@@ -547,18 +546,29 @@ impl RunnerService for Server {
...
         let paths: Vec<_> = req.into_inner().paths.into_iter().map(|p| p.0).collect();
 
-        let infos = self
-            .state
-            .store
-            .query_path_infos(&paths.iter().collect::<Vec<_>>())
-            .await
-            .map_err(|e| tonic::Status::internal(format!("query_path_infos failed: {e}")))?;
+        let mut infos = hashbrown::HashMap::new();
+        for path in &paths {
```
Conni2461: can we preallocate the correct capacity

subprojects/hydra-queue-runner/src/server/grpc.rs
```diff
@@ -547,18 +546,29 @@ impl RunnerService for Server {
...
+            infos.insert(path.clone(), info);
+        }
 
         let (tx, rx) = mpsc::unbounded_channel();
+        let pool = self.state.pool.clone();
 
-        tokio::task::spawn_blocking(move || {
-            let store = nix_utils::LocalStore::init();
-            store_transfer::export::export(&store, &paths, &infos, &tx);
+        tokio::spawn(async move {
```
Conni2461: i like something like this to make it clear that pool is only used for that async block

```rs
tokio::spawn({
  let pool = self.state.pool.clone();
  asnyc move {
    // ...
  }
});

```

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -425,12 +466,45 @@ impl State {
...
+                        .add_ca_to_store(
+                            &format!("{}.drv", basic_drv.name),
+                            harmonia_store_content_address::ContentAddressMethodAlgorithm::Text,
+                            &refs,
+                            false,
+                            source,
+                        )
+                        .await?;
+                    info.path
+                };
```
Conni2461: is this something that should happen in a utils function?

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -1058,13 +1132,13 @@ impl State {
         let task = tokio::task::spawn({
             async move {
                 loop {
-                    let local_store = nix_utils::LocalStore::init();
+                    let local_store: harmonia_store_remote::ConnectionPool = self.pool.clone();
```
Conni2461: why is the type explicit here?

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -1249,34 +1323,50 @@ impl State {
...
+                        Ok(out) => {
+                            tracing::error!(
+                                "nix copy to {dest_uri} failed: {}",
+                                String::from_utf8_lossy(&out.stderr),
+                            );
+                        }
+                        Err(e) => {
+                            tracing::error!("Failed to run nix copy to {dest_uri}: {e}");
+                        }
+                    }
```
Conni2461, line 1352: can we move this part in a separate function

Conni2461 (changes requested):
part two, not sure if i missed things. its a lot

subprojects/crates/store-transfer/src/export.rs
Ericson2314: I would say it could be an error return, but a `debug_assert` would have me worried because would definitely not want this to silently fail to be true on prod --- this would indicate an internal bug that needs to be fixed immediately.

subprojects/crates/store-transfer/src/export.rs
Ericson2314: Yeah good to deduplicate constants when we give them a name.

subprojects/crates/nix-support/src/lib.rs
Ericson2314: oops yeah this looks like a bad rebase from me at some point. Let's definitely put that `Box::pin` back.

subprojects/crates/daemon-client-utils/src/lib.rs
Ericson2314: Yes good point, because this is a physical not logical path.

subprojects/hydra-builder/src/state.rs
```diff
@@ -386,9 +411,15 @@ impl State {
...
-        let store = nix_utils::LocalStore::init();
+        let nix_config = daemon_client_utils::parse_nix_remote()
+            .map_err(|e| JobFailure::Preparing(anyhow::anyhow!(e)))?;
+        let daemon_socket = nix_config.socket;
+        let store_dir = nix_config.store_dir;
+        let pool = harmonia_store_remote::ConnectionPool::with_store_dir(
+            &daemon_socket,
+            store_dir.clone(),
+            harmonia_store_remote::PoolConfig::default(),
+        );
```
Conni2461, line 418: this was a ffi limitation that we had to recreate the same localstore every time. can we now reuse this?

subprojects/hydra-queue-runner/src/state/mod.rs
Ericson2314: Probably could go away yeah

subprojects/hydra-queue-runner/src/state/mod.rs
Ericson2314: Yeah this could just go in `harmonia-daemon-client` I think

subprojects/hydra-queue-runner/src/state/mod.rs
Ericson2314: I say we look in the history for how you had it before and match that :)

subprojects/hydra-builder/src/state.rs
Ericson2314: (from talk on call) let's consider this in a follow up PR, as we could in fact reintroduce the same GC root issue we had before by doing this.

Ericson2314:
I believe we've addressed everything, thanks again @Conni2461 for reviewing

PR #1747, opened 2026-05-17
hydra-eval-jobs: use nix-eval-jobs's --select rather --expr

Mic92:
The current use of builtins.getFlake, mean we don't have pure evaluation.


Mic92:
Test looked good. I think this is ready to ship.

Ericson2314 (approved):

PR #1748, opened 2026-05-17
queue-runner and builder: fail faster on infrastructure errors

Ericson2314:
WIP. See commit for preliminary details --- it is a long run on commit message because it's a bunch of separate edits tacked together, and it will probably become multiple commits once finished up.

I am hoping this will make it easier to fix the remaining issues, especially the ones that are hard to reproduce out of real-world testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1838,7 +1838,6 @@ dependencies = [
 name = "hydra-queue-runner"
 version = "0.1.0"
 dependencies = [
- "anyhow",
  "arc-swap",
  "async-compression",
  "atomic_float",
@@ -1847,6 +1846,7 @@ dependencies = [
  "byte-unit",
  "bytes",
  "clap",
+ "color-eyre",
  "daemon-client-utils",
  "db",
  "fs-err",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,6 +317,7 @@ dependencies = [
  "bytes",
  "configparser",
  "daemon-client-utils",
+ "eyre",
  "foldhash 0.2.0",
  "fs-err",
  "futures",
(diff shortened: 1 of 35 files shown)


Ericson2314:
@dasJ the second commit of this I would very much like your feedback on. The general rules I am thinking are:

- destination stores are always best effort (don't die if fail to upload)
- queue doesn't die on bad builders
- builders don't die on bad queue runner (but could change that?)
- queue runner *does* die on bad PostgreSQL
- when in doubt `?` and (at the top of the call stack) die

I think all the edits here fit that pattern but I definitely didn't finish reviewing them yet.

I am also confused why we had

```
tracing::error!("Rejecting new machine creation: {e}");
return Err(tonic::...);
```

in so many places before.

dasJ:
I am fully on board with most of these rules. Not dying is never a bad thing ;)

The only thing I am unsure is the "don't die if fail to upload". Wdym by that? I don't see any unwraps or expects so I assume that's "assume the build succeeded". I don't think that's what we want because it could mean we end up with an incomplete binary cache.

Ericson2314:
I think doing https://github.com/NixOS/hydra/pull/1755 first will make this more clear.

> The only thing I am unsure is the "don't die if fail to upload". Wdym by that?

To be clear, this was something I am keeping! It already is OK with "destination stores" being down, and I while I am not obvious to me that we want that, I didn't feel certain enough that we didn't to change it.

Ericson2314:
> I am fully on board with most of these rules. Not dying is never a bad thing ;)

Also to be clear, this PR should generally be making is die *more*. The counter-intuitive logic of "fail fast" is that not-dying can allow simple problems to fester, only for one to go down with a more complex problem later, and so it ends up being more sysadmin/debubgging work than it would be otherwise.

PR #1749, opened 2026-05-18
GitlabStatus: update status for cached builds

jmbaur:
This is similar to what exists for the GithubStatus plugin, since the GitLab external commit status API expects all jobs to have a status published on every commit, not just the ones that have changed between commits.


subprojects/hydra/lib/Hydra/Plugin/GitlabStatus.pm
```diff
@@ -91,4 +94,14 @@ sub buildFinished {
     common(@_, 2);
 }
 
+sub cachedBuildQueued {
```
dasJ, line 97: The API never seizes to amaze me

dasJ (approved):

PR #1753, opened 2026-05-18
`nix_utils::BaseStore`: return `Result` from all fallible methods 

Ericson2314:
Previously the trait implementations silently swallowed FFI errors as
"path not found" / empty / zero. This was dubious and could cause
mysterious failures. For example, a daemon connection error would look
like "path is present" to callers, causing paths to be skipped from
import and builds to fail with cryptic errors instead. We don't want
that! Better to have an earlier clear failure.

The C++ `query_path_info` now checks `isValidPath` first and returns
a `found: false` sentinel instead of throwing `InvalidPath` —
eliminating brittle error-message string matching on the Rust side.

All callers updated to propagate errors with `?`.

### DIFF
--- a/subprojects/crates/binary-cache/src/lib.rs
+++ b/subprojects/crates/binary-cache/src/lib.rs
@@ -56,15 +56,16 @@ pub async fn path_to_narinfo(
     store: &nix_utils::LocalStore,
     path: &StorePath,
 ) -> Result<NarInfo, CacheError> {
-    let Some(path_info) = store.query_path_info(path).await else {
-        return Err(CacheError::PathNotFound {
+    let path_info = store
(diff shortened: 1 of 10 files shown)


dasJ:
21 minutes, John

Ericson2314:
Would you like to review this one? I considered it pretty straight-forward. I also talking about a little bit with @Mic92 over the weekend (though as part of a larger conversation) about it.

https://github.com/NixOS/hydra/pull/1748 This is the more interesting one that I would much prefer eyes on, if you have limited time. (I will push to it again in a moment.)

Mic92 (approved):
Looks rather mechanical.

PR #1755, opened 2026-05-19
Replace `anyhow` with structured error types in support crates

Ericson2314:
Introduce typed error enums so callers can distinguish error kinds
(e.g. DB infrastructure failure vs bad data) without string matching.
This lays the groundwork for the future error handling cleanups, by
allowing us to see more clearly what changes are made.

db crate:
- New `db::Error` enum: `Sql` (infrastructure) vs `Data(DataError)`
- `DataError` has `StorePath`, `StorePathName`, `IntConversion`, and
  `BuildProductMissingPath { build_id, productnr }` variants
- `BuildProductRow` now carries `build` and `productnr` from the query
  so error messages identify the offending row
- Removed `anyhow` dependency

nix-utils crate:
- Replaced `Error::Anyhow` catch-all with `DerivationParse` and
  `StorePathName` variants
- Fixed CA fixed-output test to use correct output path
- Removed `anyhow` dependency

store-transfer crate:
- New `Error` enum: `Store`, `Grpc`, `Io`, `Protocol`, `Join`
- `ProtocolError` sub-enum: `EmptyStream`, `MissingHeader`,
  `MissingGrpcField`, `InvalidPathInfo`, `TruncatedNar`
- Removed `anyhow` dependency

nix-support crate:
- `anyhow::Result` → `std::io::Result` (all errors are IO)
- Removed `anyhow` dependency

hydra-tracing crate:
- `anyhow::Result` → `Result<_, Box<dyn Error>>` (init-only)
- Removed `anyhow` dependency

binary-cache crate:
- `InvalidS3Scheme` and `InvalidCompression` error structs with
  named `got` field replace `String` error variants
- `UrlParseError` variants renamed for clarity

proto crate:
- `NarInfoConvertError` now implements `std::error::Error`

Also adds `sqlx-prepare.sh` for regenerating the sqlx offline cache.


Mic92:
Sounds good. This composes better with the underlying libraries

Mic92:
Should the postgres server not also be stopped in the script?

**Update** missed the trap

subprojects/hydra-builder/src/main.rs
Ericson2314, line 1: Note: end applications are done in the next PR

subprojects/crates/nix-utils/src/drv.rs
```diff
@@ -83,7 +85,7 @@ mod tests {
             &store_dir,
             &drv_path,
             format!(
-                r#"Derive([("out","/nix/store/{HASH}-test-src","sha256","deadbeef00000000000000000000000000000000000000000000000000000000")],[],[],"{0}","{0}",[],[("name","test-src")])"#,
```
Conni2461, line 86: i still think we want `{HASH}`, a fake hash here

subprojects/crates/tracing/src/lib.rs
```diff
@@ -60,7 +60,8 @@ impl Drop for TracingGuard {
 }
 
 #[cfg(feature = "otel")]
-fn init_tracer_provider() -> anyhow::Result<opentelemetry_sdk::trace::SdkTracerProvider> {
+fn init_tracer_provider()
+-> Result<opentelemetry_sdk::trace::SdkTracerProvider, Box<dyn std::error::Error>> {
```
Conni2461: hmm i like `Box<dyn std::error::Error>` less than anyhow. can we also use `thiserror` here? or keep anyhow as long as we have anyhow in the applications?

Conni2461 (approved):
Left two nits :D other than that, we can merge this

subprojects/crates/tracing/src/lib.rs
Ericson2314: Yes good point I meant to get rid of that.

subprojects/crates/nix-utils/src/drv.rs
Ericson2314: This is a preexisting error (should be separate commit). This is actually correct, the ATerm parser will complain if the hash and the store path do not agree, which is good.

Conni2461: unrelated question but are we running `cargo test` currently in CI?

Ericson2314: No we aren't, it wasn't working when I first tried, only impurely. That would be a good thing to fix.

Ericson2314:
OK Thanks! All the `Box<dyn Error + ....>` is gone!

PR #1759, opened 2026-05-21
feat(queue-runner): implement concurrent build injection and faster reruns on early exit

Conni2461:


Conni2461:
I am still testing with that, i am convinced that it improves thing, i dont know if it fixes hexas scheduling

Ericson2314 (approved):

PR #1766, opened 2026-05-27
Remove instances of to_string_lossy

artemist:
The `to_string_lossy` function is rarely what one wants, as it silently replaces characters with replacements. Remove all usages by either handling utf-8 errors or replacing strings with  `PathBuf` objects, where paths should be stored.


Conni2461 (approved):

Ericson2314:
Thanks!

PR #1768, opened 2026-05-27
hydra-proto: add `Algorithm` `From` impls, simplify Hash conversions

Ericson2314:
Add bidirectional conversions between `harmonia_utils_hash::Algorithm`
and `nix::store::v1::hash::Algorithm`, then refactor the `Hash`
`From`/`TryFrom` impls to delegate to them instead of duplicating the
match inline.

This is how it should have been all long --- the `From`/`TryFrom` impls
should always do one thing and rely on other impls modularly --- but I
didn't notice before we were missing this factoring out.

Note: prost stores proto enum fields as `i32` and generates its own Rust
enum; there's no way to tell prost-build to use a pre-existing enum type
for a field. These `From` impls are the standard bridge between the
generated proto enum and harmonia's domain type.

### DIFF
--- a/subprojects/crates/proto/src/lib.rs
+++ b/subprojects/crates/proto/src/lib.rs
@@ -60,20 +60,39 @@ impl std::fmt::Display for NarInfoConvertError {
 
 impl std::error::Error for NarInfoConvertError {}
 
+// -- Hash Algorithm --
+
+impl From<harmonia_utils_hash::Algorithm> for nix::store::v1::hash::Algorithm {
+    fn from(algo: harmonia_utils_hash::Algorithm) -> Self {
+        use harmonia_utils_hash::Algorithm;
+        match algo {
+            Algorithm::SHA256 => Self::Sha256,
+            Algorithm::SHA512 => Self::Sha512,
+            Algorithm::SHA1 => Self::Sha1,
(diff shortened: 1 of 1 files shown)


subprojects/crates/proto/src/lib.rs
```diff
@@ -60,20 +60,41 @@ impl std::fmt::Display for NarInfoConvertError {
...
+
+    fn try_from(algo: nix::store::v1::hash::Algorithm) -> Result<Self, Self::Error> {
+        match algo {
+            nix::store::v1::hash::Algorithm::Sha256 => Ok(Self::SHA256),
+            nix::store::v1::hash::Algorithm::Sha512 => Ok(Self::SHA512),
+            nix::store::v1::hash::Algorithm::Sha1 => Ok(Self::SHA1),
+            nix::store::v1::hash::Algorithm::Md5 => Ok(Self::MD5),
+            nix::store::v1::hash::Algorithm::Blake3 => Ok(Self::BLAKE3),
+        }
+    }
```
amaanq: The `Error` variant is not used here, so we don't need this I guess?

Ericson2314: good catch!

amaanq (approved):

PR #1770, opened 2026-05-28
tests: reset `SIGCHLD` before exec'ing nix-daemon

amaanq:
yath sets `SIGCHLD=IGN`, which the spawned nix-daemon inherits, and its
fork+bind workaround for long socket paths then dies on waitpid (ECHILD).
This commit resets to the default in the `IPC::Run` init callback.


Ericson2314:
I mentioned to @amaanq it would be good to get some confirmation that Yath is actually doing this, as that is quite surprising to me.

amaanq:
I think this is an incorrect fix, I got lucky with it passing a bunch locally.

PR #1771, opened 2026-05-28
Use daemon protocol rather than shelling out to build derivation

Ericson2314:
Depends on #1745


subprojects/hydra-builder/src/state.rs
```diff
@@ -470,82 +462,116 @@ impl State {
...
-            .await
-            .map_err(|e| {
-                JobFailure::Upload(
-                    anyhow::Error::from(e).context("failed to stream build log to queue-runner"),
-                )
-            })?;
-
-        let status = child
-            .wait()
+        // Build via daemon protocol using build_paths_with_results.
```
Conni2461, line 414: can we not do this inline, in that function. the function is already quite big

subprojects/hydra-builder/src/state.rs
```diff
@@ -470,82 +462,116 @@ impl State {
...
+        let build_results = {
+            use futures::stream::StreamExt as _;
+            let build_paths = [build_path];
+            let mut result_log = std::pin::pin!(
+                harmonia_protocol::types::DaemonStore::build_paths_with_results(
+                    guard.client(),
+                    &build_paths,
+                    harmonia_protocol::daemon_wire::types2::BuildMode::Normal,
+                )
+            );
```
Conni2461, line 456: what about

```
            m.max_log_size,
            m.max_silent_time,
            m.build_timeout,
```

Conni2461: is that something we need to implement on client side? then we should do that

subprojects/hydra-builder/src/state.rs
```diff
@@ -470,82 +462,116 @@ impl State {
...
-                "nix build exited with {status}"
-            )));
-        }
+            .map_err(|e| JobFailure::Build(anyhow::anyhow!("daemon connection failed: {e}")))?;
+
+        // Set up a channel to stream log lines to the gRPC build_log
+        // concurrently with the build.  A transport error here (e.g.
+        // nginx sending an HTTP/2 GOAWAY) says nothing about whether
+        // the derivation builds, so map it to Upload so the
+        // queue-runner retries the step instead.
```
Conni2461: the comment isnt correct, it no longer maps to an upload error

subprojects/hydra-builder/src/state.rs
```diff
@@ -470,82 +462,116 @@ impl State {
...
+
+        // Check for build failure.
+        use harmonia_protocol::daemon_wire::types2::BuildResultInner;
+        let success = match build_result.result.inner {
+            BuildResultInner::Success(s) => s,
+            BuildResultInner::Failure(f) => {
+                return Err(JobFailure::Build(anyhow::anyhow!(
+                    "build failed: {:?}: {}",
+                    f.status,
+                    String::from_utf8_lossy(&f.error_msg)
```
Conni2461: another lossy string.

Conni2461 (changes requested):

subprojects/hydra-builder/src/state.rs
artemist: I think this is possible to do either server side with `SetOptions` or client side.

Ericson2314: I see it is done in the factored out function

Ericson2314:
Looks like everything was fixed here too, and it has been 2 days.

Ericson2314:
Ah I am going to push this on top of https://github.com/NixOS/hydra/pull/1786 to see if the connection pool issues go away.

PR #1772, opened 2026-05-28
Send force-resolved `BasicDerivation` to builder via protobuf

Ericson2314:
Depends on #1771

### DIFF
--- a/subprojects/hydra-queue-runner/src/state/machine.rs
+++ b/subprojects/hydra-queue-runner/src/state/machine.rs
@@ -488,6 +488,7 @@ pub enum Message {
         max_silent_time: i32,
         build_timeout: i32,
         presigned_url_opts: Option<PresignedUploadOpts>,
+        resolved_drv: Box<hydra_proto::nix::store::derivation::v1::Basic>,
     },
     AbortMessage {
         build_id: uuid::Uuid,
@@ -506,6 +507,7 @@ impl Message {
                 max_silent_time,
                 build_timeout,
                 presigned_url_opts,
+                resolved_drv,
             } => runner_request::Message::Build(BuildMessage {
                 build_id: build_id.to_string(),
                 drv: Some(hydra_proto::ProtoStorePath::from(drv)),
@@ -513,6 +515,7 @@ impl Message {
                 max_silent_time,
                 build_timeout,
                 presigned_url_opts,
+                resolved_drv: Some(*resolved_drv),
             }),
             Self::AbortMessage { build_id } => runner_request::Message::Abort(AbortMessage {
                 build_id: build_id.to_string(),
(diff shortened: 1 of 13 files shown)


subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -421,97 +421,100 @@ impl State {
...
-            if let Some(resolved_path) = resolved_path {
+        // CA floating derivations and originally-Deferred derivations
+        // both need a two-phase build: write the resolved drv to the
+        // store via the daemon, get its assigned path, and (if it
+        // differs from the original) dispatch a new step for it.
+        if was_deferred
+            || basic_drv
+                .outputs
+                .iter()
+                .all(|o| matches!(o.1, DerivationOutput::CAFloating(_)))
```
Conni2461: this is true if basic_drv.outputs is empty: we should do add a `!basic_drv.outputs.is_empty()`

subprojects/hydra-queue-runner/src/state/mod.rs
```diff
@@ -421,97 +421,100 @@ impl State {
...
-                            source,
-                        )
-                        .await?;
-                    info.path
-                };
-                // If resolution changed the drv, we need a two-phase
-                // build; otherwise just build the original directly.
-                Ok((&resolved_path != drv).then_some(resolved_path))
+                    .values()
+                    .all(|o| matches!(o, DerivationOutput::Deferred));
```
Conni2461: are always all outputs deferred? or should this be `any`, if we have both Deferred and others

subprojects/hydra-queue-runner/src/state/mod.rs
Ericson2314: I think that's fine, actually! A 0 output derivation would never happen, but if it did, it wouldn't be using input addressing, so we could do this two-stop process.

subprojects/hydra-queue-runner/src/state/mod.rs
Ericson2314: All the outputs should/must be the same type or the derivation is illegal. But I do think `any` is more the spirit of what where doing so yeah I agree let's change it.

subprojects/hydra-queue-runner/src/state/machine.rs
```diff
@@ -488,6 +488,7 @@ pub enum Message {
         max_silent_time: i32,
         build_timeout: i32,
         presigned_url_opts: Option<PresignedUploadOpts>,
+        resolved_drv: Option<Box<hydra_proto::nix::store::derivation::v1::Basic>>,
```
Ericson2314: Why is this optional? that seems wrong

PR #1773, opened 2026-05-28
feat(hydra): websocket service for live log streaming 

Conni2461:


subprojects/hydra-builder/package.nix
```diff
@@ -34,11 +34,14 @@ rustPlatform.buildRustPackage {
...
-  # The source fileset above intentionally excludes hydra-queue-runner,
+  # The source fileset above intentionally excludes hydra-queue-runner, ...,
   # so drop it from the workspace members to keep cargo from trying to
   # load its (absent) manifest.
   postPatch = ''
-    sed -i 's|"subprojects/hydra-queue-runner", ||' Cargo.toml
+    sed -i \
+      -e 's|"subprojects/hydra-queue-runner",||' \
+      -e 's|"subprojects/hydra-ws",||' \
+      Cargo.toml
```
Ericson2314:
```suggestion
    sed -i '/hydra-builder/!{/"subprojects\/hydra-/d;}' Cargo.toml
```

subprojects/hydra-queue-runner/package.nix
```diff
@@ -36,11 +36,14 @@ rustPlatform.buildRustPackage {
...
-  # The source fileset above intentionally excludes hydra-builder,
+  # The source fileset above intentionally excludes hydra-builder, ...,
   # so drop it from the workspace members to keep cargo from trying to
   # load its (absent) manifest.
   postPatch = ''
-    sed -i 's|"subprojects/hydra-builder", ||' Cargo.toml
+    sed -i \
+      -e 's|"subprojects/hydra-builder",||' \
+      -e 's|"subprojects/hydra-ws",||' \
+      Cargo.toml
```
Ericson2314:
```suggestion
    sed -i '/hydra-queue-runner/!{/"subprojects\/hydra-/d;}' Cargo.toml
```

subprojects/hydra-ws/package.nix
```diff
@@ -0,0 +1,67 @@
...
+
+  # The source fileset above intentionally excludes hydra-queue-runner, ...,
+  # so drop it from the workspace members to keep cargo from trying to
+  # load its (absent) manifest.
+  postPatch = ''
+    sed -i \
+      -e 's|"subprojects/hydra-builder",||' \
+      -e 's|"subprojects/hydra-queue-runner",||' \
+      Cargo.toml
+  '';
```
Ericson2314, line 44:
```suggestion
    sed -i '/hydra-ws/!{/"subprojects\/hydra-/d;}' Cargo.toml
```

nixos-modules/ws-server-module.nix
```diff
@@ -0,0 +1,160 @@
...
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.services.hydra-ws-dev = {
+      description = "Hydra WebSocket service";
+
+      requires = [
+        "nix-daemon.socket"
+      ];
```
Ericson2314, line 81: Is this true? I hope not!

subprojects/hydra-ws/src/handler.rs
```diff
@@ -0,0 +1,450 @@
...
+
+use crate::config::Stream;
+use crate::messages::{HydraWsRequest, HydraWsResponse};
+use crate::state::State;
+use crate::subscriptions::Subscriptions;
+use crate::tailer::LineKind;
+
+const OUT_CHANNEL_CAPACITY: usize = 4096;
+
+#[derive(Debug, thiserror::Error)]
```
Ericson2314, line 23: ~~`thiserror` -> `eyre`~~

subprojects/hydra-ws/src/handler.rs
```diff
@@ -0,0 +1,450 @@
...
+                .await
+                .map_err(|e| {
+                    tracing::error!("DB query failed: {e}");
+                    e
+                })?
+                .ok_or(LogsStartError::BuildNotFound {
+                    build_id: build_id_i32,
+                })?
+        }
+    };
```
Ericson2314, line 161: Note that longer term I would like to make this component just work in terms of drv paths, and not know about builds and build steps, but some things need to happen before that first.

This is fine for now.

Ericson2314:
There is an architecture diagram in the docs we should update for this. Please re-run `scripts/dependency-diagram.py`.

subprojects/hydra-manual/src/architecture.md
Ericson2314, line 1: Under `### Coordinator machine` we'll need to write up `hydra-ws` as another executable / service

Ericson2314 (approved):
I think I have read enough :) thankfully since this off the "critical path" as a new service, we don't have to be too concerned about things being not quite right --- they won't take down the rest of Hydra

foreman/start-hydra.sh
```diff
@@ -27,6 +27,7 @@ if [ ! -f "$HYDRA_DATA/hydra.conf" ]; then
 # test-time instances likely don't want to bootstrap nixpkgs from scratch
 use-substitutes = true
 queue_runner_endpoint = http://localhost:$HYDRA_QUEUE_RUNNER_REST_PORT
+ws_endpoint = ws://localhost:$HYDRA_WS_PORT
```
mweinelt, line 30: I think I would want to expose that endpoint through nginx for abuse handling, so the local listener and the ws endpoint url should be separate.

Conni2461: `ws_endpoint = wss://hydra.nixos.org/ws` needs to be configured in `services.hydra.extraConfig` and in nginx

```
        "/ws" = {
          proxyPass = "http://[::1]:9283";
          recommendedProxySettings = true;
          proxyWebsockets = true;
          extraConfig = ''
            proxy_read_timeout 600;
          '';
        };
```

foreman is only used for starting the development hydra.

subprojects/hydra/lib/Hydra/Helper/LogEndpoints.pm
```diff
@@ -26,6 +26,15 @@ sub showLog {
         $c->stash->{log_uri} = $log_uri . "?tail=$lines";
         $c->stash->{tail} = $lines;
         $c->stash->{template} = 'log.tt';
+
+        my $step = $c->stash->{step};
+        my $isLive = $step ? $step->busy != 0 : !$c->stash->{build}->finished;
+        if ($isLive && $c->config->{'ws_endpoint'}) {
```
dasJ, line 32: This could be one if, combined with the next one

Ericson2314: I will get this next

PR #1785, opened 2026-06-05
hydra-tests: pump nix daemon logs via `ProcessGroup`

Ericson2314:
The nix daemons (one for the queue-runner, one for the builder) were started with bare `IPC::Run` harnesses whose stdout/stderr were never flushed to test output. When a daemon misbehaved (e.g. connection reset), there was no log to diagnose it.

To fix this have `start_nix_daemon` register the daemon in the caller's `ProcessGroup`. The `ProcessGroup` machinery we made is exactly for this purpose, and we should have had this function use it.

In particular, now that it is part of the process group, `pump_logs` will pick it up. Also, there is no need for the separate `$daemon_harness` return value from `start_queue_runner` — callers no longer need to manually `kill_kill` daemon harnesses since `$pg->stop` handles everything — so that return value is removed too.

Finally, factor out the newly improved `start_nix_daemon` into its own module, since it should not be more strongly associated with either the queue-runner or the daemon.

The queue-runner daemon is labelled `queue-runner daemon` and the builder daemon `builder daemon`, making it easy to tell them apart in test output.

Hopefully this helps us debug the spurious issues we're seeing.

### DIFF
--- a/subprojects/hydra-tests/lib/NixDaemon.pm
+++ b/subprojects/hydra-tests/lib/NixDaemon.pm
@@ -1,0 +1,32 @@
+use warnings;
+use strict;
+
+package NixDaemon;
+use File::Path qw(make_path);
+our @ISA = qw(Exporter);
+our @EXPORT_OK = qw(start_nix_daemon);
+
+# Start a nix daemon for the given store config and register it in a
+# ProcessGroup so its logs are pumped alongside the other processes.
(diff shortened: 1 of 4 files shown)


Ericson2314:
I am going to merge this straight away because it is just testing logic, and it will help debug any more spurious errors we see in CI.

PR #1786, opened 2026-06-05
hydra-builder: share a single `ConnectionPool` across builds

Ericson2314:
The builder was creating a new `ConnectionPool` in every `process_build` call. Under concurrent load each pool opened its own set of daemon connections, so the total could reach `num_builds × (CPU_cores + 1)` — all hitting the same nix daemon and its SQLite database.

CI logs (now visible thanks to the daemon log pumping) showed `SQLite database is busy` errors from the builder daemon, which surfaced as `Connection reset by peer` on the builder side.

Fix by creating the pool once in `State::new` and storing it as `self.pool`.

### DIFF
--- a/subprojects/hydra-builder/src/state.rs
+++ b/subprojects/hydra-builder/src/state.rs
@@ -102,12 +102,13 @@ pub struct Config {
     pub real_store_dir: Option<std::path::PathBuf>,
 }
 
-#[derive(Debug)]
+#[allow(missing_debug_implementations)]
 pub struct State {
     pub id: uuid::Uuid,
     pub hostname: String,
     pub config: Config,
     pub max_concurrent_downloads: AtomicU32,
+    pub pool: harmonia_store_remote::ConnectionPool,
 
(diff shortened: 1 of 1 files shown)


amaanq (approved):
looks good to me

Mic92:
`SQLite database is busy` should not be a fatal error, so this might be a smoking gun. Have you correlated `connection reset` to this error?

Ericson2314:
@Mic92 I haven't reproduced it in isolation, no, but I think Nix says `error (ignored)` when it is non-fatal, so I also think it really is in fact erroneously fatal here.

artemist:
I have managed to get test failures on x86_64 by running with one processor (e.g. inside `systemd-run --pty -p User=$(whoami) -p AllowedCPUs=0 --wait $(which bash)`) but it's not the same as those on aarch64.

I was concerned about possible deadlocks because the total number of connections is dependent on number of cpus, but i have found no evidence that is the case.

PR #1793, opened 2026-06-11
binary-cache: read narinfos with harmonia's parser

Mic92:
The `StorePath` field of a narinfo holds the full path
(`/nix/store/<hash>-<name>`); cache.nixos.org serves it that way and
`harmonia_store_nar_info`, used for our own uploads, writes it that way.
Our hand-rolled `parse_narinfo` fed the value into `StorePath::from_str`,
which only accepts the base name, so every spec-conforming narinfo failed
to parse.

The queue runner therefore could not read back narinfos from its own
binary cache: `has_narinfo` failed for every path, uploads were repeated
endlessly and `copy_paths` errors failed the affected builds. Observed on
a 120-build nixos-small evaluation against an S3 cache populated by the
runner itself: 2067 `invalid value for StorePath` errors and all 120
builds failing with status 2.

harmonia has since grown its own `parse_narinfo_txt` (mirroring nix's
`nar-info.cc`), so delegate to it instead of maintaining our own. It takes
the store directory explicitly, so carry it on `S3CacheConfig`, read from
the `store` query parameter of the cache URL (the nix S3 convention,
defaulting to `/nix/store`); both the upload and download paths now use
`self.cfg.store_dir` rather than threading it in. Besides reading
`StorePath` as a full path, this also fixes two ways our parser diverged
from nix: `Deriver: unknown-deriver` and an optional `Compression` field.

### DIFF
--- a/subprojects/crates/binary-cache/src/narinfo.rs
+++ b/subprojects/crates/binary-cache/src/narinfo.rs
@@ -1,20 +1,18 @@
-use std::collections::BTreeSet;
-
 use harmonia_store_nar_info::UnkeyedNarInfo;
-use harmonia_store_path::{ParseStorePathError, StoreDir, StorePath};
+use harmonia_store_path::{StoreDir, StorePath};
 use harmonia_store_path_info::fingerprint_path;
 use harmonia_store_path_info::{NarHash, UnkeyedValidPathInfo};
 use harmonia_utils_hash::Hash;
 use harmonia_utils_hash::HashFormat as _;
-use harmonia_utils_signature::{SecretKey, Signature};
+use harmonia_utils_signature::SecretKey;
 use secrecy::ExposeSecret as _;
 
 use crate::Compression;
 
 pub use harmonia_store_nar_info::NarInfo;
 
-/// Re-export the harmonia narinfo formatter.
-pub use harmonia_store_nar_info::format_narinfo_txt;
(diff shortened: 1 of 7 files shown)


subprojects/crates/binary-cache/src/narinfo.rs
```diff
@@ -184,13 +184,17 @@ pub fn parse_narinfo(raw: &[u8]) -> Result<NarInfo, NarInfoError> {
 
         match key {
             "StorePath" => {
-                store_path_opt = Some(val.parse().map_err(|e: ParseStorePathError| {
-                    NarInfoError::InvalidValue {
+                // The field holds the full path, but `StorePath::from_str`
+                // only accepts the base name. Parsing via the store
+                // directory (assumed to be the default one throughout this
+                // function) also rejects narinfos for a different store.
+                store_path_opt = Some(StoreDir::default().parse(val).map_err(
```
Ericson2314: Fine for `hydra.nixos.org`, but the `StoreDir` here ought to be a parameter that we pass in.

(Also presumably this parsing logic is in Harmonia?)

Ericson2314: https://github.com/nix-community/harmonia/pull/1080 the logic is in Harmonia now (asked @amaanq to write from scratch)

Ericson2314 (approved):

PR #1794, opened 2026-06-16
tests: instrument flaky content-addressed early-cutoff test

Mic92:
`content-addressed/early-cutoff.t` fails intermittently in CI on the "One downstream build should be cached" assertion. The two downstream builds race to completion; when both finish before either is resolved, neither is marked cached.

This PR loops the scenario several times per run to make the race reproduce reliably in CI, and dumps per-build state (output paths, cache flags, timings) on failure.

Debugging aid to capture the race — not meant to merge as-is.


Mic92:
That one got fixed.

PR #1797, opened 2026-06-17
hydra-builder: move operational settings from CLI flags to a TOML config file

Ericson2314:
The builder took every tuning knob as a `--flag`, so the NixOS and Darwin modules reconstructed a dozen-argument `ExecStart` and each setting existed twice: once as a module option, once as a CLI flag. The queue runner already solved this with a `--config-path` TOML file; the builder now follows the same shape.

`config.rs` splits into three concerns:

- `Args` (the clap entry point) carries only `--config-path`, which is meta-config locating the file, plus a flattened `Cli`.

- `Cli` keeps the connection/security options that don't belong in a world-readable config file: gateway endpoint, mTLS cert paths, auth token.

- `AppConfig` (serde, `camelCase`, `deny_unknown_fields`) holds the operational settings, each defaulting to the old clap default, so a missing file behaves exactly as before.

`systems` and `supportedFeatures` stay `Option`: absent means "read from `nix show-config`", an explicit `[]` means "advertise none" — the modules type them as `nullOr (listOf ...)` so that distinction survives into the generated TOML. It was unclear how to take advantage of this `None` for `Some([])` distinction before --- certainly the old NixOS module was not doing it.

The modules now generate `/etc/hydra/builder.toml` with `pkgs.formats.toml` and pass only the connection/security flags.

This also drops the old `--use-substitutes` always-on bug: the modules guarded it with `useSubstitutes != null`, always true for a non-null bool, so the option never actually took effect.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1956,6 +1956,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-util",
+ "toml",
  "tonic",
  "tonic-prost",
  "tower",
--- a/darwin-modules/builder-module.nix
+++ b/darwin-modules/builder-module.nix
@@ -7,6 +7,8 @@
 let
   cfg = config.services.hydra-queue-builder-dev;
   user = config.users.users.hydra-queue-builder;
+
+  format = pkgs.formats.toml { };
 in
 {
   options = {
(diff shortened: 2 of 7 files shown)


Ericson2314:
(@mweinelt mentioned to me the issue with inheriting the default vs truly empty lists. The next step would be to make the NixOS module inherit from the Nix settings.)

Conni2461 (approved):

PR #1800, opened 2026-06-17
Fix precedence issue in Hydra API

mweinelt:
`!$project eq ""` is actually interpreted as `(!$project) eq ""`, certainly not what was intended here.


Ericson2314 (approved):

Ericson2314:
Good catch!

PR #1835, opened 2026-07-23
hydra-builder: Add delimiter to features CLI arguments

artemist:
By default clap does not split arguments on spaces, populating a `Vec<T>` in an args structure requires multiuple instances of the argument. This is sort of inconvenient, so add a delimiter where clap will split.

Using a space here instead of comma as space matches the format in the Nix configuration file.


Ericson2314:
I think https://github.com/NixOS/hydra/pull/1797 will obviate this?

artemist:
Yes, a delimiter should not be necessary in that case. I can close this PR if the other merges.

PR #1836, opened 2026-07-23
queue-runner: fix dynamic derivation scheduling and resolution

amaanq:
When trying to run nix-ninja jobs with dynamic derivation through the queue runner, I hit three bugs in the two-phase CA dance. Regular CA builds only ever produce depth-1 chains, so none of these paths had really been exercised before it seems.

- When a step resolves to a different drv path, the rdep migration used `make_rdep`, which records an empty `OutputNameChain`. As such, dynamic rdeps lost their relation, so `pop_dynamic_rdeps` found nothing when the resolved step finished and the generated derivation was never scheduled.
- `flatten_chain` pushed the outermost output name onto the pop end of the chain, so an input like `drv^out^hello` resolved outermost-first and asked the outer derivation for the inner one's output.
- A step whose resolved drv was already built errored with `ResolvedStepCreationFailed` and retried forever instead of finishing the build as cached. This is the early cutoff case the included test located at `early-cutoff-sequential.t` covers.

With these three bugs fixed, nix-ninja's example jobs build green end to end for me locally. The outer derivation runs nix-ninja inside the sandbox, the generated derivations get scheduled as real steps, and dependents resolve through the two-level chain!

### DIFF
--- a/subprojects/hydra-queue-runner/src/state/drv.rs
+++ b/subprojects/hydra-queue-runner/src/state/drv.rs
@@ -22,6 +22,24 @@ impl OutputNameChain {
     }
 }
 
+/// Walk down the nesting of `sdp`, appending each output name onto `chain`.
+/// Descending visits outermost names first, so plain pushes produce
+/// [`OutputNameChain`]'s stack order.
+fn flatten_into(
+    mut sdp: &SingleDerivedPath,
+    mut chain: OutputNameChain,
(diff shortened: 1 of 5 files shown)


tomberek:
I needed a few other fixes (https://github.com/NixOS/hydra/compare/master...tomberek:hydra:combined/dyn-drv-fixes-plus-stack-overflow), but eventually got this working also with the new ad-hoc daemon.

PR #1842, opened 2026-07-28
fix: migrate to 64bit timestamps

Conni2461:


subprojects/hydra/sql/migrations/upgrade-87.sql
```diff
@@ -0,0 +1,40 @@
...
+
+ALTER TABLE BuildSteps ALTER COLUMN startTime TYPE bigint;
+ALTER TABLE BuildSteps ALTER COLUMN stopTime TYPE bigint;
+
+ALTER TABLE BuildMetrics ALTER COLUMN timestamp TYPE bigint;
+
+ALTER TABLE CachedPathInputs ALTER COLUMN timestamp TYPE bigint;
+ALTER TABLE CachedPathInputs ALTER COLUMN lastSeen TYPE bigint;
+
+ALTER TABLE CachedCVSInputs ALTER COLUMN timestamp TYPE bigint;
```
dasJ, line 28: Isn't that VCS?

dasJ: What. It isn't

Conni2461: we could rename the table in the same migration

Ericson2314 (approved):

subprojects/hydra/sql/migrations/upgrade-87.sql
Ericson2314: I would support doing the rename, I don't care if it is in this migration or a latter one.

Ericson2314 (approved):

PR #1845, opened 2026-07-29
Add overflow caches

Mic92:
The high-level idea is that we can different buckets for staging jobsets and have a different retention policies for those.
So if a derivation is only in staging the output would only be in this bucket and later when the derivation is also in an nixos-unstable jobset, we use bucket side copy to migrate it.

This hopefully reduces the long-term storage consumption since we can throw the staging bucket away as we wish.

I have not tested this outside the NixOS test yet that I added.

### DIFF
--- a/subprojects/hydra-manual/src/configuration.md
+++ b/subprojects/hydra-manual/src/configuration.md
@@ -131,6 +131,27 @@ $ hydra-queue-runner --prometheus-address 127.0.0.1:9198
 $ hydra-queue-runner --prometheus-address [::]:9198
 ```
 
+Overflow binary cache (optional)
+--------------------------------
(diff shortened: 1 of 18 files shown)


subprojects/hydra-manual/src/configuration.md
```diff
@@ -131,6 +131,27 @@ $ hydra-queue-runner --prometheus-address 127.0.0.1:9198
 $ hydra-queue-runner --prometheus-address [::]:9198
 ```
 
+Overflow binary cache (optional)
```
Mic92, line 134: Read here if you want the full explanation.

Ericson2314 (approved):

PR #1608, opened 2026-03-27
hydra-evaluator: Rewrite in Rust

Ericson2314:
Port hydra-evaluator from C++ to Rust, removing the last C++ executable from the project. The new implementation uses sqlx and tokio, matching the async patterns already established by hydra-queue-runner and hydra-builder.

Build/packaging changes:
- Remove all C++ dependencies from `subprojects/hydra/meson.build`
- Move Rust packaging from `hydra-queue-runner/package.nix` to `subprojects/rust-package.nix` with named outputs (`queue_runner`, `builder`, `evaluator`)
- Add `evaluatorExecutable` option to the NixOS web-app module, matching how the queue runner and builder are discovered via their own modules
- Clean up stale references in `hydra-tests/meson.build` and `dev-shell.nix`

PR #1868, opened 2026-08-23
Teach Hydra about IFD

Ericson2314:
Depends on #1608


knedlsepp:
This sounds like serving instances of the new `daemon-server` over ssh could potentially be used to make hydra act as a remote builder, right? https://github.com/NixOS/hydra/issues/1193#issuecomment-1085393083 ❤️

PR #1875, opened 2026-08-25
db: force drvPath index in resolve_drv_output_chains

Mic92:
With `ORDER BY s.build DESC LIMIT 1` the planner walked `buildsteps_pkey` backwards filtering on drvPath. For a never-built drv that is a full scan of BuildSteps (>10 min on hydra.nixos.org, 338M rows), stalling the dispatcher so all builders sat idle. `+ 0` makes it use IndexBuildStepsOnDrvPath instead (0.4 ms).

EXPLAIN on hydra.nixos.org, original:
```
->  Index Scan Backward using buildsteps_pkey on buildsteps s  (cost=0.57..88972981.09 rows=859 width=40)
      Filter: ((drvpath = r_1.drv_path) AND ((status = 0) OR (status = 13)))
```
with `+ 0`:
```
->  Index Scan using indexbuildstepsondrvpath on buildsteps s  (cost=0.70..1204.70 rows=859 width=40) (actual time=0.072..0.072 rows=0.00 loops=1)
Execution Time: 0.406 ms
```


Mic92:
This is currently on the hydra commit we pinned in nixos infra...

Mic92:
we will merge the other.

PR #1886, opened 2026-08-31
hydra: Report evaluation errors from hydra-notify

Ericson2314:
The description is in the last commit --- the first two are prep commits that are good either way.


eveeifyeve:
One question, with this would I be able to setup my own custom error email template via the hydra plugin config?

Ericson2314:
@eveeifyeve Perhaps? I wasn't trying to expose more stuff to plugins. I was just trying to separate side-effects.

Ericson2314:
I've now spent a bunch of time manually testing this in the staging deployment, and it appears to be working!

PR #1894, opened 2026-09-04
hydra-evaluator: Take its settings from its own TOML file

Ericson2314:
The evaluator previously read `hydra.conf`, the Perl configuration, with a hand-rolled parser. This is not nice. Let's just use TOML like the other components.

The evaluator currently reads just one setting, `max_concurrent_evals`, which nothing else reads. So the shared file bought no sharing and cost a second syntax to parse in Rust. Do `/etc/hydra/evaluator.toml` instead, as the queue runner and the builder already have their own.

Also, make `Hydra::Config` warn when a moved setting is still in `hydra.conf`. That way, users are prompted to move them to the TOML file instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2000,8 +2000,11 @@ dependencies = [
  "fs-err",
  "futures",
  "hydra-tracing",
+ "serde",
  "sqlx",
+ "thiserror",
  "tokio",
+ "toml",
  "tracing",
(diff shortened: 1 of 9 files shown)


Ericson2314:
Pretty small diff from the LLM for me to read, and tested an eval in staging hydra.

PR #1904, opened 2026-09-14
hydra-builder: increase LimitNOFILE, avoid the OOM killer

vcunat:


vcunat:
This originated as https://github.com/NixOS/infra/pull/1212

vcunat:
These are just suggestions.  Feel free to adapt this.

PR #1910, opened 2026-09-20
nix-perl: Fix Perl Flags causing handshake mismatch on darwin

eveeifyeve:
Fixes: `lib/Nix/Store.cc: loadable library and perl binaries are mismatched (got first handshake key 0x10200080, needed 0xff80080)` when doing a nix develop or nix flake check

> Disclaimer: I used a coding agent in the creation of this patch.


Ericson2314 (approved):

eveeifyeve:
Now hydra-evaluator needs fixing, It's running into a role anonymous error.

